PR #2, opened 2017-09-28
avoid double state update before collision check

rhaschke:


v4hn:
Fair enough, I forgot about the overload here.

PR #3, opened 2017-09-28
cleanup publish plans

rhaschke:
publish one plan at a time
wait for a keypress in between
remove "handled" flag from SubTrajectory: I don't see it's use... Don't you want to iterate over all possible plans?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ target_link_libraries(${PROJECT_NAME}_subtasks ${catkin_LIBRARIES})
 add_library(${PROJECT_NAME}
 	src/subtask.cpp
 	src/task.cpp
+	src/debug.cpp
 )
 
 add_subdirectory(src/demo)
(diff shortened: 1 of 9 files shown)


rhaschke:
@v4hn I restored the "online" publishing of new solutions.
All the publishing stuff is in `debug.cpp` now, because it's not a core part of `Task`.
The set of already published `SubTrajectories` is handled in a helper class `NewSolutionPublisher`.
Next, I will probably start working on the rviz plugin.

v4hn:
> @v4hn I restored the "online" publishing of new solutions.
> All the publishing stuff is in debug.cpp now, because it's not a core part of Task.

Great, thanks! I'll push some minor changes to this later, but thanks a lot for the patch.

> Next, I will probably start working on the rviz plugin.

It would be great to see some progress there too.
I would like to have the ROS API as part of the Task class (maybe optional at runtime).
If you have a specification for the ROS interfaces, I would be glad to see them here even if their actual implementation is not yet ready (it'll go through multiple iterations anyway I guess).

PR #4, opened 2017-09-29
fix for-loop argument types

rhaschke:
It doesn't make sense to use a reference to a pointer.


v4hn:
doesn't matter too much, but I tend to agree. The important bit for me was to get rid of the `auto`s there.

PR #7, opened 2017-10-12
WIP: refactoring API

rhaschke:
If you want to merge, this is the right branch to use ;-)


v4hn:
I just applied a couple of changes on top of this to get it to compile with Qt4 and `catkin_make` and merged it as `master` branch.

It took me a while to figure out why I don't see numbers going up for successful/failure in the panel over here. But this has simply not been implemented yet, is that correct?

@rhaschke I don't expect any breakage, but could you please check if this still compiles on your setups?

I'll try to get the package structure cleaned up soon and will move it to ros-planning afterwards.

rhaschke:
@v4hn Thanks for pushing your changes.
I already continued working (also on package structure) and pushed my changes independently from yours. I think I already fixed most things we discussed 2 weeks ago. However, new features are still missing (e.g. property map). I will now merge and create a new PR.

> It took me a while to figure out why I don't see numbers going up for successful/failure in the panel over here. But this has simply not been implemented yet, is that correct?

Yes, wasn't implemented yet. Now it is. Please wait until tomorrow. I'm going to squash some of the last commits.

PR #9, opened 2017-12-12
PropertyMap implementation

rhaschke:
I implemented the `PropertyMap` we discussed a while ago.
Using this `PropertyMap` I also generalized the grasp frame specification for `GenerateGraspPose` - getting rid of hard-coded YPR angles. This allows for our PA10 example to be included easily.

PR #11, opened 2017-12-13
PropertyMap implementation

rhaschke:
Replacement for #9, implementing the `PropertyMap` we discussed a while ago.
Using this `PropertyMap` I also generalized the grasp frame specification for `GenerateGraspPose` - getting rid of hard-coded YPR angles. This allows for our PA10 example to be included easily.
I merged already with master to fix conflicts due to reordering of source files in cmake.


rhaschke:
@v4hn Could you please have a look?

v4hn:
ok, so this request has a number of small issues, but you already addressed and reworked some of them in the latter requests.

It compiles (up to one vector include on my machine) and runs, so I will merge it to get on with the other requests.

v4hn:
I should have fast-forwarded. This made the history even harder to read... *sigh*

PR #12, opened 2018-02-05
rhaschke's master

rhaschke:
This PR goes on top of #11 and adds the following features
- marker visualization (passed in solutions)
- failure introspection: storing and passing failures to rviz for introspection
- cost ordering of solutions
- priority ordering of interface states (by depth of incoming/outgoing solution and solution costs)
- various bug fixes, successfully tested for TAMS UR5, UNIBI PA10, PR2

Please **don't squash** these commits.

PR #15, opened 2018-02-19
set of bug fixes

rhaschke:
OK, thanks that you merged #12. Here comes a set of bug fixes (which become relevant for hierarchical nesting of containers). I suggest that you continue from there.
Remember to fast-forward this time ;-)


v4hn:
merged via fast-forward

PR #16, opened 2018-02-19
pick: reusable stage for picking up an object

rhaschke:
Implementation of a pick container. Example in trixi. but not yet working. Suffers from some missing configuration(s) for property forwarding - as @v4hn asked for making it all explicit ;-)

To assist the user in spotting these miss-configurations, I suggest - in case of property exceptions (we need a new exception type for this) - to climb the property tree and check where and why the definition is missing. Doing this manually is a nightmare.

### DIFF
--- a/core/demo/plan_pick_pa10.cpp
+++ b/core/demo/plan_pick_pa10.cpp
@@ -91,7 +91,7 @@ int main(int argc, char** argv){
 	{
 		auto gengrasp = std::make_unique<stages::GenerateGraspPose>("generate grasp pose");
 		gengrasp->properties().configureInitFrom(Stage::PARENT);
-		gengrasp->setGripperGraspPose("open");
+		gengrasp->setNamedPose("open");
 		gengrasp->setObject("object");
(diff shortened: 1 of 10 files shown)


v4hn:
> I suggest - in case of property exceptions (we need a new exception type for this) - to climb the property tree and check where and why the definition is missing. Doing this manually is a nightmare.

... which is exactly why I wanted to avoid dynamic property lookup in the beginning.
Now we have to deal with it one way or another and it's not too easy to simplify this for the user.

I'll have a look at the Trixi example tomorrow.

Overall the pick container looks awesome and it feels like all of this finally starts to pay off! :+1: 
(Assuming we don't hit a wall around the next corner)

rhaschke:
> ... which is exactly why I wanted to avoid dynamic property lookup in the beginning.
Now we have to deal with it one way or another and it's not too easy to simplify this for the user.

However, without property inheritance it's even more tedious and error-prone to define meta containers such as the pick container.

> I'll have a look at the Trixi example tomorrow.

Note: I just fixed the last remaining unittest in #15 and rebased both PRs #15 and #16.

Issue #21, opened 2018-02-21
MonitoringGenerator: use solutions of a monitored stage as basis for generation


rhaschke:
Sometimes its necessary to reuse a previously planned solution, e.g. to traverse it in reverse order or to access the state of another generator. The MonitoringGenerator stage hooks into the `onNewSolution()`.
- [x] For now: reference via raw pointers
- [ ] Later: make `Stage::pointer` a `shared_ptr` (instead of a `unique_ptr`)
- [x] Rewrite `CurrentScene` stage such that it fetches the current planning scene from `get_planning_scene` service (Luca)
- [x] Rewrite grasp generators to hook up on initial `CurrentScene` stage.
- [x] Rewrite `Stage::init()` to receive a `RobotModel` argument instead of the weird `PlanningScene` argument. All but generator stages, receive their planning scene through the `Interface` API.

v4hn:
I believe it would be a good idea to add a mechanism to retrieve a specific stage pointer from a container/Task, after we moved to `shared_ptr`.
Possibly recursively. Although I'm unsure whether want to use the descriptive name string as id here.

This would help with setting the monitored stage for monitoring generators.

rhaschke:
`MonitoringGenerator` is implemented. For `shared_ptr` stuff I opened the new issue #74.

PR #22, opened 2018-02-21
Monitoring generator

rhaschke:
Implementation for #21.


core/src/stage.cpp
```diff
@@ -497,6 +500,46 @@ void Generator::spawn(InterfaceState&& state, SubTrajectory&& t)
...
+
+void MonitoringGenerator::init(const moveit::core::RobotModelConstPtr& robot_model)
+{
+	Generator::init(robot_model);
+
+	auto impl = pimpl();
+	if (!impl->monitored_)
+		throw InitStageException(*this, "no monitored stage defined");
+	if (impl->monitored_->pimpl()->requiredInterface() != GENERATE)
+		throw InitStageException(*this, "monitored stage must have generator-style interface");
```
v4hn: It would be nice to avoid this restriction. I guess this is not just a matter of removing those two lines or is it?

v4hn: > This kind of stage always acts as a generator!

That's not what this checks (if I understand the code and you correctly).
The **monitoring** stage is always a generator, but I don't see what should be wrong about spawning new InterfaceStates based on solutions of any other (**monitored**) Stage.

At the moment I would like to monitor the end of solutions from a stage that modified the planning scene, so that I can spawn something based on it further back in the pipeline.

v4hn: I removed these lines locally and the pipeline behaves as expected:
for every solution generated by the (`PropagatingForward`) monitored stage the `MonitoringGenerator` can spawn new states.

Unless you see some other problem here, I would propose to remove these two lines.

rhaschke: You are right. The monitored stage doesn't need to be a generator.

rhaschke:
@v4hn I addressed all your remarks.

PR #28, opened 2018-02-26
implement path constraints in solver API

v4hn:


v4hn:
> path_constraints property should be added to other planning stages (MoveTo, MoveRelative) too. Otherwise lgtm.

I added them.

PR #29, opened 2018-02-26
add a simple GeneratePose stage

v4hn:
ComputeIK is a wrapper, so we can't just give it a pose to compute.

This stage is useful to specify link poses relative to planning scenes generated by other stages,
i.e. a generator stage for ``go here at some point after stage X has changed the planning scene''.

### DIFF
--- a/core/include/moveit/task_constructor/stages/generate_grasp_pose.h
+++ b/core/include/moveit/task_constructor/stages/generate_grasp_pose.h
@@ -38,16 +38,15 @@
 
 #pragma once
 
-#include <moveit/task_constructor/stage.h>
+#include <moveit/task_constructor/stages/generate_pose.h>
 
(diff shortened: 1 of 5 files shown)


rhaschke:
I added some more commits... @v4hn Please review.

I'm also thinking about a generic way to sample around the initial `target_pose`, i.e. generalizing the additional sampling in `GenerateGraspPose` to allow for duplication of a pose along xyz axes (position or orientation), spherical, cylindrical sampling etc. Also the sampling methods should optionally _seed_ at the initial pose and slowly move away, i.e. `delta = {0, +D, -D, +2D, -2D, +3D, -3D, ...}`.

I guess there are samplers available already in MoveIt. Can we reuse them? Where are they?

rhaschke:
Closing this as there was no feedback and `rhaschke-master` progressed much more in the meantime.

PR #30, opened 2018-02-26
ComputeIK: fix frame_id abuse

v4hn:
frame_id was used to specify the robot frame to move to target.
This is just wrong and does not leave the option to specify targets w.r.t. other frames...

Now there is a separate `target_link` property instead.

This patch changes the semantics of the `PoseStamped` message expected by `ComputeIK`,
so maybe your demo code is affected.


rhaschke:
I suggest a more general solution that also allows to specify an offset transform w.r.t. to the link frame.
Your use case is mapped onto `ComputeIK::setIKFrame(link_name)`.
Closing in favour of #36.

PR #35, opened 2018-03-25
fix travis: build not yet released dependencies

rhaschke:


v4hn:
Looks good.
Hopefully we will do the next MoveIt release soon anyway, but I see no harm in working around it until then.

rhaschke:
For some reason, travis is not triggered anymore - since this commit. However, I don't see a reason.

PR #36, opened 2018-03-25
ComputeIK: allow arbitrary link frame

rhaschke:
In contrast to #30, this PR also allows to define an arbitrary offset transform from a link to be used as IK link. This allows to replace the `ToolToGrasp` transform of `GenerateGraspPose` with this stage's more general `IKFrame`.


v4hn:
That's a nice and valid generalization, including quite a lot of cleanup. Yay!

I will probably stick with identities in your generalized reference frame, because it becomes somewhat harder to debug geometric failures with one more pose factorized in there.
But of course this is a matter of personal preference.

v4hn:
I just fast-forwarded `master` to this branch.
Looks like github missed this for some reason.

Closing.

PR #38, opened 2018-03-25
reworked connect stage

rhaschke:
As already announced, this is my attempt to generalize the `connect` stage. So far, it was not possible to connect two states that differed w.r.t. multiple joint model groups. For example, for grasping, both the gripper and arm groups have changed for the _pregrasp pose_ compared to the _initial_ pose.
The new connect stage allows to specify an (ordered) vector of `JMGs` and associated planners to resolve differences.

This PR also removes `connect` from the `pick` stage. `connect` is considered as a generator of the transit motion only and should not be part of a functional container.

Finally, I implemented releasing and placing as _inversions_ of grasping and picking, using the same base classes for each pair. An example of usage is provided in [`mtc_demos`](https://github.com/ubi-agni/mtc_demos/blob/master/src/yumi.cpp).

I also implemented merging of trajectories. Here I faced some unexpected issues: MoveIt! doesn't nicely support creating of arbitrary RobotTrajectories - they are always bound to a JMG. Hence, I had to create a JMG on-the-fly.

### DIFF
--- a/core/demo/plan_pick_pa10.cpp
+++ b/core/demo/plan_pick_pa10.cpp
@@ -73,16 +73,10 @@ int main(int argc, char** argv){
 		stage->setMaxPenetration(0.1);
 		t.add(std::move(stage));
 	}
-	{
-		auto move = std::make_unique<stages::MoveTo>("open gripper", pipeline);
-		move->restrictDirection(stages::MoveTo::FORWARD);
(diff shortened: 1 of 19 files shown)


rhaschke:
@v4hn Please review this and #29. If you like them, you can fast-forward to `rhaschke-master`.

rhaschke:
Closing this as there was no feedback and `rhaschke-master` progressed much more in the meantime.

PR #45, opened 2018-04-15
inf cost states always go last

rhaschke:
This PR was originally proposed by @v4hn. 
Also update sorted interface when state becomes inf or get's new cost.

### DIFF
--- a/core/include/moveit/task_constructor/storage.h
+++ b/core/include/moveit/task_constructor/storage.h
@@ -94,6 +94,14 @@ public:
 			                this->cost() + other.cost());
 		}
 		inline bool operator<(const Priority& other) const {
+			/* infinite cost should always be last */
+			if (std::isinf(this->cost()) && std::isinf(other.cost()))
+				return this->depth() > other.depth();
+			else if (std::isinf(this->cost()))
+				return false;
+			else if (std::isinf(other.cost()))
+				return true;
+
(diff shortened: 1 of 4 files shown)


rhaschke:
@v4hn Did you observe particular issues requiring these changes? I don't see a reason, why to deviate from the rather clean definition of state order.

Actually, an infinite-cost state should be removed from the interfaces completely. For the propagating stages, this is achieved in [`dropFailedStarts` / `dropFailedEnds`](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stage.cpp#L347L356).

v4hn:
> I don't see a reason, why to deviate from the rather clean definition of state order.

I don't see a reason to sort infinite cost solutions earlier just because they are longer.
This is arbitrary too.

> Did you observe particular issues requiring these changes?

Yes, in `Connecting` we currently abort matching pairs [when an infinite-cost trajectory is hit](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stage.cpp#L620).
Of course this can be changed too.

When I wrote this patch I had in mind that I want to have a flag to compute all solutions, even if they became infeasible somewhere else, in the future.
Without this it can be somewhat hard to identify the current problem when designing a Task.
But even then, the infeasible trajectories should go last.

PR #48, opened 2018-05-30
remaining local modifications

v4hn:
These are pretty much the useful&required modifications I apply locally.

### DIFF
--- a/core/src/stage.cpp
+++ b/core/src/stage.cpp
@@ -687,8 +687,9 @@ bool Connecting::compatible(const InterfaceState& from_state, const InterfaceSta
 
 		for (auto from_it = from_object->shape_poses_.cbegin(), from_end = from_object->shape_poses_.cend(),
 		     to_it = to_object->shape_poses_.cbegin(); from_it != from_end; ++from_it, ++to_it)
-			if (!from_it->matrix().array().isApprox(to_it->matrix().array()))
+			if (!(from_it->matrix() - to_it->matrix()).isZero(1e-4)){
 				return false;  // transforms do not match
+			}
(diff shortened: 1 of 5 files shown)


core/src/stage.cpp
```diff
@@ -687,8 +687,9 @@ bool Connecting::compatible(const InterfaceState& from_state, const InterfaceSta
 
 		for (auto from_it = from_object->shape_poses_.cbegin(), from_end = from_object->shape_poses_.cend(),
 		     to_it = to_object->shape_poses_.cbegin(); from_it != from_end; ++from_it, ++to_it)
-			if (!from_it->matrix().array().isApprox(to_it->matrix().array()))
```
rhaschke: You are right as long as matrices are concerned. However, here I turned the matrices into arrays first, which then uses element-wise operations - also for comparison.
However your notation is maybe more readable anyway. Please add spaces around the minus sign.

How do you motivate your choice of 1e-4? This is rather large, isn't it?
Why don't you choose a value close to machine precision as isApprox does?

v4hn:
> How do you motivate your choice of 1e-4? This is rather large, isn't it?
> Why don't you choose a value close to machine precision as isApprox does?

The problem is the goal tolerances of sampling planners.
While the targets are indeed precise up to machine precision, the planners generate solutions below some tolerance thresholds - usually `1e-4 ~ 1e-5` and thus states that are intuitively compatible can become incompatible.

I'm still not entirely happy with our current definition of "compatible".
Maybe I still want callbacks here to allow users to judge whether or not something is still compatible, but it's not trivial to come up with something usable.

PR #53, opened 2018-06-05
keep RobotModelLoader around

v4hn:
Otherwise the robot_model_ does not remain valid and the kinematic plugins might not be around anymore until the model is loaded again by a different RML.

### DIFF
--- a/core/include/moveit/task_constructor/task.h
+++ b/core/include/moveit/task_constructor/task.h
@@ -51,6 +51,10 @@ namespace moveit { namespace core {
 	MOVEIT_CLASS_FORWARD(RobotState)
 }}
 
+namespace robot_model_loader {
+	MOVEIT_CLASS_FORWARD(RobotModelLoader)
+}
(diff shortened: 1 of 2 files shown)


rhaschke:
`robot_model_` is a `shared_ptr`, so it should remain valid and I didn't have issues.
The kinematics loader isn't accessed from the task, is it?

v4hn:
> robot_model_ is a shared_ptr, so it should remain valid and I didn't have issues.

That logic is correct, but only applies to reasonable framework design...
With pluginlib, the lifetime of plugins is bound to the class loader, not the plugin instances.
You patched a number of things in this library before, so I thought you are well-aware of this.

This patch fixed the nasty segmentation fault I was talking about the other day.

v4hn:
Merged as https://github.com/ros-planning/moveit_task_constructor/commit/392e56500c1237452df5000c26c15b0d8fc5103a

PR #54, opened 2018-06-05
ExecuteTaskSolution capability

v4hn:
This is the `move_group` capability we've been talking about.

I called the package `capabilities` because I plan to have another capability where Tasks can be defined via yaml and planned in the `move_group`.

It's probably not polished everywhere yet, but I can use it over here with a small `actionlib::SimpleActionClient`.

Relies on https://github.com/ros-planning/moveit/pull/939 and https://github.com/ros-planning/moveit/pull/940 to execute.


capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <moveit/plan_execution/plan_execution.h>
+#include <moveit/trajectory_processing/trajectory_tools.h>
+#include <moveit/kinematic_constraints/utils.h>
+#include <moveit/move_group/capability_names.h>
+#include <moveit/robot_state/conversions.h>
+
+namespace {
+
+// TODO: move to moveit::core::RobotModel
+const moveit::core::JointModelGroup* findJointModelGroup(const moveit::core::RobotModel& model, const std::vector<std::string>& joints) {
```
rhaschke: This is an issue: The `Connect` stage creates robot trajectories that are merged from multiple `JointModelGroups`.
Hence, there doesn't exist a `JMG` for these trajectories!
However, I think `plan_execution` can handle them anyway: In https://github.com/ubi-agni/moveit_task_constructor/blob/automatica/core/python/wrapper/src/core.cpp#L102 I have used an arbitrary (the first) JMG to execute my trajectories. AFAIK the `ExecutionManager?` tries to find matching groups for any trajectory - that's the code you called over-engineered once ;-)

v4hn:
> Do you plan do add action feedback?

https://github.com/ros-planning/moveit_task_constructor/pull/54/files#diff-32963aababc0ffddcc710b2830337471R167

It's straight-forward to extend. My focus for now is to actually use the MTC with our robot.

v4hn:
Our CI seems to use the synced MoveIt release and thus fails because @2scholz's patch is not included.
Could you maybe have a look at that?

capabilities/src/execute_task_solution_capability.cpp
v4hn: Ok, things are a bit more ugly than I thought.

It is possible to create trajectories without a specified `JointModelGroup`.
But trajectory waypoints in `robot_trajectory::RobotTrajectory` are `RobotState`s which have no notion of which joints move anymore.
This is what the specified JMG is used for.

[`plan_execution` converts the `robot_trajectory::RobotTrajectory` objects to `moveit_msgs::RobotTrajectory`](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L382)
to passes them to the `TrajectoryExecution` and [`robot_trajectory::RobotTrajectory::getRobotTrajectoryMsg`](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L200) assumes that all active joints might move and thus adds all of them to the trajectory message.
This is clearly not what should happen here but we need to modify `robot_trajectory::RobotTrajectory` to account for that - possibly adding a list of active joints to the object.

Your "trick", using a random group, only works because the trajectories are forwarded to the `TrajectoryExecutionManager` as `moveit_msgs::RobotTrajectory` there, no matter which group is associated with the MGI. - Btw this also entails that these trajectories are not validated online.

For the plan execution the group actually matters as described above.

Until the `RobotTrajectory` class is changed, I propose to leave this as is and define auxiliary groups. In your case you can simply define meta-groups that contain the groups that might be merged in the srdf. I don't like the additional complexity of sets of groups here because it adds an unnecessary layer of abstraction.

PR #68, opened 2018-11-06
fix broken rvalue-forward & SerialContainer assert

v4hn:
Release mode builds previously produced broken solutions with too many entries,
debug build triggered the assert

container.cpp:334: assert(solution.empty())

The standard guarantees std::vector(&&a) leaves a.empty() == true,
so the logic there is fine as long as subsolutions is actually
used for move-construction.


v4hn:
subsumed in #69

PR #69, opened 2018-11-09
various minor improvements

v4hn:
A set of (from my perspective) uncontroversial commits to improve current master.

Feel free to cherry-pick commits to master.

PR #70, opened 2018-11-14
Remaining commits of #69

rhaschke:
This one needs discussion.
Moving the configuration of motion plan and request publishers from `init()` to `plan()` causes issues if you want to use multiple planning pipelines.
I don't remember the details anymore, but maybe it's not possible to create multiple publishers on the same topic within the same node?

### DIFF
--- a/core/include/moveit/task_constructor/solvers/pipeline_planner.h
+++ b/core/include/moveit/task_constructor/solvers/pipeline_planner.h
@@ -39,6 +39,7 @@
 #pragma once
 
 #include <moveit/task_constructor/solvers/planner_interface.h>
+#include <moveit_msgs/MotionPlanRequest.h>
 #include <moveit/macros/class_forward.h>
 
 namespace planning_pipeline {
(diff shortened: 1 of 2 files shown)


rhaschke:
Looking into this again, I figured that this PR essentially attempts to move the properties `display_motion_plans` and `publish_planning_requests` [from the  `PlannerInterface`](https://github.com/ros-planning/moveit_task_constructor/blob/1e0a9401e703ce7ec8d83c7f6f8ed0af511e3705/core/src/solvers/planner_interface.cpp#L46-L47) to the `Stage`.
Why do you want to do this? These properties are actual properties of the MoveIt! `PlanningPipeline`.

If this should be accepted (meaning that the stages using the PlannerInterface will reconfigure these properties on each `compute()` execution), the declaration of the corresponding properties should move from PlannerInterface to Stage as well.

v4hn:
I wrote this patch because I wanted to investigate the planned trajectories of *a specific* stage and the sampling-based planner automatically got shared between various stages (cluttering the debug topic).

I agree that this is not a durable patch to fix this issue and by now I believe I should have spent the time looking into the reasons for you to add implicit planner-caching, which is the real issue from my point of view.

Issue #79, opened 2019-02-13
ComputeIK: initialize timeout from ros parameter kinematics_solver_timeout


rhaschke:
Currently we use hard-coded default timeouts for all stages, 1s for ComputeIK.
When no IK can be found, this dramatically slows down computation, as the full time is used to eventually find a solution. ComputeIK should use the `kinematics_solver_timeout` specified in `kinematics.yaml` for a particular planning group, because that's a much better estimate of typical run times of the actual IK plugin.

v4hn:
Addressed in #80

PR #80, opened 2019-02-14
ComputeIK: set default timeout from JMG

rhaschke:
Addresses #79.
Also reduces default timeout for `MoveTo` / `MoveRelative` stages from 10s to 1s.

**This PR branch is included in #81. Please, fast-forward to #81 for merging.**

### DIFF
--- a/core/include/moveit/task_constructor/properties.h
+++ b/core/include/moveit/task_constructor/properties.h
@@ -65,7 +65,7 @@ boost::any fromName(const PropertyMap& other, const std::string& other_name);
  *
  * Setting the value via setValue() updates both, the current value and the default value.
  * Using reset() the default value can be restored.
- * Using setCurrentValue() only updates the current value, allowing for later reset to the original default.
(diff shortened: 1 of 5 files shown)


core/include/moveit/task_constructor/stages/compute_ik.h
```diff
@@ -107,6 +107,10 @@ class ComputeIK : public WrapperBase {
 	void setMinSolutionDistance(double distance) {
 		setProperty("min_solution_distance", distance);
 	}
+
+	void setDefaultTimeout(double timeout) {
+		properties().property("timeout").setDefaultValue(timeout);
+	}
```
v4hn: I would propose to leave this function out of the interface. It's nothing the user should touch, they can always `setTimeout` if they want to.

rhaschke: No problem. Already changed locally, need to cleanup my repo later...

rhaschke: Done.

PR #96, opened 2019-07-11
Make place rotation optional (GeneratePlacePose)

henningkayser:
Makes the place pose explicit by default and adds a property to enable sampling different place orientation.

### DIFF
--- a/core/include/moveit/task_constructor/stages/generate_place_pose.h
+++ b/core/include/moveit/task_constructor/stages/generate_place_pose.h
@@ -54,6 +54,9 @@ public:
 
 	void setObject(const std::string &object) { setProperty("object", object); }
 
+	void setAugmentRotations(bool augment_rotations) { setProperty("augment_rotations",
+                                                                  augment_rotations); }
+
 protected:
 	void onNewSolution(const SolutionBase& s) override;
 };
--- a/core/include/moveit/task_constructor/stages/generate_place_pose.h
+++ b/core/include/moveit/task_constructor/stages/generate_place_pose.h
@@ -54,8 +54,8 @@ public:
 
 	void setObject(const std::string &object) { setProperty("object", object); }
 
-	void setAugmentRotations(bool augment_rotations) { setProperty("augment_rotations",
-                                                                  augment_rotations); }
+	void setConsiderSymmetry(bool consider_symmetry) { setProperty("consider_symmetry",
+                                                                  consider_symmetry); }
 
 protected:
 	void onNewSolution(const SolutionBase& s) override;
--- a/core/src/stages/generate_place_pose.cpp
+++ b/core/src/stages/generate_place_pose.cpp
@@ -53,6 +53,7 @@ GeneratePlacePose::GeneratePlacePose(const std::string& name)
 	auto& p = properties();
 	p.declare<std::string>("object");
 	p.declare<geometry_msgs::PoseStamped>("ik_frame");
+	p.declare<bool>("augment_rotations");
 }
 
 void GeneratePlacePose::onNewSolution(const SolutionBase& s)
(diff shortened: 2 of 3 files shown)


core/include/moveit/task_constructor/stages/generate_place_pose.h
```diff
@@ -54,6 +54,9 @@ class GeneratePlacePose : public GeneratePose {
 
 	void setObject(const std::string &object) { setProperty("object", object); }
 
+	void setAugmentRotations(bool augment_rotations) { setProperty("augment_rotations",
```
rhaschke: I don't like the property name yet. What about `consider_symmetry`?
Additionally to this boolean flag, it would be nice to also set the number of z rotations to use.

core/src/stages/generate_place_pose.cpp
```diff
@@ -53,6 +53,7 @@ GeneratePlacePose::GeneratePlacePose(const std::string& name)
 	auto& p = properties();
 	p.declare<std::string>("object");
 	p.declare<geometry_msgs::PoseStamped>("ik_frame");
+	p.declare<bool>("augment_rotations");
```
rhaschke: Please add a property description and the default of `true` to maintain the previous behavior.

core/src/stages/generate_place_pose.cpp
```diff
@@ -138,25 +139,27 @@ void GeneratePlacePose::compute() {
...
-		}
-		case shapes::SPHERE:  // keep original orientation and rotate about world's z
-			target_pose.linear() = orig_object_pose.linear();
-			spawner(target_pose, 1);
-			return;
-		default:
-			break;
-		}
-	}
+  if (props.get<bool>("augment_rotations")) {
```
rhaschke: Please fix the indentation.

core/include/moveit/task_constructor/stages/generate_place_pose.h
v4hn: How about a single property then, that defaults to 1 or 0?

rhaschke: Exactly. We need both parameters independently.

v4hn: I'm inclined to make the same argument as in #98 here then.
Should `GeneratePlacePose` be a general-purpose stage that handles that many alternatives?
You can always plug together multiple of them with a parallel container :sunglasses:

rhaschke: I think, it would be great to have:
- a generic `CartesianPose` generator, which just creates a fixed pose relative to another frame (much like `FixedState`)
- a generic `VaryPose` generator that wraps CartesianPose and allows to augment different variations of the original pose(s) received from the wrapped stage.

Of course, one could also combine both stages. In any case, both GenerateGraspPose and GeneratePlacePose could be replaced by (or based on) this more generic stage.

henningkayser: > I'm inclined to make the same argument as in #98 here then.
> Should `GeneratePlacePose` be a general-purpose stage that handles that many alternatives?
> You can always plug together multiple of them with a parallel container

`GeneratePlacePose` could be a simple Cartesian pose generator and we could add a wrapper that samples for alternative orientations of the object. I think it make sense to provide a lot of primitive stages (i.e. CartesianPose) but also keep high-level versions for convenience and semantics.

rhaschke:
@henningkayser, do you have resources to implement https://github.com/ros-planning/moveit_task_constructor/pull/96#discussion_r303078785? Actually, I would like to close this PR in favor of the proposed approach.

henningkayser:
> @henningkayser, do you have resources to implement [#96 (comment)](https://github.com/ros-planning/moveit_task_constructor/pull/96#discussion_r303078785)? Actually, I would like to close this PR in favor of the proposed approach.

I will prepare something over the next days.

iamrajee:
I was curious to know if there is any update on this PR. I have a task in which I want to give some tolerance in the z-axis while placing the object? or is there a way in general to give some tolerance in both place pose and orientation? Thanks!

rhaschke:
I guess this PR is superseeded by https://github.com/ros-planning/moveit_task_constructor/pull/166. @henningkayser, can we close this here?

henningkayser:
> I guess this PR is superseeded by #166. @henningkayser, can we close this here?

yep

PR #97, opened 2019-07-11
adding cost calculations to connect and move_to

tylerjw:
cost calculations are based on sum of trajectory duration

### DIFF
--- a/core/src/stages/connect.cpp
+++ b/core/src/stages/connect.cpp
@@ -190,6 +190,13 @@ SolutionSequencePtr Connect::makeSequential(const std::vector<robot_trajectory::
 	planning_scene::PlanningSceneConstPtr start_ps = *scene_it;
 	const InterfaceState* state = &from;
 
+	// calculate cost
+	double cost = 0;
+	for (const auto& trajectory: sub_trajectories) {
+		for (const double& distance: trajectory->getWayPointDurations())
+			cost += distance;
+	}
+
 	SolutionSequence::container_type sub_solutions;
 	for (const auto &sub : sub_trajectories) {
 		planning_scene::PlanningSceneConstPtr end_ps = *++scene_it;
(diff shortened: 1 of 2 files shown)


core/src/stages/connect.cpp
```diff
@@ -219,9 +219,16 @@ SubTrajectoryPtr Connect::merge(const std::vector<robot_trajectory::RobotTraject
                                 const std::vector<planning_scene::PlanningSceneConstPtr>& intermediate_scenes,
                                 const moveit::core::RobotState& state)
 {
+	// set cost
+	double cost = 0;
+	for (const auto& trajectory: sub_trajectories) {
+		for (const double& distance: trajectory->getWayPointDurations())
```
rhaschke:
```suggestion
		cost += trajectory->getWayPointDurationFromStart();
```

core/src/stages/move_to.cpp
```diff
@@ -234,11 +234,18 @@ bool MoveTo::compute(const InterfaceState &state, planning_scene::PlanningSceneP
 		success = planner_->plan(state.scene(), *link, target_eigen, jmg, timeout, robot_trajectory, path_constraints);
 	}
 
+	// set cost
+	double cost = 0;
+	for (const double& distance: robot_trajectory->getWayPointDurations()) {
```
rhaschke: see above

core/src/stages/connect.cpp
v4hn: There is no overload for this without a parameter and `trajectory->getWayPointDurationFromStart(trajectory->getWayPointCount())`
is not any more readable or efficient than the current one.

A new `trajectory->getDuration()` function in MoveIt would be appreciated, as the class already provides a `getAverageSegmentDuration()` too...

tylerjw:
> Did you leave out the `makeSequential` case of `Connect` on purpose @tylerjw ?

No, good catch.  Fixed.

v4hn:
Pushed as 7ca49bc29be75618d49f06603fc5c9f59274d91a .
Thanks @tylerjw !

See #101 for the follow-up simplification.

PR #98, opened 2019-07-11
Can now grasps some object types from the top

simonGoldstein:
This adds the functionality to grasp an object from the top of the object in the generate_grasp_pose stage. It will have no affect if the flag is set to false, but if set to true it will now generate grasps from above the object.

### DIFF
--- a/core/include/moveit/task_constructor/stages/generate_grasp_pose.h
+++ b/core/include/moveit/task_constructor/stages/generate_grasp_pose.h
@@ -52,6 +52,7 @@ public:
 	void setEndEffector(const std::string &eef) { setProperty("eef", eef); }
 	void setObject(const std::string &object) { setProperty("object", object); }
 	void setAngleDelta(double delta) { setProperty("angle_delta", delta); }
+        void setTopGraspEnable(bool enabled) { setProperty("top_grasp_enabled", enabled); }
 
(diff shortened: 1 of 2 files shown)


core/include/moveit/task_constructor/stages/generate_grasp_pose.h
```diff
@@ -52,6 +52,7 @@ class GenerateGraspPose : public GeneratePose {
 	void setEndEffector(const std::string &eef) { setProperty("eef", eef); }
 	void setObject(const std::string &object) { setProperty("object", object); }
 	void setAngleDelta(double delta) { setProperty("angle_delta", delta); }
+        void setTopGraspEnable(bool enabled) { setProperty("top_grasp_enabled", enabled); }
```
henningkayser:
```suggestion
        void setTopGraspEnabled(bool enabled) { setProperty("top_grasp_enabled", enabled); }
```

v4hn:
I agree with @rhaschke .

`GenerateGraspPose` already does much more than it was ever supposed to do as an example generator stage.

Feel free to implement your new shape-aware logic in a new stage.
If it is reusable, we can merge it here too.

I will close this for now as *won't merge*.

henningkayser:
@v4hn I see your point. I actually like the idea of having a set of primitive grasp stages and to put all of them in an alternative container and let MTC optimize for the "best" grasp automatically.

PR #99, opened 2019-07-12
[WIP] Python API

rhaschke:
This is just a draft PR to collect TODOs and make the existing Python API visible somehow.
Major issues:
- `boost::python` doesn't support `std::unique_ptr` forcing us to use deprecated `std::auto_ptr`.
- Python-based stages are not yet working.
- I just stumbled across [pybind11](https://pybind11.readthedocs.io/en/master), which claims to be a light-weight replacement for boost::python. And it's more actively developed.


rhaschke:
As we discussed recently, the mid-term goal is to replace `std::unique_ptr` with `std::shared_ptr`, which will also have other benefits apart from better `boost::python` integration. As an alternative wrapper tool, I already proposed [pybind](https://pybind11.readthedocs.io/en/master), which seems to be very similar in API to boost::python.
I don't get your macro suggestion yet. Could you please draft an example, how this would look like in user code?

mamoll:
It'd be very cool to have Python bindings, but they can also be a real pain to maintain as the rest of the C++ code evolves. It is not only a matter of making the code compile and pass unit tests, but you also have to maintain documentation/tutorials/etc. for the Python API. I go and back forth about whether developing python bindings for OMPL was a good idea.

Pybind11 does look very promising. It contains [automatic converters for many STL data structures](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html). You may also want to look into [Binder](https://github.com/RosettaCommons/binder), a tool developed on top of Pybind11 to programmatically generate Python bindings. (I used Py++/pygccxml/castxml + Boost.Python in OMPL for this.)

v4hn:
Although I'm usually in favor of pure C++ approaches, Python bindings for MTC simplify prototyping *a lot* unless we develop a full-fledged pipeline editor in C++.
Recompiling every single time to change a minor attribute you did not put into a parameter yet can be quite annoying.

rhaschke:
@mamoll: Locally, I have already migrated to pybind11: https://github.com/ubi-agni/moveit_task_constructor/tree/wip-pybind11 and it works like a charm.
However, pybind11 doesn't support move semantics and the maintainer doesn't want to merge my [PR](https://github.com/pybind/pybind11/pull/2046) to change that: https://github.com/pybind/pybind11/pull/2046#issuecomment-619578241. I didn't have the time yet to discuss with him...

PR #102, opened 2019-07-12
WIP clang-format

v4hn:
Add a clang-format specification and format complete code base.
This is an initial attempt to discuss what style we actually want to enforce.

A big difference to the moveit-internal format we have is the use of tabs.

I glanced over the entailed changes and changed a few options compared to MoveIt core to reduce the "cleanup" diff a bit. I did not look through most of the details yet though.

@rhaschke please have a look and give your opinion on what style you prefer.


rhaschke:
@v4hn, thanks for starting this effort. I adapted some parameters. Please have a final look.

.clang-format
```diff
@@ -4,7 +4,7 @@ BasedOnStyle:  Google
 AccessModifierOffset: -3
 AlignAfterOpenBracket: Align
 AlignEscapedNewlinesLeft: true
-AlignTrailingComments: true
+AlignTrailingComments: false
```
rhaschke: I like this one.

.clang-format
```diff
@@ -6,7 +6,7 @@ AlignAfterOpenBracket: Align
 AlignEscapedNewlinesLeft: true
 AlignTrailingComments: true
 AllowAllParametersOfDeclarationOnNextLine: false
-AllowShortFunctionsOnASingleLine: Empty
+AllowShortFunctionsOnASingleLine: Inline
```
rhaschke: I explicitly decided against this one, but I'm fine with this setting as well.

rhaschke:
@v4hn, thanks for squashing the commits.

v4hn:
> @v4hn, thanks for squashing the commits.

Does not really matter as you decided catkin_lint, clang-format file and
the respective formatting patch constitute a single commit. ;)

PR #115, opened 2019-11-26
Change solution order

v4hn:
Sort solutions by cost according to the triangle icon shown in the QT table header.
The order was reversed w.r.t. the triangle symbol before.

I just found this patch lying around in one of my workspaces and briefly pushed it.

This triggers a failure in your tests @rhaschke. Could you have a brief look?
If this doesn't make sense to you (or you can't reproduce), feel free to just close it.

### DIFF
--- a/visualization/motion_planning_tasks/src/remote_task_model.cpp
+++ b/visualization/motion_planning_tasks/src/remote_task_model.cpp
@@ -587,9 +587,9 @@ void RemoteSolutionModel::sortInternal() {
 			          int comp = 0;
 			          switch (sort_column_) {
 				          case 1:  // cost order
-					          if (left->cost_rank < right->cost_rank)
+					          if (left->cost_rank > right->cost_rank)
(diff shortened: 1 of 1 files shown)


rhaschke:
Are you sure? Ascending order corresponds to a downward-pointing triangle in Qt:
image    image

v4hn:
Wow. I'm astonished I haven't encountered this any time before!
I was convinced this is universally interpreted the other way around...  :confused: 

Searching for the topic also did not turn up any good documentation that would include the visualization...

Thank you!

rhaschke:
My thunderbird uses the same downward-point arrow for ascending order:
image

Where have you seen it the other way around?

v4hn:
> Where have you seen it the other way around?

Now that you're asking, that's a very good question...
All tools I use on my end agree with your interpretation.
Maybe I stick with command line tools a bit too much these days. :D

Issue #114, opened 2019-11-26
Segmentation fault in MoveRelative::getJointStateFromOffset


progtologist:
Not a moveit guru here, but I am getting a segmentation fault that starts in [this](https://github.com/ros-planning/moveit_task_constructor/blob/145bec1ed3b6639d0b937b155e362e063860f4b3/core/src/stages/move_relative.cpp#L85) line.
I am using `MoveRelative::setDirection(const std::map<std::string, double>& joint_deltas)` with the Panda robot trying to set a relative position for the `panda_finger_joint1`. L84 on the cpp returns an index (value 14) that is out of bounds in the getJointModel, which in turn sets jm to be nullptr causing a segfault in `robot_state.enforceBounds(jm);`
Personally I fixed this by changing L85 with
```cpp
auto jm = robot_state.getRobotModel()->getJointModel(j.first);
```
but I am not sure that this is the best solution so I am not making a PR for this yet.

Issue #113, opened 2019-11-21
Demo fails for modular executable


cpt-yoshi:
The **default** modular demo works. There are 2 _SerialContainer_ added to the _Task_. But when adding 8 _SerialContainer_ it fails:
```
Task createTask() {
	Task t;
	t.stages()->setName("Reusable Containers");
	t.add(std::make_unique<stages::CurrentState>("current"));

	const std::string group = "panda_arm";
	t.add(createModule(group));
	t.add(createModule(group));
	t.add(createModule(group));
	t.add(createModule(group));
	t.add(createModule(group));
	t.add(createModule(group));
	t.add(createModule(group));
	t.add(createModule(group));

	return t;
}
```
The planning works for the first 4 _SerialContainer_ but fails at _SerialContainer_ 5:

![rviz](https://user-images.githubusercontent.com/32679594/69298175-1c5fb880-0bdb-11ea-8b0f-cfd396c673b8.png)

rhaschke:
The explanation is simple: Moving five times in a row to the left by 0.2m runs the robot into its limits. Thus the 5th attempt fails, which should be obvious from the visualization of the RobotState in rviz.

cpt-yoshi:
At the end of each subtasks the robot returns to `moveTo ready` pose. Correct me if i'm wrong but shouldn't the robot be able to do this subtask any number of times because it returns to that pose after each subtask?

rhaschke:
You are right. I missed that point. However, looking at the different solutions, you probably will observe that they are drifting into joint limits? This is because the arm is 7DoF and there is no mechanism to explicitly resolve the redundancy. I'm not sure though w/o looking into it.

PR #116, opened 2019-11-27
Forward master

rhaschke:
This set of commits comprises several minor improvements to the demos, fixing recently raised issues (#113, #114).
Most importantly, the last commit should provide compatibility to the MoveIt master branch.
If you are fine with these commits, @v4hn, please fast-forward master.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -33,6 +33,15 @@ else ()
 	set(CMAKE_CXX_STANDARD 14)
 endif ()
 
+# check for MOVEIT_MASTER
+include(CheckIncludeFileCXX)
+set(CMAKE_REQUIRED_INCLUDES ${moveit_core_INCLUDE_DIRS})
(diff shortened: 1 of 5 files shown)


v4hn:
Would it make sense to you to address `master` deprecations with the same `ifdef` style?
Afaik, this currently means relates to the two other "migration" commits [in the `moveit-master' branch](https://github.com/ros-planning/moveit_task_constructor/commits/moveit-master)

rhaschke:
I only applied the strictly required migration commit so far. Introducing ifdefs for all of them should be feasible, given the rather small amount of changes. To be honest, I didn't look into them in detail until now.

v4hn:
> To be honest, I didn't look into them in detail until now.

No need to, thanks for setting up the ifdef logic.
I mostly wanted to know whether you're ok with ifdefs to get rid of deprecation warnings.

rhaschke:
I'm fine with this, but I don't have time and the (urgent) need for it right now :wink:

PR #117, opened 2020-01-05
Only check end effector collisions for target pose in ComputeIK

henningkayser:
Fixes false positive collisions of links/objects unrelated to the (rigidly extended) end effector. A common example are collision objects or other groups that collide with other links because of the added padding like for instance objects positioned on a table.


henningkayser:
> Could you please elaborate which issue you want to tackle with this PR?
> At first sight, it doesn't make sense to me to restrict collision checking to the end-effector.
> If there are collisions with the arm, the IK solution is not feasible, is it?
> 
> Also, there is a much better approach to restrict collision planning to a subset of links:
> https://github.com/ros-planning/moveit/blob/844a0212865d66541a3d6dc6ac6abef23b0da526/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h#L275-L276

This is just the pre-check for seeing if the target pose itself is in collision (not the IK solution). The actual collision check inside the IK validation callback is not affected and would still detect collisions in the arm, therefore all IK solutions are feasible. The solved issue is simply that ComputeIK detects collisions in static links and secondary groups (i.e. other groups than the ones it should run IK for) only because of added padding. IMO ComputeIK should only "care" for collisions in the groups it was initiated with.

rhaschke:
> This is just the pre-check for seeing if the target pose itself is in collision (not the IK solution).

I missed this fact. Sorry.

> The solved issue is simply that ComputeIK detects collisions in static links and secondary groups (i.e. other groups than the ones it should run IK for) only because of added padding.

Are those unrelated links considered without padding in all remaining checks?
In this case, you are right. 

But, your implementation is still sub optimal: Instead of first checking all possible collision pairs and subsequently filtering out unrelated ones, you should restrict collision checking to the relevant pairs in first place.
Maybe `active_components_only` cannot be specified (yet) for collision checking (as an old TODO comment above the function suggests). In this case you should (further) modify the ACM to exclude unrelated link pairs.

mamoll:
@henningkayser, @rhaschke's suggestion of setting up the ACM correctly before the collision check seems reasonable to me. Can you make this change?

henningkayser:
@rhaschke @mamoll I updated this PR to only modify the ACM instead of checking the colliding links afterwards. Collisions between rigidly connected root links and scene objects are now allowed since these don't affect the IK solution as already discussed above. Also, the modifications are applied to a copied ACM so that IK is run using only the original entries.
Correct me if I'm wrong, but doesn't the collision check in `setFromIK()` still allow the disabled collisions for all group links since it's using the same sandbox_scene with the modified ACM? I didn't run a test to validate this, but looking at the code this seems to be the case and this PR would fix this.

henningkayser:
I just saw that this still doesn't cover all cases. If the colliding links are outside of the kinematic chain then the pre-check will still collide with other collision objects. My suggestion now is to really only check for collisions in the eef links in the pre-check and disable all others.

henningkayser:
@mamoll @rhaschke please re-review.
I don't think there is really a better way to fix this here but this fully covers all edge cases, including disabling collisions between static links outside of kinematic chain and scene objects while still checking for self-collisions.

henningkayser:
@rhaschke ping

core/src/stages/compute_ik.cpp
```diff
@@ -119,6 +120,16 @@ bool isTargetPoseColliding(const planning_scene::PlanningScenePtr& scene, Eigen:
...
 
+	// disable collision checking between collision objects and links not part of the rigid target group
+	const auto& world_object_ids = scene->getWorld()->getObjectIds();
+	for (const auto& robot_link : scene->getRobotModel()->getLinkModelsWithCollisionGeometry()) {
+		if (std::find(target_links.begin(), target_links.end(), robot_link) == target_links.end()) {
+			for (const auto& object_id : world_object_ids) {
+				acm.setEntry(object_id, robot_link->getName(), true);
+			}
+		}
+	}
```
rhaschke, line 131: As far as I understand, this PR wants to consider collisions of end-effector links with collision objects only, for the pre-check in `isTargetPoseColliding()`. I agree that unrelated group links (e.g. another arm) should be ignored because they will be ignored later as well.
However, collisions of links rigidly connected to the root link of the current group should be still considered, shouldn't they? That's what the code in [lines 107-121](https://github.com/ros-planning/moveit_task_constructor/pull/117/files#diff-32b368f8353f812a564f161db5e3f48fR107-R121) was doing. With your PR this code actually would become superfluous.

PR #119, opened 2020-01-09
don't install PIMPL header files

mamoll:


rhaschke:
Hm. Looks like visualization requires access to the private headers.

mamoll:
Can you think of a fix? My second commit attempted to fix this, but clearly it didn't work.

rhaschke:
I don't have time this week to look into this in detail. Need to defer to weekend...

visualization/motion_planning_tasks/src/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ target_link_libraries(${MOVEIT_LIB_NAME}
 target_include_directories(${MOVEIT_LIB_NAME}
 	PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
 	PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+	PRIVATE ${PROJECT_SOURCE_DIR}/../core/include
```
v4hn: Will this still work with build-farm releases?
I believe packages have to build independently from where they are located and whether or not others are compiled together with them.

I would have expected you add a public interface to get the required position information the visualization requires.

rhaschke: Hm. You are right, this will not work for release builds on the build farm.
Actually, I wanted to avoid rewriting the code...

v4hn: I looked into this issue some weeks ago and decided to leave it unresolved for that reason...
But I expect it would take you less time to look at the required interface changes ;)

PR #120, opened 2020-01-10
Add stage's compute time in seconds

JafarAbdi:


mamoll:
It seems a bit inelegant to have plan/runPlan, compute/runCompute methods. Perhaps it's neater to have an optional argument to plan and compute, `computeTime`, set to `false` by default, and use that to collect timing stats. If the overhead of collecting timing stats is really minimal, it could also be always on. I'll defer to the MTC maintainers.

core/src/stage.cpp
```diff
@@ -103,6 +103,7 @@ bool StagePrivate::storeSolution(const SolutionBasePtr& solution) {
 			return false;  // drop solution
 		failures_.push_back(solution);
 	} else {
+		solution->setComputeTime(compute_stop_time_ - compute_start_time_);
```
rhaschke: I don't think this can work as `compute_stop_time_` isn't yet set at this time. It's only set after `compute()` has finished, but `storeSolution()` is called as part of `compute()`.
I think you need to get time here again.

core/include/moveit/task_constructor/stage_p.h
```diff
@@ -131,6 +132,12 @@ class StagePrivate
 	bool storeSolution(const SolutionBasePtr& solution);
 	void newSolution(const SolutionBasePtr& solution);
 	bool storeFailures() const { return introspection_ != nullptr; }
+	void runCompute() {
```
mamoll: @rhaschke would it acceptable to fold the code inside runCompute() inside compute()?

core/include/moveit/task_constructor/storage.h
```diff
@@ -222,9 +223,11 @@ class SolutionBase
 
 	inline double cost() const { return cost_; }
 	void setCost(double cost);
+	inline double compute_time() const { return compute_time_; }
```
rhaschke: Remove this. See main comment for explanation.

msgs/msg/SolutionInfo.msg
```diff
@@ -4,6 +4,9 @@ uint32 id
 # associated cost
 float32 cost
 
+# time it tooks to find the solution in seconds
```
rhaschke: Remove here as well.

rhaschke:
> In a later PR it might also be nice to find how many times a new solution was computed for a given stage.

@mamoll, this is already provided in the statistics messages and shown in the rviz display: it's the sum of successful and failed solution attempts.

rhaschke:
Merged in a cleaned version in 463486257bcda2474a83985da20ea5079abff51f.
Thanks a lot for this contribution.

PR #121, opened 2020-01-10
Add public SolutionCallback interface to Task

henningkayser:
This adds a convenient callback for getting the newest solution of the current task planning instance. Using the TaskCallback would require keeping track of all solutions and compare which is the newest one as `solutions()` returns all solutions ordered by cost.

### DIFF
--- a/core/include/moveit/task_constructor/task.h
+++ b/core/include/moveit/task_constructor/task.h
@@ -105,6 +105,13 @@ public:
 	/// remove function callback
 	void erase(TaskCallbackList::const_iterator which);
 
+	typedef std::function<void(const SolutionBase& t)> SolutionCallback;
+	typedef std::list<SolutionCallback> SolutionCallbackList;
+	/// add function to be called every time a new solution is available
+	SolutionCallbackList::const_iterator addSolutionCallback(SolutionCallback&& cb);
+	/// remove function callback
+	void erase(SolutionCallbackList::const_iterator which);
+
 	/// reset all stages
 	void reset() final;
 	/// initialize all stages with given scene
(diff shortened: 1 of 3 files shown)


core/include/moveit/task_constructor/task.h
```diff
@@ -105,6 +105,13 @@ class Task : protected WrapperBase
 	/// remove function callback
 	void erase(TaskCallbackList::const_iterator which);
 
+	typedef std::function<void(const SolutionBase& t)> SolutionCallback;
```
mamoll: Small nit, because I know you're reading Effective Modern C++: Item 9. Prefer alias declarations to typedefs. :-)

core/include/moveit/task_constructor/task.h
```diff
@@ -105,6 +105,13 @@ class Task : protected WrapperBase
...
+	typedef std::list<SolutionCallback> SolutionCallbackList;
+	/// add function to be called every time a new solution is available
+	SolutionCallbackList::const_iterator addSolutionCallback(SolutionCallback&& cb);
```
mamoll: Where's the implementation of this method?

core/include/moveit/task_constructor/task.h
```diff
@@ -105,6 +105,13 @@ class Task : protected WrapperBase
...
+	/// remove function callback
+	void erase(SolutionCallbackList::const_iterator which);
```
mamoll: Where's the implementation of this method?

core/include/moveit/task_constructor/task.h
henningkayser: Well, I didn't push it...

core/include/moveit/task_constructor/task.h
henningkayser: I thought we should keep it consistent in this class

rhaschke:
@henningkayser, thanks for this fix.
As the basic `Stage` class already had an appropriate interface, which was just hidden by the protected inheritance used by `Task`, there is an even simpler solution: It suffices to expose the existing API.

PR #123, opened 2020-01-15
Add "controllers" property to Stage and pass them to PlanExecution

llach:
Users can now set a list of controller names for `Stages` (and therefore their `SubTrajectories`). This allows for solutions that switch controllers between different parts. The ability to switch controllers in MoveIt was requested [here](https://github.com/ros-planning/moveit/pull/1832).

Note: for the included demo to work, `fake_last_panda_arm_controller` and `fake_via_panda_arm_controller` need to be stored on the param server, e.g. by using a `controllers.yaml` that look like [this](https://github.com/llach/panda_moveit_config/blob/melodic-devel/config/fake_controllers.yaml#L2).

### DIFF
--- a/core/include/moveit/task_constructor/stage.h
+++ b/core/include/moveit/task_constructor/stage.h
@@ -179,6 +179,9 @@ public:
 	/// marker namespace of solution markers
 	const std::string& markerNS() { return properties().get<std::string>("marker_ns"); }
 
+	void setControllers(std::vector<std::string> controllers) { setProperty("controllers", controllers); }
+	std::vector<std::string> controllers() { return properties().get<std::vector<std::string>>("controllers"); }
+
 	/// forwarding of properties between interface states
 	void forwardProperties(const InterfaceState& source, InterfaceState& dest);
 	std::set<std::string> forwardedProperties() const {
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -201,6 +201,9 @@ bool ContainerBase::insert(Stage::pointer&& stage, int before) {
 		return false;
 	}
 
+	if (!this->controllers().empty())
+		stage->setControllers(this->controllers());
+
 	ContainerBasePrivate::const_iterator where = pimpl()->childByIndex(before, true);
 	ContainerBasePrivate::iterator it = pimpl()->children_.insert(where, std::move(stage));
 	impl->setHierarchy(this, it);
--- a/msgs/msg/SubTrajectory.msg
+++ b/msgs/msg/SubTrajectory.msg
@@ -6,3 +6,6 @@ moveit_msgs/RobotTrajectory trajectory
 
 # planning scene of end state as diff w.r.t. start state
 moveit_msgs/PlanningScene scene_diff
+
+# list of controllers to use for execution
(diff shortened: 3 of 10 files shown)


core/include/moveit/task_constructor/stage.h
```diff
@@ -179,6 +179,9 @@ class Stage
 	/// marker namespace of solution markers
 	const std::string& markerNS() { return properties().get<std::string>("marker_ns"); }
 
+	void setControllers(std::vector<std::string> controllers) { setProperty("controllers", controllers); }
```
henningkayser: I'm not sure it makes sense to offer this API at a stage level since it has nothing to do with planning. I would suggest we add generic trajectory execution properties (a separate map, message or struct) which hold information like this. This way it would also be much easier to add support for additional features and post processing steps like time parameterization or trajectory merging.

core/src/container.cpp
```diff
@@ -201,6 +201,9 @@ bool ContainerBase::insert(Stage::pointer&& stage, int before) {
 		return false;
 	}
 
+	if (!this->controllers().empty())
```
henningkayser: This could just pass those generic trajectory properties (instead of having to add additional calls for each new feature)

msgs/msg/SubTrajectory.msg
```diff
@@ -6,3 +6,6 @@ moveit_msgs/RobotTrajectory trajectory
 
 # planning scene of end state as diff w.r.t. start state
 moveit_msgs/PlanningScene scene_diff
+
+# list of controllers to use for execution
+std_msgs/String[] controller_names
```
henningkayser: I would vote for a `TrajectoryExecutionInfo` message or similar

rhaschke:
@henningkayser, I guess you misinterpreted the purpose of this PR, which Luca was preparing under my supervision. To be honest, I'm not sure what trajectory properties you are thinking about...
Could we have a phone call to clarify this?

core/include/moveit/task_constructor/stage.h
rhaschke: We want to be able to switch controllers for each and every stage. For example, for closing the grippers, we want to use a force-aware controller to stop closing at a specific grasping force...

mamoll:
@rhaschke I am interested in the related issue of controlling the time parametrization with different parameters for each stage (e.g., different velocity scaling depending on whether an object is grasped). Blending trajectories from consecutive stages without coming to a full stop becomes very tricky. It might make sense to pull the time parametrization that is now performed inside the solvers and do this only at the end when a complete solution has been found.

rhaschke:
You can easily reparameterize the full trajectory in the end if you like: Just flatten the solution into a single RobotTrajectory and apply any time parameterization you like.
However, this is unrelated to applying different time parameterizations for different stages. In this case, I think the only chance is to come to a full stop at the stage boundary.

JafarAbdi:
@rhaschke Anything blocking this? I would like to do a sync between ROS1/ROS2 and having this as part of it would be great!

sjahr:
Closing this, since changes are part of #355 now

PR #124, opened 2020-01-16
Generalised code for object_name

iamrajee:
Previous code was generating the error on changing object_name in config/panda_config.yaml file because of this bug as here we are overwriting the object_name as "object" which brought the inconsistency in code.


mamoll:
Looks good to me. Thanks!

PR #127, opened 2020-02-04
added 'PlanningScene' object in rviz by default

iamrajee:
Required to visualize robot and world in rviz while running pickplace [demo](https://ros-planning.github.io/moveit_tutorials/doc/moveit_task_constructor/moveit_task_constructor_tutorial.html).


rhaschke:
When you select individual solutions, both the robot and planning scene should be displayed.

iamrajee:
> When you select individual solutions, both the robot and planning scene should be displayed.

Thanks! Nothing major to worry about but **just wondering** if there's any way to visualize robot performing task soon after demo launch (as adding PlanningScene object in rviz would create duplicate visualization which would overlap with the visualization that appears while running individual solution.)

PR #129, opened 2020-02-12
Added C++11 flags for MOVEIT_MASTER check

progtologist:
This is required if you are compiling MTC against moveit/master in xenial (gcc-5.5 does not compile with C++11 by default)


rhaschke:
Actually, we require C++14 in all cmake files. Downgrading to C++11 doesn't make sense to me.

progtologist:
I think you missunderstood the intend, this only affects the `CHECK_INCLUDE_FILE_CXX` scope for determining if `moveit/collision_detection/collision_env.h` exists.

For more info you can check [here](https://github.com/Kitware/CMake/blob/master/Modules/CheckIncludeFileCXX.cmake#L92)

rhaschke:
Sorry, you are right. The include test already requires C++11.

PR #141, opened 2020-03-03
Avoid private API

rhaschke:
As a follow-up to #119, this PR replaces all uses of private API in the `visualization` package, making the code slightly harder to read in my opinion. @v4hn if you approve, I suggest fast-forwarding.

### DIFF
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -81,7 +81,7 @@ bool ContainerBasePrivate::traverseStages(const ContainerBase::StageCallback& pr
 
 	for (auto& stage : children_) {
 		if (!processor(*stage, cur_depth))
-			continue;
+			return false;
 		const ContainerBasePrivate* container = dynamic_cast<const ContainerBasePrivate*>(stage->pimpl());
 		if (container)
 			container->traverseStages(processor, cur_depth + 1, max_depth);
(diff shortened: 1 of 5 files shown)


core/src/container.cpp
```diff
@@ -81,7 +81,7 @@ bool ContainerBasePrivate::traverseStages(const ContainerBase::StageCallback& pr
 
 	for (auto& stage : children_) {
 		if (!processor(*stage, cur_depth))
-			continue;
+			return false;
```
rhaschke: I had to change the semantics of `traverseStages()`:
So far, returning false from the processor function, just skipped further traversing the current child (depth-wise).
Now, traversal is completely aborted, even not traversing the remaining siblings of the current child.
@v4hn: Having a single boolean return value, we cannot distinguish these two cases...

v4hn: Strictly speaking, you do not need this change, but it avoids more book-keeping in the lambdas.

Looking at the history of this, it looks like you arbitrarily defined the current semantics a long way back.
Do you have a use-case for the current semantics (continue with siblings) in mind?
I would have preferred abort-on-false to begin with.

This change deserves its own explanation in a commit though.

rhaschke: > Strictly speaking, you do not need this change, but it avoids more book-keeping in the lambdas.

I'm curious how you would solve this without the change. I guess, you wouldn't care about traversing the whole list in all cases, even if a match was already found.

Separated the traversal change into its own commit.

v4hn: > I'm curious how you would solve this without the change

If we really wouldn't want to change semantics, you could just add static boolean guards instead of returning false, but we both agree that the API is not that fixed yet.

using the traversal operator to count instead of adding (previously private) API seems like a sledgehammer solution anyway, but sure, it works.

Thank you for resolving this :+1:

PR #146, opened 2020-03-19
rework RViz panel

v4hn:
There's no need for individually resizable columns, but all numeric content should autoresize and be visible all the time.

We might want to add an option to hide the new time column in the future.

Remove intermediate AutoAdjustingTreeView layer to simplify code.
Resizing the columns did not work perfectly well before and the auto_hide_cols_ feature was not really used.

Lastly, the solution view is conceptually no tree view, so rework it to a table view.

This was a view of the old panel, without a lot of manual layout changes, illustrating issues:

mtc_gui_issues

Here is a similar view with the modifications:

mtc_gui_adjusted

### DIFF
--- a/visualization/motion_planning_tasks/src/task_list_model.cpp
+++ b/visualization/motion_planning_tasks/src/task_list_model.cpp
@@ -61,7 +61,7 @@ QVariant TaskListModel::horizontalHeader(int column, int role) {
 		case Qt::DisplayRole:
 			switch (column) {
 				case 0:
-					return tr("Name");
+					return tr("name");
 				case 1:
 					return tr(u8"✓");
 				case 2:
--- a/visualization/motion_planning_tasks/src/task_list_model.cpp
+++ b/visualization/motion_planning_tasks/src/task_list_model.cpp
@@ -114,13 +114,6 @@ QVariant BaseTaskModel::data(const QModelIndex& index, int role) const {
 	return QVariant();
 }
 
-Qt::ItemFlags BaseTaskModel::flags(const QModelIndex& index) const {
-	Qt::ItemFlags flags = QAbstractItemModel::flags(index);
-	if (index.column() == 0)
-		flags |= Qt::ItemIsEditable;  // name is editable
-	return flags;
-}
-
 QVariant BaseTaskModel::flowIcon(moveit::task_constructor::InterfaceFlags f) {
 	static const QIcon CONNECT_ICON = icons::CONNECT.icon();
 	static const QIcon FORWARD_ICON = icons::FORWARD.icon();
--- a/visualization/motion_planning_tasks/src/task_list_model.cpp
+++ b/visualization/motion_planning_tasks/src/task_list_model.cpp
@@ -81,7 +81,7 @@ QVariant TaskListModel::horizontalHeader(int column, int role) {
 			break;
 
 		case Qt::TextAlignmentRole:
-			return column == 0 ? Qt::AlignLeft : Qt::AlignRight;
+			return Qt::AlignLeft;
 
 		case Qt::ToolTipRole:
 			switch (column) {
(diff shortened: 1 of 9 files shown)


rhaschke:
I don't want to give up the AutoAdjustingTreeView. Having only a small screen, it is extremely helpful to auto-hide certain columns. From your description, I understand that you mainly complain about too narrow numeric columns, right?

v4hn:
> I don't want to give up the AutoAdjustingTreeView.

I guess you spent some work on it (or wrote it for a different context originally?).
Nonetheless
- Its usecase does not really match with the panel's requirements
  - You *don't* want to resize numeric columns by hand, ever.
  - You *don't* want to hide a column in the solutions view (right-most column). Scrolling is much more appropriate for general use and the view implements an alternative strategy to scrolling.
- it reimplements quite some standard logic that can be achieved with proper widget configuration and less intrusive code
- it does not ensure columns will stretch the whole width when columns are resized (a plain bug I guess)

> Having only a small screen, it is extremely helpful to auto-hide certain columns.

The only column you *might* want to hide is the new time column.
And even there you would want to do that in favor of the *Name* column.
Usability is ugly if you do that by changing the size of this column without affecting the numeric columns in between.
If you want to be able to hide this column I would propose a right-click menu on the header to disable to column.

Aside from that I reduced the overly big indentation of the task view tree, so the tree view is more compact now, too.


> From your description, I understand that you mainly complain about too narrow numeric columns, right?

No (see above). Sorry for not marking my issues in the first screenshot explicitly.

- solution comments were elided
- all numeric columns should auto-adjust to content
- the headers do not need to be user-resizable, aside from the tree, which stretches. It always takes a moment to get to a reasonable space allocation once one starts to adjust columns
- the tree fails to fill the widget width when users change the width of columns.

I spend a lot of thought on the current behavior and am quite happy with the patch now.

visualization/motion_planning_tasks/src/task_list_model.cpp
```diff
@@ -361,107 +362,76 @@ Qt::DropActions TaskListModel::supportedDropActions() const {
...
+		expand(index);
 }
 
-void AutoAdjustingTreeView::setModel(QAbstractItemModel* model) {
-	size_hints_.clear();
+void TaskListView::setModel(QAbstractItemModel* model) {
 	QTreeView::setModel(model);
-
-	updateGeometry();
+	if (header()->count() >= 4) {
```
henningkayser, line 377: There are some constants like this that I don't really get. Is this set to work nicely for your screen-size or panel position/width?

visualization/motion_planning_tasks/src/task_list_model.cpp
```diff
@@ -361,107 +362,76 @@ Qt::DropActions TaskListModel::supportedDropActions() const {
...
+		updateColumnWidth();
 	}
+}
 
-	QSize main_size(width, sizeHintForRow(0) * ((!preferred || !m) ? 2 : m->rowCount()));
-
-	// add size for header
-	QSize header_size(0, header()->isVisible() ? header()->height() : 0);
+void TaskListView::updateColumnWidth() {
+	for (int i = 3; i > 0; --i) {
```
henningkayser, line 391: Also here

visualization/motion_planning_tasks/src/task_list_model.cpp
henningkayser: Ok, I assume it's simply the current number of columns we have in the panels

v4hn: Yes, You don't actually have to review the code in detail.
Some feedback on the GUI would be nice though.

I'll rework part of it these days though as discussed with @rhaschke on Tuesday.

v4hn:
10 commits later. :sunglasses: 
I adjusted lots of minor stuff along the way as well as the discussed property to show/hide the time column.

I thought about adding a separate right-click menu for the header, but it's more intuitive with only the one menu, I believe.

@rhaschke Please test and review.

PR #149, opened 2020-03-21
adjust interfaces & address minor warnings

v4hn:
A small set of patches, adjusting interfaces as discussed before and cleaning up a few warnings clang comments on.

### DIFF
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -56,7 +56,7 @@ public:
 
 	/** Callback function type used by traverse functions
 	 *  The callback should return false if traversal should be stopped. */
-	typedef std::function<bool(const Stage&, int depth)> StageCallback;
+	typedef std::function<bool(const Stage&, unsigned int depth)> StageCallback;
 	/// traverse direct children of this container, calling the callback for each of them
 	bool traverseChildren(const StageCallback& processor) const;
 	/// traverse all children of this container recursively
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -62,6 +62,8 @@ public:
 	/// traverse all children of this container recursively
 	bool traverseRecursively(const StageCallback& processor) const;
 
+	void add(Stage::pointer&& stage);
+
 	virtual bool insert(Stage::pointer&& stage, int before = -1);
 	virtual bool remove(int pos);
 	virtual bool remove(Stage* child);
--- a/core/include/moveit/task_constructor/task.h
+++ b/core/include/moveit/task_constructor/task.h
@@ -92,8 +92,8 @@ public:
 	/// load robot model from given parameter
 	void loadRobotModel(const std::string& robot_description = "robot_description");
 
-	// TODO: use Stage::insert as well?
 	void add(Stage::pointer&& stage);
+	bool insert(Stage::pointer&& stage, int before = -1);
 	void clear() final;
 
 	/// enable introspection publishing for use with rviz
--- a/core/include/moveit/task_constructor/task.h
+++ b/core/include/moveit/task_constructor/task.h
@@ -82,7 +82,7 @@ public:
 	     ContainerBase::pointer&& container = std::make_unique<SerialContainer>("task pipeline"));
 	Task(Task&& other);
 	Task& operator=(Task&& other);
-	~Task();
+	~Task() override;
 
 	const std::string& id() const;
 
@@ -93,7 +93,7 @@ public:
 	void loadRobotModel(const std::string& robot_description = "robot_description");
 
 	void add(Stage::pointer&& stage);
-	bool insert(Stage::pointer&& stage, int before = -1);
+	bool insert(Stage::pointer&& stage, int before = -1) override;
 	void clear() final;
 
 	/// enable introspection publishing for use with rviz
--- a/core/include/moveit/task_constructor/task.h
+++ b/core/include/moveit/task_constructor/task.h
@@ -105,7 +105,7 @@ public:
 	/// add function to be called after each top-level iteration
 	TaskCallbackList::const_iterator addTaskCallback(TaskCallback&& cb);
 	/// remove function callback
-	void erase(TaskCallbackList::const_iterator which);
+	void eraseTaskCallback(TaskCallbackList::const_iterator which);
 
 	/// expose SolutionCallback API
 	using WrapperBase::SolutionCallback;
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -196,14 +196,11 @@ bool ContainerBase::traverseRecursively(const ContainerBase::StageCallback& proc
 
 bool ContainerBase::insert(Stage::pointer&& stage, int before) {
 	StagePrivate* impl = stage->pimpl();
-	if (impl->parent() != nullptr || !stage->solutions().empty() || !stage->failures().empty()) {
-		ROS_ERROR("cannot re-parent stage");
+	if (!impl->setParent(this))
 		return false;
-	}
-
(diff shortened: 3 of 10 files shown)


core/src/container.cpp
```diff
@@ -194,6 +194,14 @@ bool ContainerBase::traverseRecursively(const ContainerBase::StageCallback& proc
 	return pimpl()->traverseStages(processor, 1, UINT_MAX);
 }
 
+void ContainerBase::add(Stage::pointer&& stage) {
```
rhaschke: I suggest to unify the API of `add()` and `insert()`, i.e. provide an inline forward `add(stage) = insert(stage, -1)`. The error checking, should be moved into the base function `ContainerBase::insert()`.
I'm fine to unify the return value as well, i.e. switching to void and just throw?

core/src/task.cpp
```diff
@@ -195,10 +195,14 @@ void Task::loadRobotModel(const std::string& robot_description) {
 
 void Task::add(Stage::pointer&& stage) {
```
rhaschke: `Task::add` and `Task::insert` should be simple `using` forwards. There is no need to provide a new (very similar) implementation, isn't it?

core/include/moveit/task_constructor/task.h
```diff
@@ -105,7 +105,7 @@ class Task : protected WrapperBase
 	/// add function to be called after each top-level iteration
 	TaskCallbackList::const_iterator addTaskCallback(TaskCallback&& cb);
 	/// remove function callback
-	void erase(TaskCallbackList::const_iterator which);
+	void eraseTaskCallback(TaskCallbackList::const_iterator which);
```
rhaschke: IMHO, different variants of `erase` can be distinguished by their argument. But I don't mind.

core/include/moveit/task_constructor/container.h
```diff
@@ -56,7 +56,7 @@ class ContainerBase : public Stage
 
 	/** Callback function type used by traverse functions
 	 *  The callback should return false if traversal should be stopped. */
-	typedef std::function<bool(const Stage&, int depth)> StageCallback;
+	typedef std::function<bool(const Stage&, unsigned int depth)> StageCallback;
```
rhaschke: Please update all remaining usages of StageCallback functions as well, e.g.
* [`std::ostream& operator<<(std::ostream& os, const ContainerBase& container)`](https://github.com/ros-planning/moveit_task_constructor/blob/90f5d22e8371b3cd90202114f4afded70b6bc865/core/src/container.cpp#L280)
* [`Introspection::fillTaskDescription`](https://github.com/ros-planning/moveit_task_constructor/blob/90f5d22e8371b3cd90202114f4afded70b6bc865/core/src/introspection.cpp#L207)
* [`Introspection::fillTaskStatistics`](https://github.com/ros-planning/moveit_task_constructor/blob/90f5d22e8371b3cd90202114f4afded70b6bc865/core/src/introspection.cpp#L242)

core/src/container.cpp
```diff
@@ -363,7 +363,8 @@ void SerialContainer::onNewSolution(const SolutionBase& current) {
 	solution.reserve(children.size());
 	for (auto& in : incoming.solutions) {
 		for (auto& out : outgoing.solutions) {
-			InterfaceState::Priority prio(in.first.size() + 1 + out.first.size(), in.second + current.cost() + out.second);
+			InterfaceState::Priority prio(static_cast<unsigned int>(in.first.size() + 1 + out.first.size()),
```
rhaschke: Why is clang complaining here? `size()` yields an unsigned type. Is it picky about the (signed?) `1`?
What about:
```suggestion
			InterfaceState::Priority prio(in.first.size() + 1u + out.first.size(),
```

visualization/motion_planning_tasks/src/task_panel.cpp
```diff
@@ -192,7 +192,7 @@ void setExpanded(QTreeView* view, const QModelIndex& index, bool expand, int dep
 	// recursively expand all children
 	if (depth != 0) {
 		for (int row = 0, rows = index.model()->rowCount(index); row < rows; ++row)
-			setExpanded(view, index.child(row, 0), expand, depth - 1);
+			setExpanded(view, index.model()->index(row, 0, index), expand, depth - 1);
```
rhaschke: Is this an issue with a new Qt version?

core/src/task.cpp
v4hn: That's sadly not possible because of the additional Wrapper.
`Task` is actually a `WrapperBase`, but forwards `add` and `insert` to the wrapped stage.

That's why I considered, but rejected, to remove the Wrapper inheritance altogether.

core/src/container.cpp
v4hn: If we move away from the `bool` return values, `insert` does not mirror the stl container interface anymore (although `insert` is astonishingly not defined in a named concept as far as I know). Aside from being a standard container, I have never seen a reason for this function to begin with, as users should never insert stages anywhere but at the end.

Would it be sufficient for you if I add another check for a valid stage inside `insert` (which should have been there either way)?
I would like to keep the distinction between `bool` return and exceptions and also the one between "invalid pointer passed" and "insert operation failed".

core/src/container.cpp
v4hn: It's actually about `Priority` expecting an `unsigned int` but this expression computing an `unsigned long`.
The alternative would be to adjust the `Priority` class, but I don't think that's necessary.

visualization/motion_planning_tasks/src/task_panel.cpp
v4hn: Yes, [5.12 deprecated it](https://doc.qt.io/qt-5.12/qmodelindex-obsolete.html)

core/src/task.cpp
rhaschke: I see. But the new implementations should just forward to the corresponding methods of the wrapped stage then (instead of reimplementing the same logic again).

core/src/container.cpp
rhaschke: In which fashion do you want to distinguish these two error cases? Currently, they can only be distinguished by their error message, but not programmatically (both throw a runtime error).
Adding an additional check for stage validity in `insert` makes sense. Do you plan to return false or throw in this case? A reasonable distinction between insert and add could be as follows:
* `insert` is considered "low-level", not throwing, but only returning false in case of an error.
* `add` is considered "high-level", throwing (the current) exceptions on error.

> ... users should never insert stages anywhere but at the end.

I disagree. Considering preconfigured serial containers, e.g. for grasping, it is useful to be able to inject further stages, e.g. to modify the ACM.

core/src/task.cpp
v4hn: True, I addressed it.

core/src/container.cpp
v4hn: > In which fashion do you want to distinguish these two error cases?

By error message, as you say. Programmatically, there is no reason to discriminate between them as user code could check the pointer before calling. Telling the programmer his pointer is invalid does have merit though. :)

As [`setParent` already warns using `ROS_ERROR`](https://github.com/ros-planning/moveit_task_constructor/pull/149/files#diff-e6edf6dc18e6748048d781f47155c8c6R122), I moved the message to another `ROS_ERROR` log in `insert` now.

The `add`/`insert` distinction you give is what I had in mind, yes.
If you just put together a bunch of add calls, you do not want to check every single return value that way. At the same time, if you need programmatic flexibility, you can use `insert` instead.

-- 

> > ... users should never insert stages anywhere but at the end.
> 
> I disagree.

I know :grin: . I guess we just accept our difference in opinion here, as long as both use-cases are covered.

v4hn:
All feedback addressed and CI passed (though github does not show that). pushed.

PR #155, opened 2020-04-07
Allow constructing PipelinePlanner from PlanningPipelinePtr

JafarAbdi:
This adds the ability to construct PipelinePlanner from PlanningPipelinePtr, the current implementation always construct  `planner_`  by using `Task::createPlanner` [see](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/solvers/pipeline_planner.cpp#L71) but this expect the parameters to be loaded under `move_group` namespace [see](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/include/moveit/task_constructor/task.h#L78) which isn't always the case (for example when using moveit_cpp) -- another solution would be adding a new property `p.declare<std::string>("planner_ns", "move_group", "desc");` and using it in the `PipelinePlanner::init` function but doing so is resulting into a segfault with the following backtrace (it find the solution once after that it segfault), not sure if it's related to OMPL planning pipeline is already being loaded in moveit_cpp

Screenshot from 2020-04-07 22-35-08

### DIFF
--- a/core/src/solvers/pipeline_planner.cpp
+++ b/core/src/solvers/pipeline_planner.cpp
@@ -67,11 +67,17 @@ PipelinePlanner::PipelinePlanner() {
 	                    planning_pipeline::PlanningPipeline::MOTION_PLAN_REQUEST_TOPIC);
 }
 
-void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
-	planner_ = Task::createPlanner(robot_model);
+PipelinePlanner::PipelinePlanner(const planning_pipeline::PlanningPipelinePtr& planning_pipeline) : PipelinePlanner() {
+	planner_ = planning_pipeline;
+}
 
-	planner_->displayComputedMotionPlans(properties().get<bool>("display_motion_plans"));
-	planner_->publishReceivedRequests(properties().get<bool>("publish_planning_requests"));
+void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
+	if (!planner_) {
+		planner_ = Task::createPlanner(robot_model);
+
+		planner_->displayComputedMotionPlans(properties().get<bool>("display_motion_plans"));
+		planner_->publishReceivedRequests(properties().get<bool>("publish_planning_requests"));
+	}
 }
 
 void initMotionPlanRequest(moveit_msgs::MotionPlanRequest& req, const PropertyMap& p,
--- a/core/src/solvers/pipeline_planner.cpp
+++ b/core/src/solvers/pipeline_planner.cpp
@@ -74,10 +74,9 @@ PipelinePlanner::PipelinePlanner(const planning_pipeline::PlanningPipelinePtr& p
 void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
 	if (!planner_) {
 		planner_ = Task::createPlanner(robot_model);
-
-		planner_->displayComputedMotionPlans(properties().get<bool>("display_motion_plans"));
-		planner_->publishReceivedRequests(properties().get<bool>("publish_planning_requests"));
 	}
+	planner_->displayComputedMotionPlans(properties().get<bool>("display_motion_plans"));
+	planner_->publishReceivedRequests(properties().get<bool>("publish_planning_requests"));
 }
 
 void initMotionPlanRequest(moveit_msgs::MotionPlanRequest& req, const PropertyMap& p,
--- a/core/src/solvers/pipeline_planner.cpp
+++ b/core/src/solvers/pipeline_planner.cpp
@@ -74,6 +74,10 @@ PipelinePlanner::PipelinePlanner(const planning_pipeline::PlanningPipelinePtr& p
 void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
 	if (!planner_) {
 		planner_ = Task::createPlanner(robot_model);
+	} else if (robot_model != planner_->getRobotModel()) {
+		throw std::runtime_error(
+		    "The robot model of the planning pipeline isn't the same as the task's robot model -- "
+		    "use Task::setRobotModel for setting the robot model when using custom planning pipeline");
 	}
 	planner_->displayComputedMotionPlans(properties().get<bool>("display_motion_plans"));
 	planner_->publishReceivedRequests(properties().get<bool>("publish_planning_requests"));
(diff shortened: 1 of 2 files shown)


core/src/solvers/pipeline_planner.cpp
```diff
@@ -67,11 +67,17 @@ PipelinePlanner::PipelinePlanner() {
...
+}
+
 void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
-	planner_ = Task::createPlanner(robot_model);
+	if (!planner_) {
+		planner_ = Task::createPlanner(robot_model);
 
-	planner_->displayComputedMotionPlans(properties().get<bool>("display_motion_plans"));
-	planner_->publishReceivedRequests(properties().get<bool>("publish_planning_requests"));
+		planner_->displayComputedMotionPlans(properties().get<bool>("display_motion_plans"));
```
henningkayser, line 78: We should probably still apply those settings somehow

JafarAbdi: Thanks! I thought they are parameters that we should load to the parameter server

core/src/solvers/pipeline_planner.cpp
```diff
@@ -67,9 +67,14 @@ PipelinePlanner::PipelinePlanner() {
...
 
-void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
-	planner_ = Task::createPlanner(robot_model);
+PipelinePlanner::PipelinePlanner(const planning_pipeline::PlanningPipelinePtr& planning_pipeline) : PipelinePlanner() {
+	planner_ = planning_pipeline;
+}
 
+void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
+	if (!planner_) {
+		planner_ = Task::createPlanner(robot_model);
```
rhaschke, line 76: In case an existing PlanningPipeline is used, it should be checked that the RobotModelPtrs are identical and an exception should be raised if not.

JafarAbdi: Do you know how could I check if two robot_models are identical .? 

Should I add two equality operators .? 
`bool operator==(const srdf::Model& lhs, const srdf::Model& rhs)` 
`bool operator==(const urdf::ModelInterface& lhs, const urdf::ModelInterface& rhs)`

I don't think this's a good approach since ModelInterface have a lot of elements and this means adding the equality operator to each element type

Thanks!

rhaschke: My suggestion is to just test for equality of the RobotModel (shared) pointers: they should represent the very same model. Do you think this is too restrictive?

v4hn:
Looks good from my perspective, aside from Robert's request to check the model pointers.

@henningkayser Indeed this allows for a reasonable workaround. Though I still don't like the planner cache in general. :)

rhaschke:
The planner cache was needed due to a limitation of MoveIt's PipelinePlanner, which creates a service. As services need to be unique within a namespace, it was not possible to create multiple planners.

v4hn:
> The planner cache was needed due to a limitation of MoveIt's PipelinePlanner, which creates a service. As services need to be unique within a namespace, it was not possible to create multiple planners.

I am aware of this. Still there are at least two potential ways around that problem:
- the service could be optional
- PipelinePlanners can use namespaces if you are only interested in multiple independent planners (not copying planners already there)

This patch supports the latter approach, although we could also forward the namespace parameters to the PipelinePlanner constructor as Henning mentioned.

Still, this patch is quite useful if you want to reuse planners already constructed.

henningkayser:
> > The planner cache was needed due to a limitation of MoveIt's PipelinePlanner, which creates a service. As services need to be unique within a namespace, it was not possible to create multiple planners.

> I am aware of this. Still there are at least two potential ways around that problem:
- the service could be optional

Actually, not a bad idea imo

- PipelinePlanners can use namespaces if you are only interested in multiple independent planners (not copying planners already there)
This patch supports the latter approach, although we could also forward the namespace parameters to the PipelinePlanner constructor as Henning mentioned.

That's exactly what we're doing already. We are keeping a set of planning pipeline instances (initialized with MoveItCpp) and pass the desired pipeline to the planner wrapper.
However, I think it makes sense to add an overload for namespace forwarding.

core/src/solvers/pipeline_planner.cpp
JafarAbdi: Done, Thanks!

core/src/solvers/pipeline_planner.cpp
```diff
@@ -67,9 +67,18 @@ PipelinePlanner::PipelinePlanner() {
 	                    planning_pipeline::PlanningPipeline::MOTION_PLAN_REQUEST_TOPIC);
 }
 
-void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
-	planner_ = Task::createPlanner(robot_model);
+PipelinePlanner::PipelinePlanner(const planning_pipeline::PlanningPipelinePtr& planning_pipeline) : PipelinePlanner() {
+	planner_ = planning_pipeline;
```
henningkayser, line 71: Please initialize planner_ using the constructor initializer list.
Also, could you add an overloaded constructor for passing a non-default namespace?

JafarAbdi: C++ doesn't allow using member initialize list when using delegating constructor [see](https://stackoverflow.com/questions/12190051/member-initialization-while-using-delegated-constructor)

JafarAbdi: I would say creating a new property `p.declare<std::string>("planner_ns", "move_group", "desc");` and using it in the `planner_ = Task::createPlanner(robot_model, properties().get<std::string>("planner_ns"));` would replace the need to create new constructor + new member variable, but the reason why I didn't push it is because of the segfault I'm getting

JafarAbdi: Sorry I forget to answer the segfault problem, I'm looking into it now but still can't figure out why it's happening

henningkayser: With multiple constructors I would declare the properties in a separate function `declareProperties()`, so we don't need to delegate. Also, regarding the segfault, did you specify "planner_ns" or leave it empty?

JafarAbdi: Both segfault, but when I upload the parameters to different namespace and set the property to that one it would segfault in case of ompl and work normally with chomp

henningkayser: Ok, so that seems like an unrelated issue. I'm fine with merging this as is then.

PR #157, opened 2020-04-10
Make pregrasp pose optional in GenerateGraspPose

henningkayser:
The pregrasp pose is set as an empty string by default so that no pregrasp pose is applied. Instead, the stage will simply forward the current joint state of the gripper. There are several scenarios where it makes sense to let the pregrasp pose unspecified by default and to simply use the current state instead. For instance, a gripper could have different types of "open", depending on the monitored stage or could even use dynamically computed pregrasp poses matching to the size of the object. Also, an optional pregrasp pose enables to use this stage for suction grippers that don't have and predefined poses configured at all, so that it's easy to set up the same grasping pipeline for finger and suction grippers.

### DIFF
--- a/core/src/stages/generate_grasp_pose.cpp
+++ b/core/src/stages/generate_grasp_pose.cpp
@@ -54,7 +54,7 @@ GenerateGraspPose::GenerateGraspPose(const std::string& name) : GeneratePose(nam
 	p.declare<std::string>("object");
 	p.declare<double>("angle_delta", 0.1, "angular steps (rad)");
 
-	p.declare<boost::any>("pregrasp", "pregrasp posture");
+	p.declare<boost::any>("pregrasp", std::string(""), "pregrasp posture");
 	p.declare<boost::any>("grasp", "grasp posture");
 }
 
@@ -79,12 +79,15 @@ void GenerateGraspPose::init(const core::RobotModelConstPtr& robot_model) {
 	if (!robot_model->hasEndEffector(eef))
 		errors.push_back(*this, "unknown end effector: " + eef);
 	else {
-		// check availability of eef pose
-		const moveit::core::JointModelGroup* jmg = robot_model->getEndEffector(eef);
-		const std::string& name = props.get<std::string>("pregrasp");
-		std::map<std::string, double> m;
-		if (!jmg->getVariableDefaultPositions(name, m))
-			errors.push_back(*this, "unknown end effector pose: " + name);
+		// if pregrasp pose is defined, check if it's valid
+		const std::string& pregrasp_name = props.get<std::string>("pregrasp");
+		if (!pregrasp_name.empty()) {
+			// check availability of eef pose
+			std::map<std::string, double> m;
+			const moveit::core::JointModelGroup* jmg = robot_model->getEndEffector(eef);
+			if (!jmg->getVariableDefaultPositions(pregrasp_name, m))
+				errors.push_back(*this, "unknown end effector pose: " + pregrasp_name);
+		}
 	}
 
 	if (errors)
@@ -122,8 +125,11 @@ void GenerateGraspPose::compute() {
 	const std::string& eef = props.get<std::string>("eef");
 	const moveit::core::JointModelGroup* jmg = scene->getRobotModel()->getEndEffector(eef);
 
-	robot_state::RobotState& robot_state = scene->getCurrentStateNonConst();
-	robot_state.setToDefaultValues(jmg, props.get<std::string>("pregrasp"));
+	// Apply pregrasp pose if defined
+	if (!props.get<std::string>("pregrasp").empty()) {
+		robot_state::RobotState& robot_state = scene->getCurrentStateNonConst();
+		robot_state.setToDefaultValues(jmg, props.get<std::string>("pregrasp"));
+	}
 
 	geometry_msgs::PoseStamped target_pose_msg;
 	target_pose_msg.header.frame_id = props.get<std::string>("object");


core/src/stages/generate_grasp_pose.cpp
```diff
@@ -54,7 +54,7 @@ GenerateGraspPose::GenerateGraspPose(const std::string& name) : GeneratePose(nam
 	p.declare<std::string>("object");
 	p.declare<double>("angle_delta", 0.1, "angular steps (rad)");
 
-	p.declare<boost::any>("pregrasp", "pregrasp posture");
+	p.declare<boost::any>("pregrasp", std::string(""), "pregrasp posture");
```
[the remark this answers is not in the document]
JafarAbdi: Out of curiosity, is there a reason to use `boost::any` rather than `std::string` directly .? despite the only way we use the `pregrasp` property is as `std::string`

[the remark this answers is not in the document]
rhaschke: Very good point. I think, the idea was to enable some message alternatively. But you are right, currently, we only use strings.

henningkayser:
> Generally, I approve. However, I suggest keeping the unset default value,
> requiring a user to explicitly set the pregrasp posture to an empty string if he wants to have the new behavior.

I also thought about this, but I actually think that setting an empty string smells like a weird workaround and not like well-defined default behavior. I would suggest we don't check for an empty string, but rather if the property is specified at all (checking the value of boost_any). The pregrasp pose property can actually be set to a moveit_msgs::RobotState, but this is not implemented and would result in a bad_any_cast exception if specified. Right now I'm fixing this by adding conditions that check for the property type.
What is your proposed semantics for using the RobotState message as pregrasp pose? I would expect it to override the full robot state including attached collision objects, otherwise I would use a JointState message for this. Also, should we filter by eef joints or just apply the joint state values for the full robot? The same semantics needs to be discussed for the "grasp" property, but since I don't see it being used anywhere it doesn't cause any trouble.

core/src/stages/generate_grasp_pose.cpp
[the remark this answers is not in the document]
henningkayser: see comment below

rhaschke:
Hi @henningkayser,
Using `propertry().defined()` you can check for definition of a property. However, I'm not sure we should have a _default_ at all. Usually, we follow the concept that settings need to be explicitly decided for. On the other hand, if no posture is desired, doing nothing is a reasonable approach.

rhaschke:
Going for a JointState message is fine. However, only the gripper joints should be considered.

v4hn:
Wow, Easter vacations, and suddenly there is patches turning up. :smile: 

> The pregrasp pose property can actually be set to a moveit_msgs::RobotState, but this is not implemented and would result in a bad_any_cast exception if specified. Right now I'm fixing this by adding conditions that check for the property type.
> What is your proposed semantics for using the RobotState message as pregrasp pose? I would expect it to override the full robot state including attached collision objects, otherwise I would use a JointState message for this.

`RobotState` also supports `MultiDOFJointState` and up to now there was no reason to exclude those. As you implied, specifying attached objects here pretty much breaks a lot later on (probably in `Connect` stages), but sure, your attached objects might be tweezers, so we might as well apply them. *shrug*

At the same time, if you add `JointState` and we remove `RobotState` until someone actually needs it, that's fine by me.

> I also thought about this, but I actually think that setting an empty string smells like a weird workaround and not like well-defined default behavior. I would suggest we don't check for an empty string, but only if the property is specified at all (checking the value of boost_any).

I agree with Robert in that explicit is better than implicit in MTC.
Defaults might make some things look easier, but it's also easier to forget about them when you would need to change them for your task (a problem the old MoveIt code base shows a lot...).
But actually, this stage functionally requires neither `pregrasp`, nor `grasp`. Even more interesting, `grasp` does not have a default value, but it works just fine not to specify it. Robert's `SimpleGrasp` container requires both properties to be set (and probably to be of type `string` or `RobotState`?), but it's not the only user of this stage.
So I would argue:
- either the `pregrasp`/`grasp` property support could be implemented in a wrapper altogether (conceptually cleaner, but quite ugly to look at...)
- or the Stage can be refactored and a derived class could add the properties
- or, indeed, we explicitly make both optional, but also do not pass them on in the state properties if they are the default.

I'd support the second and third approach.

> Also, should we filter by eef joints or just apply the joint state values for the full robot?

This stage seeds the IK solver, so you conceptually don't know the full joint state at this point and shouldn't touch it. Filtering for joints of the eef is reasonable, possibly complaining if anything unexpected was found.

> The same semantics needs to be discussed for the "grasp" property, but since I don't see it being used anywhere it doesn't cause any trouble.

It is exposed in the interface states and read by `SimpleGrasp` too.



All in all it is rather ridiculous how much discussion there is around this trivial stage that is meant as an example...

henningkayser:
> * either the pregrasp/grasp property support could be implemented in a wrapper altogether (conceptually cleaner, but quite ugly to look at...)

Still on the menu, right now this stage really works fine for basic use cases, though

> * or the Stage can be refactored and a derived class could add the properties

Hmm, sounds a little bit like too much overhead for this feature

> * or, indeed, we explicitly make both optional, but also do not pass them on in the state properties if they are the default.

I don't really get why default behavior (that doesn't imply/change things based on some arbitrary assumptions but simply skips a feature) needs to be explicit. To me, 'pregrasp' and 'grasp' properties seem like optional features with certain use cases based on first pick pipeline designs. Couldn't we add the concept of optional properties that just don't do anything if not specified? These could also be checked on a property level without having to look into default values.

Also, I just added support for JointState messages. If you both think setting an empty string is the way to go to disable the pregrasp pose, I can add that as well.

core/src/stages/generate_grasp_pose.cpp
```diff
@@ -78,13 +78,41 @@ void GenerateGraspPose::init(const core::RobotModelConstPtr& robot_model) {
...
+		const moveit::core::JointModelGroup* eef_jmg = robot_model->getEndEffector(eef);
+		const boost::any& pregrasp_prop = props.get("pregrasp");
+		if (pregrasp_prop.type() == typeid(std::string)) {
+			// check if the specified pregrasp pose is a valid named target
+			const auto& pregrasp_name = boost::any_cast<std::string>(pregrasp_prop);
+			std::map<std::string, double> m;
+			if (!eef_jmg->getVariableDefaultPositions(pregrasp_name, m))
+				errors.push_back(*this, "pregrasp is set to unknown end effector pose: " + pregrasp_name);
+		} else if (pregrasp_prop.type() == typeid(sensor_msgs::JointState)) {
+			// check if the specified pregrasp pose is a valid named target
```
v4hn, line 92: The comment does not make sense.

core/src/stages/generate_grasp_pose.cpp
```diff
@@ -78,13 +78,41 @@ void GenerateGraspPose::init(const core::RobotModelConstPtr& robot_model) {
...
+			const auto& pregrasp_state = boost::any_cast<sensor_msgs::JointState>(pregrasp_prop);
+			if (pregrasp_state.name.size() == pregrasp_state.position.size() &&
+			    pregrasp_state.name.size() == pregrasp_state.velocity.size() &&
+			    pregrasp_state.name.size() == pregrasp_state.effort.size()) {
```
v4hn, line 96: Why would you require velocity/effort to be set?
Positions suffice, though dynamics *might* be specified to define the dynamic grasp behavior.

core/src/stages/generate_grasp_pose.cpp
```diff
@@ -78,13 +78,41 @@ void GenerateGraspPose::init(const core::RobotModelConstPtr& robot_model) {
...
+				// We only apply the joint state for for joints of the end effector
```
v4hn, line 97:
```suggestion
				// We only apply the joint state for joints of the end effector
```

core/src/stages/generate_grasp_pose.cpp
```diff
@@ -78,13 +78,41 @@ void GenerateGraspPose::init(const core::RobotModelConstPtr& robot_model) {
...
+						eef_state.name.push_back(pregrasp_state.name[i]);
+						eef_state.position.push_back(pregrasp_state.position[i]);
+						eef_state.velocity.push_back(pregrasp_state.velocity[i]);
+						eef_state.effort.push_back(pregrasp_state.effort[i]);
+					}
+				}
+				if (eef_state.name.empty())
+					errors.push_back(*this, "pregrasp JointState doesn't contain joint values for end effector group");
+				else
+					properties().set("pregrasp_state", eef_state);  // Override property with filtered joint state
```
v4hn, line 111: I'll leave it to Robert to comment on whether changing properties dynamically is a good idea.
I would consider adding a private member variable instead.

core/src/stages/generate_grasp_pose.cpp
```diff
@@ -78,13 +78,41 @@ void GenerateGraspPose::init(const core::RobotModelConstPtr& robot_model) {
...
+	else if (props.property("pregrasp").defined()) {
+		// if pregrasp pose is defined, check if it's valid
+		const moveit::core::JointModelGroup* eef_jmg = robot_model->getEndEffector(eef);
+		const boost::any& pregrasp_prop = props.get("pregrasp");
+		if (pregrasp_prop.type() == typeid(std::string)) {
+			// check if the specified pregrasp pose is a valid named target
+			const auto& pregrasp_name = boost::any_cast<std::string>(pregrasp_prop);
+			std::map<std::string, double> m;
+			if (!eef_jmg->getVariableDefaultPositions(pregrasp_name, m))
+				errors.push_back(*this, "pregrasp is set to unknown end effector pose: " + pregrasp_name);
```
v4hn, line 90: If you lookup the named target anyway, you might as well save the result for the compute call to avoid branching on the `any` again there.

core/src/stages/generate_grasp_pose.cpp
rhaschke: Fine with me.

core/src/stages/generate_grasp_pose.cpp
rhaschke: Currently, the code only uses positions. Dynamic grasp behaviour comes into play if a trajectory would be specified. That's not the case.

rhaschke:
@henningkayser: What's the status of this PR?

PR #164, opened 2020-04-25
add property to maximize clearance during IK stage

mamoll:
This can be useful when using an analytical solver. When maximize_clearance is set to true, the `ComputeIK` stage will loop through all IK solutions and assign a cost to them inversely proportional to the distance to self-collision.

Some initial feedback on whether you'd be willing to consider adding this feature would be appreciated.

### DIFF
--- a/core/src/stages/compute_ik.cpp
+++ b/core/src/stages/compute_ik.cpp
@@ -358,31 +358,27 @@ void ComputeIK::compute() {
 	IKSolutions ik_solutions;
 	auto is_valid = [sandbox_scene, ignore_collisions, min_solution_distance, maximize_clearance, &ik_solutions](
 	    robot_state::RobotState* state, const robot_model::JointModelGroup* jmg, const double* joint_positions) {
-		if (!maximize_clearance)
-			for (const auto& sol : ik_solutions) {
-				if (jmg->distance(joint_positions, sol.first.data()) < min_solution_distance)
-					return false;  // too close to already found solution
-			}
+		for (const auto& sol : ik_solutions) {
+			if (jmg->distance(joint_positions, sol.first.data()) < min_solution_distance)
+				return false;  // too close to already found solution
+		}
 		state->setJointGroupPositions(jmg, joint_positions);
 		ik_solutions.emplace_back();
 		state->copyJointGroupPositions(jmg, ik_solutions.back().first);
 		if (maximize_clearance)
 		{
-			auto& acm = sandbox_scene->getAllowedCollisionMatrix();
 			collision_detection::CollisionRequest req;
 			collision_detection::CollisionResult result;
 			req.distance = true;
-			sandbox_scene->getCollisionEnv()->checkRobotCollision(req, result, *state, acm);
+			sandbox_scene->checkSelfCollision(req, result, *state);
 			ik_solutions.back().second = result.distance;
-			ROS_DEBUG_STREAM_NAMED("ComputeIK", "clearance = " << result.distance << ", number of objects = " <<
-				sandbox_scene->getCollisionEnv()->getWorld()->size());
 			return false;
 		}
 
 		return ignore_collisions || !sandbox_scene->isStateColliding(*state, jmg->getName());
 	};
 
-	uint32_t max_ik_solutions = maximize_clearance ? 1 : props.get<uint32_t>("max_ik_solutions");
+	uint32_t max_ik_solutions = props.get<uint32_t>("max_ik_solutions");
 	bool tried_current_state_as_seed = false;
 
 	double remaining_time = timeout();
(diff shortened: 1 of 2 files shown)


mamoll:
@v4hn @rhaschke any opinion on whether this is worth adding to MTC?

core/src/stages/compute_ik.cpp
```diff
@@ -61,7 +61,8 @@ ComputeIK::ComputeIK(const std::string& name, Stage::pointer&& child) : WrapperB
 	p.declare<uint32_t>("max_ik_solutions", 1);
 	p.declare<bool>("ignore_collisions", false);
 	p.declare<double>("min_solution_distance", 0.1,
-	                  "minimum distance between seperate IK solutions for the same target");
+	                  "minimum distance between separate IK solutions for the same target");
+	p.declare<bool>("maximize_clearance", false);
```
rhaschke, line 65: Please provide a property description.

core/src/stages/compute_ik.cpp
```diff
@@ -352,17 +353,39 @@ void ComputeIK::compute() {
...
 		for (const auto& sol : ik_solutions) {
-			if (jmg->distance(joint_positions, sol.data()) < min_solution_distance)
+			if (jmg->distance(joint_positions, sol.first.data()) < min_solution_distance)
 				return false;  // too close to already found solution
 		}
 		state->setJointGroupPositions(jmg, joint_positions);
 		ik_solutions.emplace_back();
-		state->copyJointGroupPositions(jmg, ik_solutions.back());
+		state->copyJointGroupPositions(jmg, ik_solutions.back().first);
+		if (maximize_clearance) {
```
rhaschke, line 381: Do you mind moving the computation of clearance down to cost calculation?
This will simplify the code (e.g. no need to store the clearance in the solutions vector)
and facilitate later migration into a separate cost function.

core/src/stages/compute_ik.cpp
```diff
@@ -395,15 +418,22 @@ void ComputeIK::compute() {
...
 			rviz_marker_tools::appendFrame(solution.markers(), ik_pose_msg, 0.1, "ik frame");
 
-			if (succeeded && i + 1 == ik_solutions.size())
+			if (maximize_clearance) {
+				// compute cost as 1. / clearance (larger clearance is better)
+				double clearance = ik_solutions.back().second;
+				if (clearance > 0.)
+					solution.setCost(1. / (clearance + 1e-4));
+				else
+					solution.markAsFailure();
```
rhaschke, line 427: This section should be moved into the `if (succeeded && i + 1 == ik_solutions.size())` branch,
which indicates that we have found a valid IK solution at all.

v4hn:
Well, the topic still includes "WIP"...

This is a great idea and will be merged one way or another.
I won't be able to do a code review of this today, but will also try it out soon.

mamoll:
@rhaschke let me know if this is what you had in mind.

mamoll:
rebased against master

v4hn:
As a short note, I am working on a generic cost interface at the moment, which also provides a [`ClearanceCost` interface](https://github.com/v4hn/moveit_task_constructor/commit/0e7d9391849bc2bab4658188c39cd9d1ad0d03cf) term.

As this is a fully new interface though, it might take a while to settle in `master`, so I'm not sure you want to wait for that.
I should finally have some time to continue work on this next week.

core/src/stages/compute_ik.cpp
```diff
@@ -352,9 +354,23 @@ void ComputeIK::compute() {
...
+	auto acm = sandbox_scene->getAllowedCollisionMatrix();
+	if (maximize_clearance) {
+		auto& all_link_names = sandbox_state.getRobotModel()->getLinkModelNames();
+		auto& group_link_names = jmg->getLinkModelNames();
+		std::vector<std::string> non_group_links;
+
+		for (const auto& link : all_link_names)
+			if (std::find(group_link_names.begin(), group_link_names.end(), link) == group_link_names.end())
+				non_group_links.push_back(link);
+		acm.setEntry(non_group_links, non_group_links, true);
```
v4hn, line 369: If you specify and `enableGroup` the active components in your distance request further down, there should be no need for this overhead at all. The collision interface ignores all checks which do not contain at least one active component.
Weren't you aware of this or was there some other reason why you went for the explicit matrix here?

mamoll: I wasn't aware of this. Is this documented somewhere?

v4hn: Well, it's ROS, so Read Our Sources. :angel: 

https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/src/collision_common.cpp#L426

The rest of the documentation I can find only refers to "only check active components", which can be interpreted in both ways.

rhaschke:
@mamoll, I think this is superseeded by the [new CostTerm functionalies](https://github.com/ros-planning/moveit_task_constructor/blob/9f18219993fd7a4a9a9e42b7e7fe4cc6be402e17/core/include/moveit/task_constructor/cost_terms.h#L149). Can you please confirm?

v4hn:
Obsolete as @rhaschke mentioned.

PR #165, opened 2020-05-01
Add ability to add/remove objcets to/from planning scene

JafarAbdi:


JafarAbdi:
Thanks @rhaschke, I made addObject have a collision object msg as an argument -- for removeObject I think having it accept the object id is better to avoid user pass in collision object with other operation (no need for an extra checking for the operation), please let me know if you prefer having it accept collision object msg as an argument

core/include/moveit/task_constructor/stages/modify_planning_scene.h
```diff
@@ -41,6 +41,7 @@
 #include <moveit/task_constructor/stage.h>
 #include <moveit/task_constructor/properties.h>
 #include <moveit/task_constructor/type_traits.h>
+#include <geometry_msgs/PoseStamped.h>
```
rhaschke, line 44: I think this `include` has become obsolete. Maybe replace it with `moveit_msgs/CollisionObject.h`?

PR #166, opened 2020-05-15
Implement generic pose randomizer

henningkayser:
This extension of `GeneratePose` allows setting random distribution samplers for each pose dimension (x,y,z,roll,pitch.yaw). Currently, only `std::uniform_real_distribution` and `std::normal_distribution` are supported. I wasn't sure if this should maybe go into a separate stage, but the name `GeneratePose` already implies something like pose sampling IMO.


core/include/moveit/task_constructor/stages/generate_pose.h
```diff
@@ -57,10 +59,69 @@ class GeneratePose : public MonitoringGenerator
...
 protected:
 	void onNewSolution(const SolutionBase& s) override;
 	ordered<const SolutionBase*> upstream_solutions_;
+
+private:
+	/* \brief Allocate the sampler function for the specified random distribution */
+	template <template <class Realtype = double> class RandomNumberDistribution>
+	PoseDimensionSampler getPoseDimensionSampler(double /* distribution_param */) {
+		static_assert(sizeof(RandomNumberDistribution<double>) == -1, "This distribution type is not supported!");
+		throw 0;  // suppress -Wreturn-type
```
v4hn, line 114: why add a comment on -Wreturn-type if you can also specify a reasonable exception?
```suggestion
		throw std::logic_error("This distribution type is not supported.");
```

core/src/stages/generate_pose.cpp
```diff
@@ -40,13 +40,47 @@
...
 #include <rviz_marker_tools/marker_creation.h>
 
+#include <tf2/LinearMath/Quaternion.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+#include <tf2/convert.h>
+
+#include <chrono>
+
+namespace {
+// TODO(henningkayser): support user-defined random number engines
```
v4hn, line 50: Definitely needs to be provided by the user for controlled determinism.
Please address.

core/src/stages/generate_pose.cpp
```diff
@@ -40,13 +40,47 @@
...
 namespace moveit {
 namespace task_constructor {
 namespace stages {
 
 GeneratePose::GeneratePose(const std::string& name) : MonitoringGenerator(name) {
 	auto& p = properties();
 	p.declare<geometry_msgs::PoseStamped>("pose", "target pose to pass on in spawned states");
+	p.declare<size_t>("max_solutions", 20,
+	                  "limit of the number of spawned solution in case randomized sampling is enabled");
+	p.property("timeout").setDefaultValue(10.0 /* seconds */);
```
v4hn, line 64: 10 seconds to generate random poses seems ridiculous.
Limit this to max 1 second?

core/include/moveit/task_constructor/stages/generate_pose.h
```diff
@@ -57,10 +59,69 @@ class GeneratePose : public MonitoringGenerator
...
+	}
+
+	std::map<PoseDimension, PoseDimensionSampler> pose_dimension_samplers_;
 };
+template <>
+GeneratePose::PoseDimensionSampler
+GeneratePose::getPoseDimensionSampler<std::normal_distribution>(double distribution_param);
```
v4hn, line 121:
```suggestion
GeneratePose::getPoseDimensionSampler<std::normal_distribution>(double variance);
```

core/include/moveit/task_constructor/stages/generate_pose.h
```diff
@@ -57,10 +59,69 @@ class GeneratePose : public MonitoringGenerator
...
+template <>
+GeneratePose::PoseDimensionSampler
+GeneratePose::getPoseDimensionSampler<std::uniform_real_distribution>(double distribution_param);
```
v4hn, line 124:
```suggestion
GeneratePose::getPoseDimensionSampler<std::uniform_real_distribution>(double range);
```

... would imply the parameter is the *full* range, but you currently use it as half the range.

core/src/stages/generate_pose.cpp
```diff
@@ -76,15 +110,58 @@ void GeneratePose::compute() {
...
+	tf2::fromMsg(seed_pose.pose.orientation, seed_q);
+	double elapsed_time = 0.0;
+	const auto start_time = std::chrono::steady_clock::now();
+	size_t spawned_solutions = 0;
+	while (elapsed_time < timeout() && spawned_solutions++ < properties().get<size_t>("max_solutions") - 1) {
+		// Randomize pose using specified dimension samplers.
+		// RPY dimensions are using 0.0 as seed, the randomized rotation is multiplied to the target orientation.
+		double rand_roll = 0.0, rand_pitch = 0.0, rand_yaw = 0.0;
+		for (const auto& pose_dim_sampler : pose_dimension_samplers_) {
+			if (pose_dim_sampler.first == PoseDimension::X)
```
v4hn, line 143: switch?

core/src/stages/generate_pose.cpp
```diff
@@ -40,13 +40,47 @@
...
 namespace task_constructor {
 namespace stages {
 
 GeneratePose::GeneratePose(const std::string& name) : MonitoringGenerator(name) {
 	auto& p = properties();
 	p.declare<geometry_msgs::PoseStamped>("pose", "target pose to pass on in spawned states");
+	p.declare<size_t>("max_solutions", 20,
+	                  "limit of the number of spawned solution in case randomized sampling is enabled");
+	p.property("timeout").setDefaultValue(10.0 /* seconds */);
+}
```
v4hn, line 65: I guess it would make sense to set the random distributions from parameters?
Did you attempt to do that?

core/include/moveit/task_constructor/stages/generate_pose.h
henningkayser: The exception is never thrown because of the static assertion. So I thought this would be less confusing.

core/src/stages/generate_pose.cpp
henningkayser: I somehow found it weird to store functions or distribution instances as parameters. Do we plan to share them? Otherwise, we could think about just storing meta-information (like instance type + parameter value) and initialize the samplers in `compute()`

v4hn: Yes, I was thinking about distribution specifications, not about the C++ instances.

core/include/moveit/task_constructor/stages/generate_pose.h
v4hn: Personally, I find `throw 0;` more confusing, but I don't really care.

core/include/moveit/task_constructor/stages/generate_pose.h
```diff
@@ -57,10 +59,67 @@ class GeneratePose : public MonitoringGenerator
...
+	 * The distribution_param is applied to the specified distribution method while the target pose value is used as
+	 * seed.
+	 * Currently supported distributions are:
+	 * * std::uniform_real_distrubtion: distribution_param specifies the full value range around the target_pose value
+	 * * std::normal_distribution: distribution_param is used as stddev, target_pose value is mean
+	 *
+	 * Other distributions can be used by setting the PoseDimensionSampler directly using the function overload.
+	 */
+	template <template <class Realtype = double> class RandomNumberDistribution>
+	void sampleDimension(const PoseDimension pose_dimension, double distribution_param) {
```
mamoll, line 89: Would it make sense to make `distribution_param` a [parameter pack](https://en.cppreference.com/w/cpp/language/parameter_pack) to enable support for distributions with more than one parameter?

core/src/stages/generate_pose.cpp
mamoll: I imagine it'd be difficult to add support for distribution functions if the number/type of params is unknown at compile time.

v4hn: But they *are* known, because `SampleRandomPose`, or whatever Henning will call the derived class :wink: , needs to have support for it compiled in.

mamoll: Ignore my comment. This is fine as-is.

core/include/moveit/task_constructor/stages/generate_pose.h
mamoll: Ignore my comment. This is fine as-is.

core/src/stages/generate_pose.cpp
```diff
@@ -76,15 +107,65 @@ void GeneratePose::compute() {
...
+				case X:
+					target_pose.pose.position.x = pose_dim_sampler.second(seed_pose.pose.position.x);
+					break;
+				case Y:
+					target_pose.pose.position.y = pose_dim_sampler.second(seed_pose.pose.position.y);
+					break;
+				case Z:
+					target_pose.pose.position.z = pose_dim_sampler.second(seed_pose.pose.position.z);
+					break;
+				case ROLL:
```
mamoll, line 150: Note that uniform sampling of Euler angles gives you very non-uniform sampling of orientations (you oversample at the "poles"). See https://github.com/ompl/ompl/blob/master/src/ompl/util/src/RandomNumbers.cpp#L279-L285.

mamoll: Never mind. I see you can you still do what the code does I just referred to. You would have three stages, one each for RPY. Roll and yaw can be sampled uniformly. You'd just have to pass a special sampler for pitch that computes arccos of a uniformly sampled number with some offset.

core/src/stages/generate_pose.cpp
```diff
@@ -76,15 +107,65 @@ void GeneratePose::compute() {
...
+		return;
+
+	// Use target pose as seed pose
+	geometry_msgs::PoseStamped seed_pose = target_pose;
+	tf2::Quaternion seed_q, rand_q;
+	tf2::fromMsg(seed_pose.pose.orientation, seed_q);
+	double elapsed_time = 0.0;
+	const auto start_time = std::chrono::steady_clock::now();
+	size_t spawned_solutions = 0;
+	while (elapsed_time < timeout() && spawned_solutions++ < properties().get<size_t>("max_solutions") - 1) {
```
rhaschke, line 135: Simplify the logic here:
```suggestion
	while (elapsed_time < timeout() && ++spawned_solutions < properties().get<size_t>("max_solutions")) {
```
Also, you should retrieve the `max_solutions` parameter _once_ and _reuse_ it.

core/src/stages/generate_pose.cpp
rhaschke: We don't have control over the order of sampling yet (it's a map) and the already sampled RPY arguments are not passed to the sampler yet. So, I think, your original argument is still valid @mamoll.

henningkayser: The stage now supports specifying the sampling order

JafarAbdi:
@v4hn @rhaschke What's blocking this PR .? I'm working on finalizing it this week

core/src/stages/generate_random_pose.cpp
```diff
@@ -0,0 +1,162 @@
...
+	size_t spawned_solutions = 0;
+	const size_t max_solutions = properties().get<size_t>("max_solutions");
+	while (elapsed_time < timeout() && ++spawned_solutions < max_solutions) {
+		// Randomize pose using specified dimension samplers applied in the order
+		// in which they have been specified
+		sample = seed;
+		for (const auto& pose_dim_sampler : pose_dimension_samplers_) {
+			switch (pose_dim_sampler.first) {
+				case X:
+					sample.translate(Eigen::Vector3d(pose_dim_sampler.second(seed_pose.pose.position.x), 0, 0));
```
JafarAbdi, line 134: This's a bug, we either need to `sample.translation().x() = pose_dim_sampler.second(seed_pose.pose.position.x);` or `sample.translate(Eigen::Vector3d(pose_dim_sampler.second(0.0), 0, 0));` same for Y & Z

core/src/stages/generate_random_pose.cpp
```diff
@@ -0,0 +1,162 @@
...
+					break;
+				case Y:
+					sample.translate(Eigen::Vector3d(0, pose_dim_sampler.second(seed_pose.pose.position.y), 0));
```
JafarAbdi, line 137:
```suggestion
					sample.translate(Eigen::Vector3d(0, pose_dim_sampler.second(0.0), 0));
```

core/src/stages/generate_random_pose.cpp
```diff
@@ -0,0 +1,162 @@
...
+					break;
+				case Z:
+					sample.translate(Eigen::Vector3d(0, 0, pose_dim_sampler.second(seed_pose.pose.position.z)));
```
JafarAbdi, line 140:
```suggestion
					sample.translate(Eigen::Vector3d(0, 0, pose_dim_sampler.second(0.0)));
```

core/src/stages/generate_random_pose.cpp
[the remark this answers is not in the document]
henningkayser: Looks like it, but a short explanation wouldn't hurt.
So, `sample` is already set to the target pose and translating it with a distance of the target pose dimensions is redundant. I think, the first version, sample was set to identity instead of the target pose.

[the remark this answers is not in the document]
JafarAbdi: Sorry for not adding an explanation, as you said sample is already set to the target pose so this will add the specified dimension back + random value

JafarAbdi:
@rhaschke @v4hn Friendly ping. I would like to have this in the next ROS1/ROS2 sync, I've been using this feature for multiple projects for a while without any problem!

JafarAbdi (approved):

core/include/moveit/task_constructor/stages/generate_random_pose.h
```diff
@@ -0,0 +1,120 @@
...
+{
+public:
+	GenerateRandomPose(const std::string& name = "generate random pose");
+
+	bool canCompute() const override;
+	void compute() override;
+
+	/*
+	 * Function specification for pose dimension samplers.
+	 * The input paramter is the target_pose value used as seed, the returned value is the sampling result.
```
sjahr, line 59:
```suggestion
	 * The input parameter is the target_pose value used as seed, the returned value is the sampling result.
```

sjahr (approved):

PR #167, opened 2020-05-20
minor changes and clang tidy

v4hn:
A set of minor adjustments generally improving things.

@rhaschke please review. I do not expect (many) complains from you about this set of commits.


visualization/motion_planning_tasks/src/task_list_model.cpp
```diff
@@ -251,7 +251,8 @@ void TaskListModel::processTaskDescriptionMessage(const std::string& id,
 	// retrieve existing or insert new remote task for given id
 	auto it_inserted = remote_tasks_.insert(std::make_pair(id, nullptr));
 	bool created = it_inserted.second;
-	RemoteTaskModel*& remote_task = it_inserted.first->second;
+	auto& task_it = it_inserted.first;
```
rhaschke, line 254: nitpick: I think, we don't need a reference here. An iterator is essentially a pointer anyway and we don't want to risk to overwrite the iterator accidentally.

visualization/motion_planning_tasks/src/task_list_model.cpp
```diff
@@ -291,6 +291,7 @@ void TaskListModel::processTaskStatisticsMessage(const std::string& id,
                                                  const moveit_task_constructor_msgs::TaskStatistics& msg) {
 	auto it = remote_tasks_.find(id);
 	if (it == remote_tasks_.cend()) {
+		// TODO(v4hn): cache as we might receive this before the actual description
```
rhaschke, line 294: Yeah, probably we should handle this case... Have you actually seen this warning already?

core/src/task.cpp
```diff
@@ -292,7 +292,6 @@ void Task::compute() {
 
 bool Task::plan(size_t max_solutions) {
 	auto impl = pimpl();
-	reset();
```
rhaschke, line 290: Out of curiosity: What should `{ plan(1); plan(1); }` do? Are you expecting it to find more solutions?
The stages are still (re)initialized...
The ROS message you are mentioning in the log message is required to indicate to rviz that this task started over.

visualization/motion_planning_tasks/src/task_display.cpp
```diff
@@ -189,7 +189,12 @@ void TaskDisplay::changedRobotDescription() {
 }
 
 inline std::string getUniqueId(const std::string& process_id, const std::string& task_id) {
-	return process_id + "/" + task_id;
+	std::string id{ process_id };
```
rhaschke, line 192: Has this change any semantic benefit? It's just about having trailing slash or not in case of an empty task_id, isn't it?
I would prefer the simpler code over this marginal/questionable improvement.

core/include/moveit/task_constructor/container.h
```diff
@@ -49,14 +49,14 @@ class ContainerBase : public Stage
 {
 public:
 	PRIVATE_CLASS(ContainerBase)
-	typedef std::unique_ptr<ContainerBase> pointer;
+	using pointer = std::unique_ptr<ContainerBase>;
```
rhaschke, line 52: Please add modernize-use-using to the `.clang-tidy` config as well.

core/include/moveit/task_constructor/storage.h
```diff
@@ -200,6 +200,7 @@ class SolutionBase
 	inline const InterfaceState* start() const { return start_; }
 	inline const InterfaceState* end() const { return end_; }
 
+	/// connected solution segments FORWARD or BACKWARD
```
rhaschke, line 203:
```suggestion
	/// Retrieve outgoing (FORWARD) or incoming (BACKWARD) solution segments
```

visualization/motion_planning_tasks/src/task_list_model.cpp
rhaschke: By value would be (slighly) more efficient, because we don't need an additional dereference...

v4hn: No compiler would care with reasonable optimizations switched on.
By semantics, it's a shorter name for a more complex expression, not a copied value.

v4hn:
Rebased to include #173. Let's see whether codeconv gets triggered again.

visualization/motion_planning_tasks/src/task_list_model.cpp
rhaschke: I don't think so. A reference is essentially a pointer, which requires a dereference to access its contents. This cannot be optimized away by the compiler. By the benefit will we marginal in any case.

core/src/solvers/cartesian_path.cpp
```diff
@@ -80,8 +80,8 @@ bool CartesianPath::plan(const planning_scene::PlanningSceneConstPtr& from, cons
 	kinematic_constraints::KinematicConstraintSet kcs(sandbox_scene->getRobotModel());
 	kcs.add(path_constraints, sandbox_scene->getTransforms());
 
-	auto isValid = [&sandbox_scene, &kcs](moveit::core::RobotState* state, const moveit::core::JointModelGroup* jmg,
-	                                      const double* joint_positions) {
+	auto is_valid = [&sandbox_scene, &kcs](moveit::core::RobotState* state, const moveit::core::JointModelGroup* jmg,
```
rhaschke, line 83: Ah, clang-tidy considers is_valid as a (function) _variable_ here and thus enforces a lower_case name. 
I would still consider it a function. Sigh.

visualization/motion_planning_tasks/src/task_list_model.cpp
v4hn: Spot the difference: https://godbolt.org/z/C58u38
Point of fact, there is none.

core/src/solvers/cartesian_path.cpp
v4hn: The borders are quite fuzzy around that these days. :)
Technically it *is* a variable, although `auto` for closures creates a unique type signature for them...

v4hn:
I just pushed. If you notice any issue that might be related to the removed reset just file an issue.

I am convinced there is none unless people expect `plan(1); plan(1);` to reset and plan the same thing again, in which case they should explicitly `reset()`.

PR #175, opened 2020-06-12
SolutionMsg: always fill start_scene

### LINKED ISSUE #171: Collision objects missing in Solution.msg
The `moveit_task_constructor_msgs::Solution` start_scene does not include information of the collision objects.

In the pick and place demo, the information about the `table` and the `object` is missing in the `start_scene.world`:
```cpp
bool PickPlaceTask::execute() {
    ...
    moveit_task_constructor_msgs::ExecuteTaskSolutionGoal execute_goal;
    task_->solutions().front()->fillMessage(execute_goal.solution);
	
    //(gdb) print execute_goal.solution.start_scene.world
    //(gdb) print *task_->solutions().front()->start()->scene()->world_
    ...
}
```
The objects are not stored in the start_scene of the solution:
```shell
(gdb) print execute_goal.solution.start_scene.world
$9 = {collision_objects = std::vector of length 0, capacity 0, octomap = {header = {seq = 0, stamp = {<ros::TimeBase<ros::Time, ros::Duration>> = {sec = 0, nsec = 0}, <No data fields>}, frame_id = ""}, 
    origin = {position = {x = 0, y = 0, z = 0}, orientation = {x = 0, y = 0, z = 0, w = 0}}, octomap = {header = {seq = 0, stamp = {<ros::TimeBase<ros::Time, ros::Duration>> = {sec = 0, 
            nsec = 0}, <No data fields>}, frame_id = ""}, binary = 0 '\000', id = "", resolution = 0, data = std::vector of length 0, capacity 0}}}
```
The objects are stored in the planning scene:
```shell
(gdb) print *task_->solutions().front()->start()->scene()->world_
$11 = {_vptr.World = 0x7ffff4699b00 <vtable for collision_detection::World+16>, objects_ = std::map with 2 elements = {
    ["object"] = std::shared_ptr<collision_detection::World::Object> (use count 198, weak count 0) = {get() = 0x555555f93cf0}, 
    ["table"] = std::shared_ptr<collision_detection::World::Object> (use count 740, weak count 0) = {get() = 0x5555560e54d0}}, observers_ = std::vector of length 2, capacity 2 = {0x555555fe8be0, 
    0x555556039b10}}
```

I would need this information because I do some post processing on some of the sub_trajectories. For example, I would change a relative move so that it is handled by an admittance controller. This requires knowledge of the object frame (ex. a button on a device) that I want to apply a constant force on, It is a parameter needed by my controller. Note that the device will be moved in another location in the same task (Grasping pose), so I cannot know in advance the object frame.

For now, I use a relative distance wrt the end-effector to create the object frame , because there is knowledge of the robot and its attachments in the `start_scene`. A cleaner solution would be able to get the transformation directly from the `start_scene` + diff of the sub_trajectory of interest like this:

```cpp
const auto& frame = planning_scene.getFrameTransform("object");
```

rhaschke:
So far, the `start_scene` field of a `SolutionMsg` was only filled by `Introspection::fillSolution()`,
but not yet by `Task::execute()`. This PR fixes #171.

I also considered filling `start_scene` in `SolutionBase::fillMessage()`. However, for sequences, this field would have been filled multiple times then, which is not an option.

### DIFF
--- a/core/src/introspection.cpp
+++ b/core/src/introspection.cpp
@@ -133,9 +133,10 @@ void Introspection::registerSolution(const SolutionBase& s) {
 
 void Introspection::fillSolution(moveit_task_constructor_msgs::Solution& msg, const SolutionBase& s) {
 	s.fillMessage(msg, this);
+	s.start()->scene()->getPlanningSceneMsg(msg.start_scene);
+
 	msg.process_id = impl->process_id_;
 	msg.task_id = impl->task_->id();
-	s.start()->scene()->getPlanningSceneMsg(msg.start_scene);
 }
 
 void Introspection::publishSolution(const SolutionBase& s) {
--- a/core/src/task.cpp
+++ b/core/src/task.cpp
@@ -318,6 +318,8 @@ moveit_msgs::MoveItErrorCodes Task::execute(const SolutionBase& s) {
 
 	moveit_task_constructor_msgs::ExecuteTaskSolutionGoal goal;
 	s.fillMessage(goal.solution, pimpl()->introspection_.get());
+	s.start()->scene()->getPlanningSceneMsg(goal.solution.start_scene);
+
 	ac.sendGoal(goal);
 	ac.waitForResult();
 	return ac.getResult()->error_code;


henningkayser:
> I also considered filling `start_scene` in `SolutionBase::fillMessage()`. However, for sequences, this field would have been filled multiple times then, which is not an option.

Is this necessary if we cover all use cases otherwise? I would think that `SolutionSequence::fillMessage()` could also check if  `sub_solutions` is empty and populate the start scene in that case.

rhaschke:
Primarily we need to ensure that the `SubTrajectory::fillMessage()` doesn't overwrite the `start_scene` again. In my point of view, this results in overly complicated logic. For now, there are only two places where messages are filled and this PR just covers the one case which was missing to fill the start_scene.

rhaschke:
@v4hn can we merge this and thus close #171?

v4hn:
I'm actually not sure I endorse this change and did not have the time to think about it further.
If you ask to *execute* a Solution, the first scene in this message *has to be* your current scene.
I do not really see @cpt-yoshi's problem in that the controller should have access to the planning scene representation (and thus the frames he was looking for).

As all other planning scene fields in this data structure are diffs (though I believe that is not well-implemented everywhere), that means additional overhead for unrelated objects in the scene.

Would it not make sense to leave it out in general (also in `fillSolution`) and provide a `Solution::getStartScene(msg)`?

rhaschke:
I see your point. In case of execution, I agree that the scene should match the current scene.
However, I argue that the message should be consistent and complete in any case. This would allow to check on the controller side that the assumptions of the plan about the scene (i.e. the start_scene) indeed match the current scene.

v4hn:
As long as the `start_scene` message is marked as `is_diff` (w.r.t. the current world state) the message is still consistent.

If a specific controller wants to verify *the entire scene* before execution, users could just as well fill in the scene by hand, possibly as an intelligent diff to validate only what is actually needed.
Either way the underlying trajectory execution will make sure invalid joint state assumptions will not execute.

cpt-yoshi:
> As long as the `start_scene` message is marked as `is_diff` (w.r.t. the current world state) the message is still consistent.

Would the `start_scene` always be a `diff` w.r.t the current world state for a task that depends on a previous task? See example below.

### Task description
Task1 stages:
1. `CurrentState`
2. `ModifyPlanningScene`

Task2 stages:
1. `FixedState`
2. `ModifyPlanningScene`

### Planning of Task1
The starting scene is obtained from the CurrentState stage and thus is equal to the current world state.

SolutionMsg `start_scene` = Empty w.r.t. the current world state (`is_diff=true`)

### Planning of Task2
The starting scene given to the FixedState stage of Task2 is obtained from Task1 **end** scene `SolutionBase.end().scene()`.

SolutionMsg `start_scene` = ???

- Changes from Task1 w.r.t. the current world state (`is_diff=true`)
- Empty w.r.t. Task1 Solution end scene (`is_diff=true`)
- Task1 Solution end scene  (`is_diff=false`)

cpt-yoshi:
> Would it not make sense to leave it out in general (also in fillSolution) and provide a Solution::getStartScene(msg)?

This brings another question... what scene would return `Solution::getStartScene(msg)` from Task2 of my previous example?

rhaschke:
> As long as the `start_scene` message is marked as `is_diff` (w.r.t. the current world state) the message is still consistent.

As @cpt-yoshi's example shows, this is wrong: If you started planning from a state, which was not the current state (or the current state changed since you started planning), then a start_state comprising a diff only is inconsistent. For this reason, the start_state should always be complete!

v4hn:
> Would the start_scene always be a diff w.r.t the current world state for a task that depends on a previous task? See example below.

This patch adds the start scene output to the task solution *sent to execute*. In this context the start scene will always be interpreted as diff w.r.t. the current scene anyway.
As @rhaschke pointed out, the `start_scene` is already filled in for introspection (so if you look at the trajectory in the RViz panel you will see the scene from planning as expected).

I will merge this now, because I see how this can be useful with rosbag debugging.

But I might want to add an API for the previous behavior  later on (e.g., `Task::execute(const SolutionBase& s, bool populate_start_scene= true)`), to avoid sending a lot of of overhead which is never looked at.

Issue #174, opened 2020-06-12
Robot Support


danem:
I'm trying to get moveit_task_constructor to work with a Rethink Sawyer robot in simulation. As a quick test I modify the Cartesian demo to specify the correct joints and planning group for my robot. When running, the UI correctly updates with the specified task, but the robot doesn't move. I notice the following warning in the terminal

    ros.moveit_task_constructor_visualization: unknown task: 21234@ttt/

where `ttt` is my username.

And you can see the task planning state [here](https://i.imgur.com/yuY4NnO.png?1).

Are there any extra steps I need to take to get a robot that works well with moveit in general to work with MTC?

Thanks

iamrajee:
Your state `x+0.2` may not be accessible, you can click on the failed solution(denoted in red colour) to better understand why it's failing.
Also, you don't need 'Motion Planning` in the `Displays` tab while using MTC.

rhaschke:
@danem, do you start from a straight robot config? 
Possibly you just cannot move along x-axis by 20cm from this state?

danem:
@rhaschke That was a good suggestion, but I'm not sure if that was the problem. My robot is being simulated in gazebo, and I noticed the cartesian demo wasn't getting the actual robot state (eg the joints were all zeroed out). I added a `PlanningSceneMonitor` to the example and verified that the state I'm now getting is correct, but I'm still running into the same problem.

@iamrajee For future reference, I might be misunderstanding the UI here, but I don't see any additional information detailing the cause of failure. You mean the panel with the ∞ symbol, correct?

Thanks for the help and all the work you put into this.

rhaschke:
> I added a PlanningSceneMonitor to the example.

I'm not sure what exactly you mean by this. A PSM is not necessary. It suffices to start with the `CurrentState` stage like in the modular.cpp demo:
https://github.com/ros-planning/moveit_task_constructor/blob/f262d3058f78a451d122c9316efd032553f195d0/demo/src/modular.cpp#L101

> ... and verified that the state I'm now getting is correct, but I'm still running into the same problem.

So, the initial robot state shown when you select the failed "solution" of stage `x+0.2` is not straight anymore?

iamrajee:
> You mean the panel with the ∞ symbol, correct?

Yes.
Also, It's good practice to provide the code as a gist than as an attachment.

danem:
[...]

PR #176, opened 2020-06-12
MoveRelative: provide failure message

rhaschke:
As pointed out in https://github.com/ros-planning/moveit_task_constructor/issues/174#issuecomment-643458939, the failure comment was missing in case MoveRelative could not plan for the full distance and there was no min_distance specified (the default).


v4hn:
merged in 1610f5ab1dfec444d451aeec228f8ea315bd6b73

PR #181, opened 2020-06-29
Updated CMakeLists

YannickRiou:
Avoid error about moc_xxx files not found.

Without this, compiler only search in local directory instead of catkin build directory where the moc_xxx files are generated. @henningkayser repository had this correction in his CMakelists.


v4hn:
Thanks for the pull-request. I'm wondering why @henningkayser didn't file that right away...

Issue #162, opened 2020-04-21
Attach functions to stages 


dave992:
The current stages of this package seem to be focussing purely on the motion aspect of a task and do not seem to incorporate ways to call functions at the start, during, or at the end of a motion stage. Having this ability would allow for far more complex tasks, which interact with other equipment via IO, service calls or other custom functions. 

Are there any plans for incorporating function stages, of attaching functions to the current motion stages? 

Do you have any recommendations on how this could be achieved if this is currently not planned? I was thinking of writing a custom stage for this purpose. 

Thanks in advance!

v4hn:
> The current stages of this package seem to be focussing purely on the motion aspect of a task 

This is not true, as any kind of scene updates are specifically incorporated.

> and do not seem to incorporate ways to call functions at the start, during, or at the end of a motion stage.

Yes, this is because the default execution happens by sending the whole sequence of trajectories to the `move_group` node for execution. Passing functions there is not feasible.

Currently, the trivial way to add arbitrary user-code between stages is by splitting up the final solution message before sending the separate parts for execution one after the other.

A rather ugly alternative is to add auxiliary objects to the scene to indicate progress through the execution. The object appearance can be used as a trigger.

The third, in a sense "proper", way to go is to fully move away from the `move_group` node altogether and allow users to add custom code hooks between stages.
We will probably go with this option in the future, but I will not make promises on time frames.

v4hn:
One more addendum: If you want to activate / deactivate end-effectors (such as suction grippers), the preferred way should be to add custom `GripperCommand` or `FollowJointTrajectory` actions and support them through MoveIt.

This way, you can also run simulation properly and debug issues from motion traces.

dave992:
Thank you for your reply and the explanation! 
> Yes, this is because the default execution happens by sending the whole sequence of trajectories to the move_group node for execution. Passing functions there is not feasible.

That makes sense indeed. 

> Currently, the trivial way to add arbitrary user-code between stages is by splitting up the final solution message before sending the separate parts for execution one after the other.

This solution should be useful for most of the more complex tasks I envisioned. I will have a look at how this can be done. 

> The third, in a sense "proper", way to go is to fully move away from the move_group node altogether and allow users to add custom code hooks between stages. We will probably go with this option in the future, but I will not make promises on time frames.

Are there any plans or ideas on how this would be done? Just asking to see if I could be of assistance or that some of my work may be used for this in the future. 

[...]

Issue #112, opened 2019-11-20
Implement new pick&place plan capability


henningkayser:

rhaschke:
We aim to have 3 move_group capabilities:
- Pick
- Place
- Pick+Place

# Key concepts / architecture
* Pick and Place will be implemented as reusable MTC containers (improving the [existing ones](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/pick.cpp)),
* which instantiate configurable stages for a GraspProvider and PlaceProvider respectively.
* To handle bimodal grasping (i.e. choosing left or right gripper for grasping) all relevant properties should be passed via solution properties from the chosen gripper and `grasp_msg`.
* Those provider stages should be instantiated via pluginlib.

## Picking
We will provide two basic GraspProvider stages:
* One taking a predefined list of possible grasps (`grasp_msg`).
* One being an action client calling an external action server to provide grasps.
  * The server might return grasps both in its `feedback` and its `result` messages (to allow for early start of planning).
* The GraspProvider will need the following parameters:
  - eef defining (eef group, arm group, parent link in arm group)
  - a grasp_frame (PoseStamped)
  - a generic key-value parameter map that will be passed on to the action server
    - action server name (with default: `grasp_provider`)
    - config: e.g. selector for left/right/both hands (to be used by action server)
  
## Grasping Messages
Fields of the grasp_msg are not fully discussed yet. Probably we will need many fields of the existing [grasp message](http://docs.ros.org/melodic/api/moveit_msgs/html/msg/Grasp.html).
- grasp_frame as PoseStamped (relative to the `parent_link`)
- `ignore_collision_links` of the hand during finger closing

felixvd:
@henningkayser Did anything happen about this since it was posted? If yes, could you update to avoid potentially duplicating work?

rhaschke:
Unfortunately, we didn't make progress on that in Bielefeld...

v4hn:
> Unfortunately, we didn't make progress on that in Bielefeld.

Nor in Hamburg. There is always more urgent things to do, though it is still on my (long) list of todos. :/

felixvd:
We have implemented something similar that we plan to reshape into this PR soon. Just saying in case someone was planning to tackle this during the holidays.

rhaschke:
+1 I'm curious to see your approach.

PR #192, opened 2020-07-29
Add creator stage name to SolutionInfo

felixvd:
In #162 , @v4hn suggested the option of executing the solution sequentially to execute custom functions. We tried using this in an application, but when stepping through the solution's subtrajectories, we are somewhat blind as the solution does not seem to contain any semantic cues by itself*. If it does, a student of mine and I could not find them. This is a complex piece of software.

This PR extends the subtrajectory message by the name of the stage that created it so that the user can tell which part of the task they are at and execute arbitrary code (e.g. gripper open/close, set I/O, change display, call service...) when desired. Naming stages appropriately seems to be the easiest way to implement what @v4hn called "the trivial way". 

Adding code hooks may be the proper way, but as mentioned in #162 there is no time frame on this feature, and this workaround is simple, lightweight and works in Python.

This does not need to be merged, it can just be a reference for whoever wants to execute solutions this way.

*edit: Yes, you could get the `stage_id` from the `SolutionInfo` of each trajectory and search through the `TaskDescription` to retrieve the name, but it is awkward.

### DIFF
--- a/core/src/storage.cpp
+++ b/core/src/storage.cpp
@@ -146,6 +146,7 @@ void SolutionBase::fillInfo(moveit_task_constructor_msgs::SolutionInfo& info, In
 	info.comment = this->comment();
 	const Introspection* ci = introspection;
 	info.stage_id = ci ? ci->stageId(this->creator()->me()) : 0;
+	info.creator_name = this->creator()->name();
 
 	const auto& markers = this->markers();
 	info.markers.resize(markers.size());
@@ -160,7 +161,6 @@ void SubTrajectory::fillMessage(moveit_task_constructor_msgs::Solution& msg, Int
 	if (trajectory())
 		trajectory()->getRobotTrajectoryMsg(t.trajectory);
 
-	t.creator_name = this->creator()->name();
 	this->end()->scene()->getPlanningSceneDiffMsg(t.scene_diff);
 }
 
--- a/core/src/storage.cpp
+++ b/core/src/storage.cpp
@@ -160,6 +160,7 @@ void SubTrajectory::fillMessage(moveit_task_constructor_msgs::Solution& msg, Int
 	if (trajectory())
 		trajectory()->getRobotTrajectoryMsg(t.trajectory);
 
+	t.creator_name = this->creator()->name();
 	this->end()->scene()->getPlanningSceneDiffMsg(t.scene_diff);
 }
 
--- a/msgs/msg/SolutionInfo.msg
+++ b/msgs/msg/SolutionInfo.msg
@@ -10,5 +10,8 @@ string comment
 # id of stage that created this trajectory
 uint32 stage_id
 
+# name of the stage that created this trajectory
+string creator_name
+
 # markers, e.g. providing additional hints or illustrating failure
 visualization_msgs/Marker[] markers
--- a/msgs/msg/SubTrajectory.msg
+++ b/msgs/msg/SubTrajectory.msg
@@ -6,3 +6,6 @@ moveit_msgs/RobotTrajectory trajectory
 
 # planning scene of end state as diff w.r.t. start state
 moveit_msgs/PlanningScene scene_diff
+
+# Name of the stage that created this trajectory
+string creator_name
--- a/msgs/msg/SubTrajectory.msg
+++ b/msgs/msg/SubTrajectory.msg
@@ -6,6 +6,3 @@ moveit_msgs/RobotTrajectory trajectory
 
 # planning scene of end state as diff w.r.t. start state
 moveit_msgs/PlanningScene scene_diff
-
-# Name of the stage that created this trajectory
-string creator_name


v4hn:
> *edit: Yes, you could get the stage_id from the SolutionInfo of each trajectory and search through the TaskDescription to retrieve the name, but it is awkward.

Indeed, but what's wrong with extracting the `stage_id` once from the description once and using the id to find the correct solution?
The id is much more appropriate for your use-case anyway, as there is no strict unique-name assumption among the stages.

Alternatively, assuming you have the Task object in the same process you can also extract the id directly:
```
auto state = make_unique<Stage>();
auto* stage_ptr { &stage };
task.add(move(state))
task.init();
uint32_t stage_id = task.introspection().stageId(stage_ptr);
```

Should we add a simpler interface for that last line?

felixvd:
If I understand correctly, you would have to store the task description along with the solution to know which `stage_id` to look up, which to me is just another opportunity to mix up numbers. A cleartext description in the solution to describe the purpose of each trajectory seems a lot more intuitive.

Also, I imagine that users will call action servers that return the solution for a task (Pick/Place/PickPlace). Even if the task description was returned alongside the solution, searching through the message manually is a hassle. By contrast, stepping through the trajectories and checking each stage name is a breeze.

> Should we add a simpler interface for that last line?

I am not sure I even understood the `introspection` pattern, so I don't think I can answer that completely. But I'll note that we call action servers and execute the solutions in Python, so I don't think it will make a difference to us.

v4hn:
> If I understand correctly, you would have to store the task description along with the solution to know which stage_id to look up,

Yes if you want to infer the correct stage remotely from the description.
But no, if you directly get the correct id from the task - probably using the new method in fae4520 as I outlined above.

>  which to me is just another opportunity to mix up numbers. A cleartext description in the solution to describe the purpose of each trajectory seems a lot less error-prone and more intuitive.

As long as you do not use the same stage description twice that's an option.
From my perspective I think both approaches are equally error-prone, it's just a bit easier to debug discrepancies with string. :-)

For a better solution, what do you think about adding something like `Property[] execution_properties` to the SolutionInfo.
My motivation is to allow user-defined flags like "enable_suction", "switch_display_mode", "change_safety_parameter" to be specified with any SubTrajectory. That way, you can have a sane execution management that simply acts according to these flags instead of hard-coding stage names and at the same time you could have a Propagator stage that sets these properties in your Task model already.

felixvd:
> if you directly get the correct id from the task

That is assuming that you have access to the task, which you would not if you get the solution from a `move_group` capability as described in #112.

> From my perspective I think both approaches are equally error-prone, it's just a bit easier to debug discrepancies with string. :-)
> 
> For a better solution, what do you think about adding something like `Property[] execution_properties` to the SolutionInfo.
> My motivation is to allow user-defined flags like "enable_suction", "switch_display_mode", "change_safety_parameter" to be specified with any SubTrajectory. That way, you can have a sane execution management that simply acts according to these flags instead of hard-coding stage names and at the same time you could have a Propagator stage that sets these properties in your Task model already.

I like it, although the name `execution_properties` makes me expect modifying parameters or things I should not touch as a user. `custom_flags` or similar would probably be clearer.

That said, even though using the stage name to transmit flags/signals like that is definitely hacky, I think the stage name is also helpful for debugging and displaying, so including it in the solution makes sense to me.

felixvd:
> But for any kind of interleaved-execution or custom hooks, you would need [access to the Task] anyway. The only other way to support similar functionality is by additional information in the solution message, as we discuss right now.

I assumed that one should be able to execute the solution without knowing the task description, just like one can follow a manual without knowing all of the specifications and constraints that were needed to create the manual. I consider MTC a problem solving engine, so I would not expect to need the Task object (part of the solver) to execute the solution, and it seems reasonable to add information to the solution message. That's just my interpretation though!

> The concept of properties is very general in MTC and you already modify the properties of stages to setup your problem. We could also go for `Property[] external` or `external_properties`, I would just like to avoid ambiguities with the Properties that are used to generate a specific solution (we do not export those), although maybe we should.

I'm not sure which properties are used where, and which properties are used to generate a "specific" solution rather than the returned message.

> Properties can be boolean flags, but providing arbitrary data, e.g. to specify the pressure for each element in a suction array, is supported here, so "flags" is no good name.

Then maybe `custom_properties`? I know that technically everything can be customized, but the point is to convey that it's a setting the user should configure. When I read through MTC, I found it difficult to know what is essential to the framework (= "better not touch"), and what can/should be changed when making a new task. I would hope a name like this makes it easier to grok.

> But then we would add redundant information to the messages and it also makes sense to avoid that. The information is now cleanly split between task description, statistical data and actual solutions and we decided not to use the stage name as linking key. If you want to debug MTC plans, you should save the description/statistics topics anyway or debug from within the C++ code in which you run the Task.

True, it would be somewhat redundant, which is why I had doubts about merging this. I guess it's a question of how self-contained you want the solution to be, and what users should be able to do with the solution message alone.

For completeness: a use case outside of debugging is to display the stage name to show what the robot is currently doing.

AndyZe:
@v4hn I've tried something very similar to what you suggested above. You suggested this:

```
auto state = make_unique<Stage>();
auto* stage_ptr { &stage };
task.add(move(state))
task.init();
uint32_t stage_id = task.introspection().stageId(stage_ptr);
```

I tried this:

```
moveit::task_constructor::Stage* open_fingers_stage_;
...
// Note, I use moveit::task_constructor::stages::MoveTo rather than Stage
auto stage = std::make_unique<MoveTo>(
  PICK_STAGE_MAP.at(PickStageEnum::OPEN_FINGERS),
  sampling_planner);
stage->setGroup(finger_group_);
stage->setGoal(arm_prefix_ + "fingers_open");
open_fingers_stage_ = &stage;
task_.add(std::move(stage));
```

It doesn't compile because of the type difference.

```
 error: cannot convert ‘std::unique_ptr<moveit::task_constructor::stages::MoveTo, std::default_delete<moveit::task_constructor::stages::MoveTo> >*’ to ‘moveit::task_constructor::Stage*’ in assignment
  130 |     open_fingers_stage_ = &stage;
```

I believe the MoveTo type is correct, and I need the Stage* type to be able to retrieve the ID. So... I do not see a way around this.

rhaschke:
Obviously, stage is a `unique_ptr`. To get a raw address, you need to use `stage.get()`!

AndyZe:
Thank you Robert!

I'm getting close but there is one more issue:

`uint32_t stage_id = task_.introspection().stageId(open_fingers_stage_);`

error: ‘uint32_t moveit::task_constructor::Introspection::stageId(const moveit::task_constructor::Stage*)’ is private within this context

AndyZe:
Would you accept a PR to make `stageId()` public?

rhaschke:
The `const` version of `stageId()` _is_ public. Please don't hijack PRs / issues with unrelated questions, but open a separate thread instead.

AndyZe:
Alright. I would not say the discussion is off-topic because it is related to accomplishing the purpose of this PR. It would be useful to anybody reading this PR.

Also, you said:  

> Obviously, stage is a unique_ptr

What isn't so obvious is that MoveTo inherits from Stage. If I read the [header file](https://github.com/ros-planning/moveit_task_constructor/blob/ros2/core/include/moveit/task_constructor/stages/move_to.h), MoveTo inherits from `PropagatingEitherWay`. I guess there are several layers of inheritance, which can make it difficult to find things.

rhaschke:
Closing in favour of #194

PR #193, opened 2020-07-29
Another batch of minor patches

v4hn:
most of them are cosmetic, f07f16d is the biggest actual change, adding an interface to access the stage id.

@felixvd

### DIFF
--- a/core/include/moveit/task_constructor/solvers/joint_interpolation.h
+++ b/core/include/moveit/task_constructor/solvers/joint_interpolation.h
@@ -47,7 +47,10 @@ namespace solvers {
 
 MOVEIT_CLASS_FORWARD(JointInterpolationPlanner)
 
-/** Use MoveIt's PlanningPipeline to plan a trajectory between to scenes */
+/** Interpolate a trajectory between states in joint space
+ *
+ * Fails if direct joint space interpolation fails.
+ */
 class JointInterpolationPlanner : public PlannerInterface
 {
 public:
--- a/core/src/stage.cpp
+++ b/core/src/stage.cpp
@@ -253,7 +253,7 @@ void Stage::reset() {
 	impl->properties_.reset();
 }
 
-void Stage::init(const moveit::core::RobotModelConstPtr& robot_model) {
+void Stage::init(const moveit::core::RobotModelConstPtr& /* robot_model */) {
 	// init properties once from parent
 	auto impl = pimpl();
 	impl->properties_.reset();
--- a/core/src/stage.cpp
+++ b/core/src/stage.cpp
@@ -283,6 +283,12 @@ void Stage::setName(const std::string& name) {
 	pimpl_->name_ = name;
 }
 
+uint32_t Stage::introspectionId() const {
+	if (!pimpl_->introspection_)
+		throw std::runtime_error("Task is not initialized yet or Introspection was disabled.");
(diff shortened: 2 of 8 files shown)


v4hn:
@rhaschke Feel free to review at any time. I will just keep pushing here and probably create the next request once this batch is merged.

core/src/stage.cpp
```diff
@@ -283,6 +283,13 @@ void Stage::setName(const std::string& name) {
 	pimpl_->name_ = name;
 }
 
+uint32_t Stage::introspectionId() const {
+	if (!pimpl_->introspection_)
+		throw std::runtime_error(
+		    "Introspection not available. Initialize the full Task before calling introspectionId()");
+	return const_cast<const moveit::task_constructor::Introspection*>(pimpl_->introspection_)->stageId(this);
```
rhaschke, line 290: Why is it necessary to explicitly cast to `const` here? As the method is `const`, all members of this should be const as well, shouldn't they? Or do you want to make it explicit?
```suggestion
	return pimpl_->introspection_->stageId(this);
```

core/src/stage.cpp
```diff
@@ -283,6 +283,13 @@ void Stage::setName(const std::string& name) {
 	pimpl_->name_ = name;
 }
 
+uint32_t Stage::introspectionId() const {
+	if (!pimpl_->introspection_)
+		throw std::runtime_error(
+		    "Introspection not available. Initialize the full Task before calling introspectionId()");
```
rhaschke, line 289: It's not about initializing the task, but about enabling introspection.
```suggestion
		throw std::runtime_error("Introspection wasn't enabled.");
```

core/include/moveit/task_constructor/solvers/joint_interpolation.h
```diff
@@ -47,7 +47,10 @@ namespace solvers {
 
 MOVEIT_CLASS_FORWARD(JointInterpolationPlanner)
 
-/** Use MoveIt's PlanningPipeline to plan a trajectory between to scenes */
+/** Interpolate a trajectory between states
```
rhaschke, line 50:
```suggestion
/** Interpolate a trajectory between states in joint space
```

core/src/stage.cpp
v4hn: No, it's actually about initialization because the `introspection_` instance is only passed to the stages during init and the stages register with the instance only once the first description is published via `fillTaskDescription`.

core/src/stage.cpp
v4hn: It actually fails to compile here without the const cast because calling `stageId` on a non-const introspection requires private visibility (in contrast to the const version) and gcc's implementation rejects the call because of that.

The private method would actually *name* the stage (give it a new number) without checking whether it already has one.

v4hn:
addressed all comments.

core/src/stage.cpp
rhaschke: You are right. I updated the error text proposal.

core/src/stage.cpp
rhaschke: Yes, I was assuming that the introspection instance should be const because of the main method being const...

core/src/stage.cpp
v4hn: I pushed a rephrased version.

core/src/stage.cpp
v4hn: No, it's just the `introspection_` pointer that's const.

PR #194, opened 2020-07-30
add external properties to SolutionInfo

v4hn:
This interface can be customized by the user to state when/where additional execution hooks should run.

This is an initial thread to discuss/extend this idea as discussed in #192

### DIFF
--- a/msgs/msg/SolutionInfo.msg
+++ b/msgs/msg/SolutionInfo.msg
@@ -12,3 +12,7 @@ uint32 stage_id
 
 # markers, e.g. providing additional hints or illustrating failure
 visualization_msgs/Marker[] markers
+
+# user-defined values that can provide additional information,
+# e.g., whether to switch a suction gripper on or off or when to call some other execution interface
+Property[] external


v4hn:
@felixvd

msgs/msg/SolutionInfo.msg
```diff
@@ -12,3 +12,7 @@ uint32 stage_id
 
 # markers, e.g. providing additional hints or illustrating failure
 visualization_msgs/Marker[] markers
+
+# user-defined values that can provide additional information,
+# e.g., whether to switch a suction gripper on or off or when to call some other execution interface
+Property[] external
```
felixvd, line 18: The description and the structure looks good to me, I'm only thinking about the name, as mentioned in the other thread. For the examples we discussed, this might be a good one though.

rhaschke: I don't like the name `external` as well. Here, I would just name them `properties` as well.
```suggestion
Property[] properties
```

v4hn: If we go for the generic name, we will also include all INTERFACE-configured stage properties & interface state properties used for computing the solution.
These are currently unobservable from the outside.

I'm looking into these aspects right now anyway, so if you agree @rhaschke I can prepare a first patch these days.

rhaschke: I don't see why the generic name implies that all the internal properties should be exposed as well.
Internally (in the MTC source code), I'm fine to name those properties `external_props`.

> I'm looking into these aspects right now anyway, so if you agree @rhaschke I can prepare a first patch these days.

Yes, in general, I like the idea of exported solution properties - just to introduce yet another name suggestion :wink:

v4hn:
> Note that a Property's value field is a string comprising a serialized version of an arbitrary property.

That's exactly why I would prefer it over some string indicator fields for this.

> The serialization/deserialization methods are not (yet) exposed and might be subject to changes. 

Same for everything else in MTC.

> Hence, the Property type might not be the best choice to implement the behavior here.
> At least we would need to expose the deserialization methods...

From a pragmatic point of view it works just fine with `std::string` properties, because their serialization is the string itself. In general though, we definitely want to expose deserialization functions if we want users to actually use these properties. Currently, the deserialization is mixed in with the rviz/visualization code...

msgs/msg/SolutionInfo.msg
v4hn: > I don't see why the generic name implies that all the internal properties should be exposed as well.

It's a matter of perspective of course, but for me "the properties of the solution" clearly refers to, e.g., the approach angle used to compute the particular grasp candidate and not so much to the question whether or not a suction gripper should be switched on in this solution.

This is why I proposed the name `external`. I would be fine with mixing these two concepts though if we introduce a single "solution properties" list.

rhaschke: I think we agree that the semantics of the properties exposed in the message is fully up to the user. Hence, arguing with semantics is not meaningful here.
I see two types of solution properties (from an abstract level):
- properties required during computation
- properties exposed to other processes, e.g. execution via the solution message

Of course, one could publish both types of properties via this interface, but I don't see a reason to do so.
How we name those two types of properties is a different story. We should first decide on the intended semantics.
As the user (from the external view) only sees the exposed properties, I would name them in the message as `properties`.

v4hn: I mostly agree with your assessment and wholeheartedly disagree with your conclusion. :smile: 

> I think we agree that the semantics of the properties exposed in the message is fully up to the user.

That was the original idea for this thread. However, after you started arguing for the generic name I noticed the missing message interface to expose INTERFACE properties (and other internal variables the stage might consider as properties).
All static properties are of course part of the `description` topic, but these properties obviously can't be specified there.
The properties list in the panel lists empty entries for dynamic properties, even when you browse through the individual solutions. From a GUI perspective it would help a lot if these missing entries are read from the `Solution`.
So if we introduce a generic `properties` field here, I believe it should definitely include them.

At the same time, having two separate `Properties[]` fields in the `SolutionInfo` seems overly redundant and thus I'm open to merge the two concepts.
As a consequence, users can simply set own properties in a stage to expose them with the solution which avoids the need for another interface.

felixvd: I'm not sure what you mean by "interface properties", but from a user perspective, including the intended usage in the field's name would be helpful, even if it does not cover every use case. I was thinking of something specific rather than general, but `exported_properties` might not be bad.

rhaschke: > So if we introduce a generic properties field here, I believe it should definitely include them.

Ok, I agree that publishing those properties might be useful for debugging and thus we should publish both types of properties. If we do so, a generic name like `properties` is perfect.

core/include/moveit/task_constructor/properties/serialize.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+	/** ROS messages are serialized using the provided binary encoding */
+	template <typename Q = T>
+	static std::enable_if_t<IsROSMessage<Q>, std::string> serialize(const boost::any& value) {
+		Q message{ boost::any_cast<Q>(value) };
+		uint32_t serial_size = ros::serialization::serializationLength(message);
+		std::unique_ptr<uint8_t[]> buffer(new uint8_t[serial_size]);
+		ros::serialization::OStream stream(buffer.get(), serial_size);
+		ros::serialization::serialize(stream, message);
+		return reinterpret_cast<const char*>(base64::encode(buffer.get(), serial_size).c_str());
```
rhaschke, line 139: Why do you think we need Base64 encoding here? A `std::string` can contain any `char` - in contrast to `char*`, which cannot include `\0` as this already encodes the end of the string. However, a std::string's length is encoded separately.

rhaschke (commented):
Just a few remarks... We definitely should have a phone call to sync a little bit on our activities :wink:

core/include/moveit/task_constructor/properties/serialize.hpp
rhaschke: My new commits prove that this isn't required.

rhaschke:
I was long thinking about why I decided against your proposed solution when I worked on Property serialization.
Eventually, I remembered and found the corresponding PR https://github.com/ros-planning/moveit_task_constructor/pull/72: 
For serialization/deserialization to work, they need to access registered methods. This is simple on the serialization end, because we know the types to serialize:
https://github.com/ros-planning/moveit_task_constructor/blob/0aff5d56dd45b91b7c6c7c63c9674eec84e49aed/core/include/moveit/task_constructor/properties.h#L266-L270
However, on the deserialization end, we typically don't know the type in advance. Thus I decided to go for the generic YAML output generated by `operator<<` for ROS messages. This has two advantages:
* The serialized output is human-readable, e.g. when intercepting `rostopic`.
* rviz, the main deserialization site, can use a [generic deserializer](https://github.com/ros-planning/moveit_task_constructor/blob/5dacc85ff98b48a76586924eed76e8300f55a9ef/visualization/motion_planning_tasks/properties/property_from_yaml.cpp#L308-L315) to display values, even if no type-specific one is registered. Using ROS message serialization we lose this capability: Note, how `ik_frame` is displayed before and after this PR:

|old|new|
|--------------------|---------------------|
|old | new |

rhaschke:
We discussed pros and cons of the old and the proposed approach in a phone conference. This is the summary:
| | old, YAML-based serialization | new, binary serialization |
|----------|---------|-----------|
| **pro** | human-readable, e.g. with rostopic | exact saving/loading of floating point numbers |
| deserialization for rviz | existing C++ YAML parser | python deserialize ➔ python YAML ➔ existing C++ YAML parser |
| msg deserialization | YAML ➔ python msg ➔ python serialize ➔ C++ deserialize | ROS deserialize |
| open issues | YAML from ROS msgs partly broken (requires fix in gencpp or workarounds) | will require `uint8[]` representation |

v4hn (commented):
thanks for the summary. here are some more comments I wrote before the talk reviewing your changes.

core/include/moveit/task_constructor/properties/serialize.hpp
```diff
@@ -57,30 +58,24 @@ namespace task_constructor {
...
+struct IsStreamDeserializable : std::false_type
 {};
 
 template <typename T>
-struct hasExtractionOperator<T, decltype(std::declval<std::istream&>() >> std::declval<T&>())> : std::true_type
+struct IsStreamDeserializable<T, decltype(std::declval<std::istream&>() >> std::declval<T&>())> : std::true_type
 {};
 
-template <typename T>
-constexpr bool hasExtractionOperator_v = hasExtractionOperator<T>::value;
```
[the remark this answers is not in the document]
rhaschke: I don't care too much. But use:
`template <typename T> Has*Operator_v = Has*Operator<T>::value;`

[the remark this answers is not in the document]
v4hn: I honestly have no idea why I didn't... Probably a left-over because I pushed incomplete code...

core/include/moveit/task_constructor/properties.h
```diff
@@ -101,18 +131,39 @@ class Property
 	/// reset to default value (which can be empty)
 	void reset();
 
-	/// the current value defined or will the fallback be used?
-	inline bool defined() const { return !value_.empty(); }
+	/// is a value defined?
+	inline bool defined() const { return !(value_.empty() && default_.empty()); }
```
rhaschke, line 135: Here you change the semantics of `defined()`!
I guess you want to distinguish `isDefined()` from `isSet() == hasCurrentValue()`?
Let's discuss the naming over phone...

core/include/moveit/task_constructor/properties.h
```diff
@@ -101,18 +131,39 @@ class Property
...
 	inline const boost::any& value() const { return value_.empty() ? default_ : value_; }
-	inline boost::any& value() { return value_.empty() ? default_ : value_; }
+
+	/// get typed value of property. Throws bad_any_cast on type mismatch, undefined if !defined().
+	template <typename T>
+	inline const T& value() const {
+		const boost::any& v{ value() };
+		if (v.empty())
+			throw Property::undefined();
+		return boost::any_cast<const T&>(value());
```
rhaschke, line 149:
```suggestion
		return boost::any_cast<const T&>(v);
```
See also my comment on `PropertyMap::get()`.

rhaschke (commented):
I'm not sure about the status of this PR. @v4hn, maybe we should schedule another phone call?

captain-yoshi:
What is the status of this PR ? Would be really useful. I think a lot of people got their own way of dealing with this.

---
BTW this PR needs this patch,
```cpp
void SolutionBase::fillInfo(moveit_task_constructor_msgs::SolutionInfo& info, Introspection* introspection) const {
    info.id = introspection ? introspection->solutionId(*this) : 0;                                                
    info.cost = this->cost();                                                                                      
    info.comment = this->comment();                                                                                
    const Introspection* ci = introspection;                                                                       
    info.stage_id = ci ? ci->stageId(this->creator()) : 0;                                                         
    
    // got to add this
    creator_->properties().fillMsgs(info.properties);
                                                              
    const auto& markers = this->markers();                                                                         
    info.markers.resize(markers.size());                                                                           
    std::copy(markers.begin(), markers.end(), info.markers.begin());                                               
}                                                                                                                 
```
to actually print the properties.
```cpp
  for (const auto& solution : task->solutions())
  {
    moveit_task_constructor_msgs::Solution solution_msg;
    solution->fillMessage(solution_msg, &task->introspection());

    for (const auto& sub_traj : solution_msg.sub_trajectory)
    {
      std::cout << "==========" << std::endl;
      std::cout << "stage_id: " << sub_traj.info.id << std::endl;

      auto props = sub_traj.info.properties;

      for (moveit_task_constructor_msgs::Property& p : props)
      {
        std::cout << "name: " << p.name << std::endl;
        std::cout << "type: " << p.type << std::endl;
        std::cout << "val : " << p.value << std::endl;
        std::cout << "---------" << std::endl;
      }
    }
  }


==========
stage_id: 1
name: forwarded_properties
type: St3setINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESt4lessIS5_ESaIS5_EE
val : 
---------
name: hey
type: NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
val : 
---------
name: ignore_collisions
type: b
val : 0
---------
name: marker_ns
type: NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
val : current state
---------
name: timeout
type: d
val : 
---------

...
```

captain-yoshi:
Heum... if a add a property in the `open hand` stage it get's copied to the `approach object` stage, `lift object` and so on. This is not good for user's who wants to add additional hooks.

Maybe the API should have the properties from stages separated from properties that users could use...

captain-yoshi:
@v4hn Trying to serialize the ros msg `control_msgs::GripperCommand`. Do you have an idea why it does not serialize ? Other ros msg are ok... 

```cpp
control_msgs::GripperCommand controller_goal;
controller_goal.max_effort = 100;
controller_goal.position = robot_state.joint_state.position.front();

stage->setProperty("point", point);
stage->setProperty("controller_goal", controller_goal);

// Prints
name: point
type: geometry_msgs/Vector3Stamped
val : thermal_cycler���Q��?
--------
name: controller_goal
type: 
val : 
```

core/include/moveit/task_constructor/properties.h
```diff
@@ -288,18 +239,22 @@ class PropertyMap
 	Property& property(const std::string& name);
 	const Property& property(const std::string& name) const { return const_cast<PropertyMap*>(this)->property(name); }
 
+	void fillMsgs(std::vector<moveit_task_constructor_msgs::Property>& msgs) const;
+	void fromMsgs(std::vector<moveit_task_constructor_msgs::Property>& msgs);
```
captain-yoshi, line 243:
```suggestion
	void fromMsgs(const std::vector<moveit_task_constructor_msgs::Property>& msgs);
```

core/src/properties.cpp
```diff
@@ -212,6 +214,22 @@ Property& PropertyMap::property(const std::string& name) {
...
+void PropertyMap::fillMsgs(std::vector<moveit_task_constructor_msgs::Property>& msgs) const {
+	msgs.reserve(size());
+	for (const auto& pair : *this) {
+		msgs.emplace_back();
+		msgs.back().name = pair.first;
+		pair.second.fillMsg(msgs.back());
+	}
+}
+
+void PropertyMap::fromMsgs(std::vector<moveit_task_constructor_msgs::Property>& msgs) {
```
captain-yoshi, line 226:
```suggestion
void PropertyMap::fromMsgs(const std::vector<moveit_task_constructor_msgs::Property>& msgs) {
```

visualization/motion_planning_tasks/properties/property_factory.cpp
```diff
@@ -43,40 +43,57 @@
 #include <rviz/properties/property_tree_model.h>
 #include <rviz/properties/string_property.h>
 #include <rviz/properties/float_property.h>
+#include <rviz/properties/vector_property.h>
```
captain-yoshi, line 46: Compile error on Noetic:
```bash
Compilation error:
In file included from /opt/ros/noetic/include/rviz/properties/vector_property.h:32,
                 from /home/captain-yoshi/ws/ros/mimik_ws/src/moveit_task_constructor/visualization/motion_planning_tasks/properties/property_factory.cpp:46:
/opt/ros/noetic/include/rviz/ogre_helpers/ogre_vector.h:2:10: fatal error: OgrePrerequisites.h: No such file or directory
    2 | #include <OgrePrerequisites.h>
    ```

**Comment by captain-yoshi** [comment]
Trying to recreate a `MoveRelative` from the Property.msg throws an error:
```cpp
stage->properties().fromMsgs(property_msg);

terminate called after throwing an instance of 'moveit::task_constructor::Property::type_error'
  what():  Property: type v doesn't match property's declared type St3setINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESt4lessIS5_ESaIS5_EE

```
The culprit is the `forwarded_properties` which is of type `std::set<std::string>`. This type should be added to the serialization.

Test to add: 
1- Create a MoveRelative stage1
2- stage->fillMsgs(property_msg)
3- Create another MoveRelative stage2 (should also work on stage1) 
4- stage2->fromMsgs(property_msg)

Output of Property.msg
```yaml
name: controller_id
type: NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
value: joint_position_controller
description:

name: direction
type: geometry_msgs/Vector3Stamped
value: pcr����Mb@�
description: motion specification

name: direction_offset
type: geometry_msgs/Pose
value: <�O��nr?��MbX�?-C��6J?�;f���?�;f���?
description: frame offset wrt. the direction frame_id

name: forwarded_properties
type: St3setINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESt4lessIS5_ESaIS5_EE
value: 
description: set of interface properties to forward

name: group
type: NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
value: eef_pipette
description: name of planning group

name: ik_frame
type: geometry_msgs/PoseStamped
value: 
        pipette_tcp�?
description: frame to be moved in Cartesian direction

name: marker_ns
type: NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
setPlanningSceneDiffMsg: value = Move Relative Above Bottom Container
description: marker namespace

name: max_distance
type: d
value: 0
description: maximum distance to move

name: min_distance
type: d
value: -1
description: minimum distance to move

name: path_constraints
type: moveit_msgs/Constraints
value: 
description: constraints to maintain during trajectory

name: timeout
type: d
value: 1
description: timeout per run (s)

name: trajectory_execution_info
type: moveit_task_constructor_msgs/TrajectoryExecutionInfo
value: 
description: settings used when executing the trajectory
```

captain-yoshi:
I have added tests and fixes in this [branch](https://github.com/captain-yoshi/moveit_task_constructor/tree/external-properties). This now works for my use case.

I would also prefer the YAML serialization over binary. You can even have exact saving/loading of floating point numbers, at the cost of size increase.

PR #196, opened 2020-08-11
New grasp pose generator stage

bostoncleek:
This stage uses action messages to communicate with an action server external to MTC. The action server can use machine learning pipelines to sample grasps given a point cloud/depth image. This removes complex dependencies from MTC. The stage sends a request for grasps. If a time out is reached or no grasps are found then planning fails.

This stage uses a template parameter which is the action message and inherits from an action base class. This allows other stages to inherit these properties and use action messages.


mamoll:
The intent is to add a base class that can be used to implement support for GPD and Dex-Net. These grasp synthesis pipelines have a whole slew of dependencies, so the GPD/Dex-Net support would likely live in a separate packages (say, `moveit_task_constructor_gpd`, and `moveit_task_constructor_dexnet`).

mamoll:
@rhaschke @v4hn perhaps it's helpful to see how this code can be used. See https://github.com/PickNikRobotics/deep_grasp_demo/blob/master/src/gpd_pick_place_task.cpp for an example.

core/include/moveit/task_constructor/stages/action_base.h
```diff
@@ -0,0 +1,124 @@
+/*********************************************************************
+ * BSD 3-Clause License
+ *
+ * Copyright (c) 2020 PickNik LLC.
```
davetcoleman, line 4:
```suggestion
 * Copyright (c) 2020 PickNik Inc.
```

core/include/moveit/task_constructor/stages/action_base.h
```diff
@@ -0,0 +1,130 @@
...
+};
+
+template <class ActionSpec>
+ActionBase<ActionSpec>::ActionBase(const std::string& action_name, bool spin_thread, double goal_timeout,
+                                   double server_timeout)
+  : action_name_(action_name), server_timeout_(server_timeout), goal_timeout_(goal_timeout) {
+	clientPtr_.reset(new actionlib::SimpleActionClient<ActionSpec>(nh_, action_name_, spin_thread));
+
+	if (server_timeout_ < 0.0 || goal_timeout < 0.0) {
+		ROS_WARN("Timeouts cannot be negative");
```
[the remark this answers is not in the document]
JStech: I see now that a negative `goal_timeout_` will be treated like 0. The `server_timeout_` just gets passed to the `SimpleActionClient` and it's not clear what would happen if it's negative. Just for clarity, though, I'd clamp the values to be non-negative.

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
```diff
@@ -0,0 +1,310 @@
...
+	double timeout_time = ros::Time::now().toSec() + ActionBaseT::goal_timeout_;
+
+	while (ActionBaseT::nh_.ok()) {
+		ros::spinOnce();
+
+		if (found_candidates_) {
+			break;
+		}
+
+		if (ros::Time::now().toSec() > timeout_time && monitor_timeout) {
```
JStech, line 164: Nit: I'd reverse the two conditions, so that if `monitor_timeout` is false, the other check is shortcutted entirely.

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
```diff
@@ -0,0 +1,310 @@
...
+	ROS_INFO_NAMED(LOGNAME, "Generate grasp goal now active");
+	found_candidates_ = false;
+}
+
+template <class ActionSpec>
+void DeepGraspPose<ActionSpec>::feedbackCallback(const FeedbackConstPtr& feedback) {
+	// each candidate should have a cost
+	if (feedback->grasp_candidates.size() != feedback->costs.size()) {
+		ROS_ERROR_NAMED(LOGNAME, "Invalid input: each grasp candidate needs an associated cost");
+	} else {
```
JStech, line 185: Rather than the extra `else` block, you could just return here:
```suggestion
          return;
	}
```
(Then also remove the `}`.)

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
```diff
@@ -0,0 +1,310 @@
...
+	}
+	planning_scene::PlanningScenePtr scene = upstream_solutions_.pop()->end()->scene()->diff();
+
+	// set end effector pose
+	const auto& props = properties();
+	const std::string& eef = props.get<std::string>("eef");
+	const moveit::core::JointModelGroup* jmg = scene->getRobotModel()->getEndEffector(eef);
+
+	robot_state::RobotState& robot_state = scene->getCurrentStateNonConst();
+	robot_state.setToDefaultValues(jmg, props.get<std::string>("pregrasp"));
```
JStech, line 259: Should these pose names be set as `constexpr`s at the top? Not saying it's necessarily better that way, but suggesting that it be considered.

core/include/moveit/task_constructor/stages/action_base.h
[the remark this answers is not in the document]
bostoncleek:
```suggestion
	// Negative timeouts are set to zero
	server_timeout_ = server_timeout_ < std::numeric_limits<double>::epsilon() ? 0.0 : server_timeout_;
	goal_timeout_ = goal_timeout_ < std::numeric_limits<double>::epsilon() ? 0.0 : goal_timeout_;
```

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
bostoncleek: Good catch I will move the found candidates check to an else if after the timeout

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
bostoncleek: The reason I don't return here is I thought the error and info messages provide the user with feedback to let them know if the action message fields where completed properly.

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
bostoncleek: I am not sure about this one

core/include/moveit/task_constructor/stages/action_base.h
bostoncleek: The LLC was updated to INC in both header files

bostoncleek:
It might be useful to check out the implementation and documentations for the [demos](https://github.com/PickNikRobotics/deep_grasp_demo).

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
JStech: I don't mean you should remove the error, but you can return immediately after it, instead of using an `else` block. It would be functionally equivalent to what you have implemented.
```
template <class ActionSpec>
void DeepGraspPose<ActionSpec>::feedbackCallback(const FeedbackConstPtr& feedback) {
	// each candidate should have a cost

	if (feedback->grasp_candidates.size() != feedback->costs.size()) {
		ROS_ERROR_NAMED(LOGNAME, "Invalid input: each grasp candidate needs an associated cost");
		return;
	}
	ROS_INFO_NAMED(LOGNAME, "Grasp generated feedback received %lu candidates: ", feedback->grasp_candidates.size());

	grasp_candidates_.resize(feedback->grasp_candidates.size());
	costs_.resize(feedback->costs.size());

	grasp_candidates_ = feedback->grasp_candidates;
	costs_ = feedback->costs;

	found_candidates_ = true;
}
```

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
JStech: No worries--I'm sure one of the friendly local maintainers will let you know if they want it changed.

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
bostoncleek: I see that does simplify it. I will make that change.

bostoncleek:
Can I get a review please? @v4hn @rhaschke

lianghongzhuo:
Looks good for me when I test this with our ur5 robot.

JStech:
I'm using this for a project and I was surprised to see it's still not merged. @v4hn and @rhaschke could you please take a look at this PR?

JStech:
I had to rebase this to get it to build on Noetic (it was a very easy rebase). [Here](https://github.com/JStech/moveit_task_constructor/tree/pr-deep_grasp_stage) is my branch.

rhaschke:
Thanks  @JStech for providing the rebased branch link. Currently, there is so much other work, that I cannot catch up with all PR reviews. As this one is a larger one, it will probably take some more time... Please be patient (and continue using your own branch meanwhile).

lianghongzhuo:
@JStech 's rebase link works for me, (ubuntu 18, ros melodic with moveit master branch)

davetcoleman:
@bostoncleek think this could get merged now?

bostoncleek:
> @bostoncleek think this could get merged now?

This was ready to be merged back on August 11th. It is a matter of the maintainers pressing the merge button!

Looks like it does need a rebase though. I will squash and rebase today.

davetcoleman (approved):
I did a very quick skim, lgtm

bostoncleek:
> It's not just merge, but also review (and possibly fixup). I'm sorry for not getting around to do it. There is always more pressing issues and often two crying kids in the background since Corona homeoffice. Thanks for rebasing in advance!

Any change requests will be fixed so please make a comment on what you would like.

v4hn:
As I said, I did not have the time to look through this in detail, and I'm sorry for that.

My biggest concern, as I can recall from discussion with @henningkayser and @lianghongzhuo was that the use-case to receive candidate poses in an any-time fashion through feedback and spawn solutions based on them continuously should be well supported, even if you did not use that approach in your project.
Such an approach might be valuable for projects [like this](https://github.com/ros-planning/moveit_task_constructor/issues/239) too.

bostoncleek:
> As I said, I did not have the time to look through this in detail, and I'm sorry for that.
> 
> My biggest concern, as I can recall from discussion with @henningkayser and @lianghongzhuo was that the use-case to receive candidate poses in an any-time fashion through feedback and spawn solutions based on them continuously should be well supported, even if you did not use that approach in your project.
> Such an approach might be valuable for projects [like this](https://github.com/ros-planning/moveit_task_constructor/issues/239) too.

Don't have the time? ... its World MoveIt Day! of course you have the time. Grow the project a little don't be afraid to let others make a contribution towards your work. 

If it doesn't happen today I will see you on World MoveIt Day 2022!

davetcoleman:
@bostoncleek could you fix the conflict?
@JafarAbdi would be a good candidate to review and merge this, given his extensive work with MTC

rhaschke:
I shortly looked into this PR and I was surprised by the many unrelated commits. @bostoncleek, it would be great if you could rebase and cleanup the history.

bostoncleek:
@rhaschke @JafarAbdi  I squashed the commits and now it should be easy for you to review. Conflict is also resolved.

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
```diff
@@ -0,0 +1,299 @@
...
+ * @brief Generate grasp candidates using deep learning approaches
+ * @param ActionSpec - action message (action message name + "ACTION")
+ * @details Interfaces with a deep learning based grasp library using a action client
+ */
+template <class ActionSpec>
+class DeepGraspPose : public GeneratePose, ActionBase<ActionSpec>
+{
+private:
+	typedef ActionBase<ActionSpec> ActionBaseT;
+	ACTION_DEFINITION(ActionSpec);
```
rhaschke, line 66: Isn't this redundant, because the action types are pulled in from `ActionBase<ActionSpec>`?

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
```diff
@@ -0,0 +1,299 @@
...
+  : GeneratePose(stage_name), ActionBaseT(action_name, false, goal_timeout, server_timeout), found_candidates_(false) {
+	auto& p = properties();
+	p.declare<std::string>("eef", "name of end-effector");
+	p.declare<std::string>("object");
+	p.declare<boost::any>("pregrasp", "pregrasp posture");
+	p.declare<boost::any>("grasp", "grasp posture");
+
+	ROS_INFO_NAMED(LOGNAME, "Waiting for connection to grasp generation action server...");
+	ActionBaseT::clientPtr_->waitForServer(ros::Duration(ActionBaseT::server_timeout_));
+	ROS_INFO_NAMED(LOGNAME, "Connected to server");
```
rhaschke, line 129: Couldn't this be part of `ActionBase`?

core/include/moveit/task_constructor/stages/deep_grasp_pose.h
```diff
@@ -0,0 +1,299 @@
...
+
+template <class ActionSpec>
+void DeepGraspPose<ActionSpec>::feedbackCallback(const FeedbackConstPtr& feedback) {
+	// each candidate should have a cost
+	if (feedback->grasp_candidates.size() != feedback->costs.size()) {
+		ROS_ERROR_NAMED(LOGNAME, "Invalid input: each grasp candidate needs an associated cost");
+	} else {
+		ROS_INFO_NAMED(LOGNAME, "Grasp generated feedback received %lu candidates: ", feedback->grasp_candidates.size());
+
+		grasp_candidates_.resize(feedback->grasp_candidates.size());
```
rhaschke, line 182: There might be still unprocessed `grasp_candidates_`! Hence, you don't want to override the existing list.
Consider, for example, a grasp provider sending several feedback messages in a row...
Also, you need to mutex-protect the access to this list.

rhaschke (changes requested):
Thanks for cleaning up this PR. Now, it's much easier to review. I like some of the ideas herein, namely:
- asynchronously receiving results
- using a modular class structure combining GeneratePose and ActionBase

But, the current implementation has also several design flaws:
- You should resort to the existing [`grasping_msgs::GraspPlanningAction`](https://github.com/mikeferguson/grasping_msgs/blob/ros1/action/GraspPlanning.action), which is used by several existing grasp providers. This way, the new stage automatically generalizes to all of them.
- As a consequence, the stage should be named `GraspProvider` (or the like). There is nothing specific of this stage regarding deep learning.
- The stage depends on a concrete message type via duck-typing, e.g. when preparing the goal or receiving results. So, practically, I don't see an advantage of templating the class. Please implement it just for [`grasping_msgs::GraspPlanningAction`](https://github.com/mikeferguson/grasping_msgs/blob/ros1/action/GraspPlanning.action).
- Doing so, also will allow incorporating other grasp-specific parameters like pregrasp and grasp posture provided by the action server. Please have a look at my own implementation of a GraspProvider stage [here](https://github.com/ubi-agni/mtc_demos/blob/iros19/stages/grasp_provider.cpp). Combining both approaches will yield a great solution.
- You don't yet handle multiple feeback messages and thread safety.

bostoncleek:
@rhaschke Thank you for the feedback. I will work on the change requests.

v4hn:
Thanks for the review @rhaschke !
@bostoncleek I will wait for his feedback to be addressed and review afterwards.
If I'll have further requests for changes then I will patch them myself and you got my official permission to kick me if I did not do it by the maintainer meeting in April. :muscle: (Of course assuming you addressed the feedback before :smile: )

bostoncleek:
@v4hn  Thank you for offering to patch future requests yourself. I am addressing the larger issues now so hopefully it will be set to merge in soon.

bostoncleek:
@v4hn I have addressed @rhaschke requests. Please let me know if there are other patches that need to be made.

lianghongzhuo:
any updates?

lianghongzhuo:
@bostoncleek Could you rebase this PR? There is some upstream change that makes this PR unable to compile.
Like add this https://github.com/ros-planning/moveit_task_constructor/commit/437cc550f2a68895de56f025e2507770bff2faca

bostoncleek:
@lianghongzhuo yes I will rebase

bostoncleek:
This is failing because of clang-tidy.

```
error: unknown warning option '-Wno-unused-but-set-parameter'; did you mean '-Wno-unused-parameter'? [clang-diagnostic-unknown-warning-option]
```

I don't see where `add_compile_options()` are set using a call to `-Wno`. 

Can a maintainer provide help?

v4hn:
#274 (which was not included here yet) recently made the CI configuration quite a bit stricter and addressed this particular error.
Let's see whether new ones pop up now. :-)

core/include/moveit/task_constructor/stages/grasp_provider.h
```diff
@@ -0,0 +1,107 @@
...
+	void setGraspPose(const std::string& grasp) { properties().set("grasp", grasp); }
+	void setGraspPose(const moveit_msgs::RobotState& grasp) { properties().set("grasp", grasp); }
+
+protected:
+	void onNewSolution(const SolutionBase& s) override;
+
+private:
+	std::mutex grasp_mutex_;  // Protects grasp candidates
+	std::atomic_bool found_candidates_;  // Flag indicates the discovery of grasps
+	std::vector<moveit_msgs::Grasp> grasp_candidates_;  // Grasp Candidates
```
bostoncleek, line 103: @JafarAbdi Can you take a look a these members? The intent is to address Robert's comment about thread safety and multiple feedback messages. I don't think the mutex in the case can cause deadlock. Please let me know if you see anything wrong with this.

bostoncleek:
WMD 2022! 

I made some changes and ready for a review. I would really like to get this PR merged. @rhaschke @v4hn @JafarAbdi

davetcoleman:
Thanks for revitalizing this! 

Possible reviewers: @henningkayser @mamoll @vatanaksoytezer

bostoncleek:
> Thanks for revitalizing this!
> 
> Possible reviewers: @henningkayser @mamoll @vatanaksoytezer

If this can get in I will update the tutorial as well.

msgs/CMakeLists.txt
```diff
@@ -33,6 +33,6 @@ generate_messages(DEPENDENCIES ${MSG_DEPS})
 
 catkin_package(
 	CATKIN_DEPENDS
-		message_runtime
```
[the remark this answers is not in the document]
bostoncleek: Spaces are now reverted.

core/src/stages/grasp_provider.cpp
```diff
@@ -0,0 +1,206 @@
...
+	const moveit::core::JointModelGroup* jmg = scene->getRobotModel()->getEndEffector(eef);
+
+	robot_state::RobotState& robot_state = scene->getCurrentStateNonConst();
+	robot_state.setToDefaultValues(jmg, props.get<std::string>("pregrasp"));
+
+	// compose/send goal
+	composeGoal();
+
+	// monitor feedback/results
+	// blocking function untill timeout reached or results received
```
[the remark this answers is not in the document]
bostoncleek: Removed typo.

core/src/stages/grasp_provider.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+void GraspProvider::doneCallback(const actionlib::SimpleClientGoalState& state,
+                                 const grasping_msgs::GraspPlanningResultConstPtr& result) {
+	if (state == actionlib::SimpleClientGoalState::SUCCEEDED) {
+		ROS_DEBUG_STREAM_NAMED(LOGNAME, "Found grasp candidates");
+	} else {
+		ROS_ERROR_NAMED(LOGNAME, "No grasp candidates found (state): %s", clientPtr_->getState().toString().c_str());
+	}
+}
```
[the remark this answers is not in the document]
bostoncleek: The reason the mutex is here is to handle multiple feedback messages. After discussing with @rhaschke he was concerned there could be many feedback messages containing grasp samples. The flag is used to indicate that at least one message was recieved.

core/src/stages/action_base.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+ActionBase::ActionBase(const std::string& action_name, bool spin_thread)
+  : action_name_(action_name), server_timeout_(0.0), goal_timeout_(0.0) {
+	clientPtr_.reset(
+	    new actionlib::SimpleActionClient<grasping_msgs::GraspPlanningAction>(nh_, action_name_, spin_thread));
+
+	ROS_DEBUG_STREAM_NAMED(LOGNAME, "Waiting for connection to grasp generation action server...");
+	clientPtr_->waitForServer(ros::Duration(server_timeout_));
+	ROS_DEBUG_STREAM_NAMED(LOGNAME, "Connected to server");
+}
```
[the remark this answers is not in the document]
bostoncleek: Thanks for catching this, much cleaner.

core/src/stages/grasp_provider.cpp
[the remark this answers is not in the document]
JafarAbdi: But as soon as we enter the if statement `if (monitorGoal()) {` in ` void GraspProvider::compute() {` we will not update the `grasp_candidates_` since the mutex is locked, and it will get unlocked just before we exit the function which mean next compute will send a new goal, and we only get the chance to handle the first feedback, right?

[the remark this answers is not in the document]
bostoncleek: I see that would simply this by removing the mutex. Yeah this does only handle the first feedback message if `compute()`.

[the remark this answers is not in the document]
bostoncleek: I would like it to be able to handle multiple feedback messages. Do you see a way that could work?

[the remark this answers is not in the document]
bostoncleek: What do you think about storing all the grasps from the feedback and results fields into a container. Then after the done callback flags success the `compute()` function operates over all those grasp candidates?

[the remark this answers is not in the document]
JafarAbdi: See https://github.com/bostoncleek/moveit_task_constructor/pull/2 for a proposal :grimacing:

PR #197, opened 2020-08-19
Minor Changes

v4hn:
Here comes the next badge of minor changes.

The visibility changes are required for some tests I started writing for the CostTerm interface.
The rest shouldn't be controversial.


core/src/stages/move_to.cpp
```diff
@@ -120,6 +120,16 @@ bool MoveTo::getJointStateGoal(const boost::any& goal, const moveit::core::Joint
 	} catch (const boost::bad_any_cast&) {
 	}
 
+	try {
+		const std::map<std::string, double>& joint_map = boost::any_cast<std::map<std::string, double>>(goal);
+		for (const auto& joint : joint_map) {
+			state.setVariablePosition(joint.first, joint.second);
```
rhaschke, line 126: Should we check for the validity of the joint name, i.e. whether it is part of JMG?

core/include/moveit/task_constructor/stages/predicate_filter.h
```diff
@@ -65,6 +65,7 @@ class PredicateFilter : public WrapperBase
 	void onNewSolution(const SolutionBase& s) override;
 
 	void setPredicate(const Predicate& p) { setProperty("predicate", p); }
+	void setIgnoreFilter(bool ignore) { setProperty("ignore_filter", ignore); }
```
rhaschke, line 68: Please add a description for this property. It's semantics is not clear to me yet.
Why would you like to ignore the filter at all? If so, you can simply remove the wrapper stage, can't you?

core/include/moveit/task_constructor/storage.h
```diff
@@ -100,6 +100,9 @@ class InterfaceState
 	InterfaceState(const planning_scene::PlanningScenePtr& ps);
 	InterfaceState(const planning_scene::PlanningSceneConstPtr& ps);
 
+	// provide an initial priority for the state
+	InterfaceState(const planning_scene::PlanningSceneConstPtr& ps, const Priority& p);
```
rhaschke, line 104: I'm not sure we should publicly expose such an interface. There is no reason a user will ever create an interface with an initialized priority. If you need this for testing, provide a utility class there?

rhaschke (commented):
I'm not sure about exposing WrappedSolution (3ff0da84363fb3905c6ab073ea2cd989fe36f7d5, e17e9f572be3a8663e3de5ae7c916db2e56cb7e8).
This is a class only used internally by `ParallelContainerBase`.
Note: This PR breaks Travis - there is a failing test.

core/include/moveit/task_constructor/stages/predicate_filter.h
v4hn: it's just simpler to set the property than change the structure of the task if you optionally want to filter, e.g., by additional path constraints.

core/include/moveit/task_constructor/storage.h
v4hn: The priority property is actually private, so a utility class does not help.
If it makes you feel any better, I could also change the visibility to protected instead and inherit just to set an initial priority.

It feels rather absurd to me to have a data structure around that cannot *at all* initialize one of its members though (priority can only be set by `SolutionBase` and updated by `Interface` via `friend` definitions).

Eventually, I only need this to get around [a somewhat questionable assert](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/storage.cpp#L106). If the `InterfaceState` does not have any `SolutionBase` objects connected *it can't* be initialized with the current code.
I would be just as happy to remove the assert. ;)

v4hn:
> Note: This PR breaks Travis - there is a failing test.

Fixed via #199 and #200 .

> I'm not sure about exposing WrappedSolution (3ff0da8, e17e9f5). This is a class only used internally by ParallelContainerBase.

That's not "only". All `WrapperBase` stages use it and it *is* exposed to the user via `solutions()`.
If you want to access a specific part of the wrapped solution, this is not possible without these patches.

core/include/moveit/task_constructor/stages/predicate_filter.h
rhaschke: OK. A comment would be nice anyway.

core/include/moveit/task_constructor/storage.h
rhaschke: > It feels rather absurd to me to have a data structure around that cannot at all initialize one of its members though (priority can only be set by SolutionBase and updated by Interface via friend definitions).

I just want to avoid the risk of users messing around with the priority, e.g. wrongly initializing it...

> Eventually, I only need this to get around a somewhat questionable assert.

Maybe the assert is wrong. Do you have example code that violates the assert?

> If the InterfaceState does not have any SolutionBase objects connected it can't be initialized with the current code.

The default constructor should initialize the priority to `(0, 0.0)`.

rhaschke:
> > I'm not sure about exposing WrappedSolution ([3ff0da8](https://github.com/ros-planning/moveit_task_constructor/commit/3ff0da84363fb3905c6ab073ea2cd989fe36f7d5), [e17e9f5](https://github.com/ros-planning/moveit_task_constructor/commit/e17e9f572be3a8663e3de5ae7c916db2e56cb7e8)). This is a class only used internally by ParallelContainerBase.
> 
> That's not "only". All `WrapperBase` stages use it and it _is_ exposed to the user via `solutions()`.
> If you want to access a specific part of the wrapped solution, this is not possible without these patches.

Serializing the solution should work, as WrappedSolution is still a SolutionBase. Do you want to mess around with the internals?

core/include/moveit/task_constructor/stages/predicate_filter.h
v4hn: I already added a property description.

core/include/moveit/task_constructor/storage.h
v4hn: > Maybe the assert is wrong.

It's not *wrong* in itself, but inconvenient. It is unreachable without the private interfaces.
But using the private interfaces, it is still not possible to fulfill it with a custom `InterfaceState` without a full Solution attached.
It's currently not possible to add a standalone `InterfaceState` to an `Interface` (e.g. via
`InterfaceState is{...}; state->pimpl()->starts().add(is);`

If you deem it better, I can also remove the assert (and thus allow new `InterfaceState` entries with zero priority).
That's still very restrictive because it does not allow to sort multiple new entries by priority, but it fulfills my need.

v4hn:
> Serializing the solution should work, as WrappedSolution is still a SolutionBase. Do you want to mess around with the internals?

Sure `fillMessage` works as it should. But if I only want to know the cost of the SubTrajectory of a particular substage, I don't need to serialize the whole solution if I can also walk through the "internals" (which are public interfaces except for `WrappedSolution` before this patch.

PR #206, opened 2020-09-17
tf2 compatibility for Noetic

rhaschke:
Just a simple compile-time switch to handle removed rviz' `getTFClient()` in Noetic.

### DIFF
--- a/visualization/visualization_tools/src/marker_visualization.cpp
+++ b/visualization/visualization_tools/src/marker_visualization.cpp
@@ -72,13 +72,8 @@ bool MarkerVisualization::createMarkers(rviz::DisplayContext* context, Ogre::Sce
 	Ogre::Vector3 pos;
 
 	try {
-		std::string error_msg;
 #ifdef RVIZ_TF1
 		tf::TransformListener* tf = context->getFrameManager()->getTFClient();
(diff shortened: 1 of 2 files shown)


visualization/visualization_tools/src/marker_visualization.cpp
```diff
@@ -69,8 +72,9 @@ bool MarkerVisualization::createMarkers(rviz::DisplayContext* context, Ogre::Sce
 	Ogre::Vector3 pos;
 
 	try {
-		tf::TransformListener* tf = context->getFrameManager()->getTFClient();
 		std::string error_msg;
+#ifdef RVIZ_TF1
+		tf::TransformListener* tf = context->getFrameManager()->getTFClient();
 		if (!tf->canTransform(planning_frame_, fixed_frame, ros::Time(), &error_msg)) {
```
v4hn, line 78: Could you remove this `canTransform` check? Github forbids me to add a suggestion for it...
The block is not there for the new TF2 version, and it's not needed either because exceptions are caught below (`tf::*Exception` are plain typedefs for their `tf2::` counterparts)

v4hn (approved):
Please consider my inline comment before merging.

These API changes are awfully annoying... The blocks look way too alike.

PR #207, opened 2020-09-22
Several minor fixes

rhaschke:
This PR includes some simple improvements and fixes, mainly in the `Merger` stage (separate branch).
Please "merge" via fast-forward.

### DIFF
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -204,20 +204,21 @@ bool ContainerBase::insert(Stage::pointer&& stage, int before) {
 	return true;
 }
 
-bool ContainerBasePrivate::remove(ContainerBasePrivate::const_iterator pos) {
+Stage::pointer ContainerBasePrivate::remove(ContainerBasePrivate::const_iterator pos) {
 	if (pos == children_.end())
-		return false;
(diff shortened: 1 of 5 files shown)


core/src/container.cpp
```diff
@@ -825,6 +825,8 @@ void MergerPrivate::onNewPropagateSolution(const SolutionBase& s) {
 		ROS_ERROR_NAMED("Merger", "Only simple trajectories are supported");
 		return;
 	}
+	if (s.isFailure() || !trajectory->trajectory())  // ignore failure solutions
+		return;
```
v4hn, line 830: It's quite frustrating that Github forbids suggestions that include collapsed context lines...

If the container really can't work with empty trajectories, could you add `!trajectory->trajectory()` as an alternative condition to the check right above? If the merge receives a SubTrajectory without a real trajectory this should trigger an error and not be discarded silently.

v4hn (approved):
Have a look at my inline comment. Aside from that feel free to push the commits to master (I don't see the need for the merge commit, but at least the grouped commits make sense for one of the two merged branches. :)

PR #213, opened 2020-10-23
Task: Add ability to set timeout

JafarAbdi:
I have found that when one of the stages is struggling to find a solution the task can get stuck planning forever, so in this PR I exposed the setTimeout function and made the task fail if it exceeded the time


mamoll:
This seems generally useful. While MTC heuristically guides the search for a solution, there is often still a combinatorial explosion of candidate solutions that MTC will consider. It is often not desirable to exhaustively try all candidate solutions.

rhaschke (approved):
Nice contribution. Computing the remaining time incrementally is somewhat hard to follow.
I replaced this with the difference between _current_ and _start_ time, which should be less than the timeout.

Issue #218, opened 2020-11-10
Task report success despite one of the stages always failing


JafarAbdi:
I have a bug where the task report success despite one of the stages never generate a successful solution (in particular Connect), as you can see in the video below the task did succeed but the number of successful solutions in `move to pick` is zero, when I click on the task to visualize it somehow a solution get generated, also, executing the task start from the new solution of `move to pick`

It should be reproducible using this [branch](https://github.com/JafarAbdi/moveit_task_constructor/tree/task_bug) (I just added a constraint to make the stage always fail)

- Tested on Ubuntu 18.04 - Melodic - the master branch of MoveIt - both master branch of MTC and https://github.com/ros-planning/moveit_task_constructor/pull/216
- [Video link](https://youtu.be/yy-oe3NU-U8)

JafarAbdi:
I think I found the bug 

When plan inside [Connect::compute](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L133) [fails](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L173) it will call [makeSequential](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L180) which will insert a subsolution not marked as a failure with setting the start/end states (they will call addOutgoing/addIncoming) the returned SolutionSequence will be marked as failure [here](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L175) but this doesn't mark the subsolutions as failure

@rhaschke any idea how this could be fixed .? should we pass the success flag and mark the subsolutions as a failure, too .?

rhaschke:
Sorry for being unresponsive. I'm quite busy these days. I will try to look into this issue over the weekend. There is some other pending bug in Connect...

mamoll:
@rhaschke what is the other pending bug you're referring to? Perhaps we can help. We have a workaround for the bug Jafar reported, but it'd be nice to use a "clean" solution.

rhaschke:
The other bug is #182. But, I didn't have time yet to look into both of these issues.

PR #219, opened 2020-11-13
Don't register the first/last subsolutions to the start/end states of SolutionSequence

JafarAbdi:
Fix https://github.com/ros-planning/moveit_task_constructor/issues/218

### DIFF
--- a/core/include/moveit/task_constructor/storage.h
+++ b/core/include/moveit/task_constructor/storage.h
@@ -213,20 +213,22 @@ public:
 
 	/** set the solution's start_state_
 	 *
-	 * Must not be used with different states because it registers the solution with the state as well.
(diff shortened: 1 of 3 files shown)


JafarAbdi:
Another solution https://github.com/JafarAbdi/moveit_task_constructor/tree/pr-fix_connect_stage_1

rhaschke:
> Another solution https://github.com/JafarAbdi/moveit_task_constructor/tree/pr-fix_connect_stage_1

I think this is the correct solution to the problem: the sequence of sub solutions of a `Connect` stage, should become decoupled from its interface states. Please, give me some more time to think about this more deeply and write some real unit tests.

rhaschke:
Closing in favor of #220. Thanks, @JafarAbdi, for providing search directions!

PR #221, opened 2020-12-08
Rework Pruning

rhaschke:
With #220 I essentially disabled pruning to fix some urgent issues. It turned out that this feature wasn't fully thought about and it was broken literally everywhere. This is my work-in-progress to re-enable that important feature (or make it work in the first place).

I noticed several things:
- The action to be taken for pruning is highly dependent based on the stage that triggers it. 
  - Generator-like stages shouldn't trigger pruning at all: They just don't push new states to their push interfaces.
  - Propagator stages can _remove_ all interface states that are incoming to the failed source state.
However, this rule only applies to the default propagators MoveTo and MoveRelative that forward states in a 1:1 fashion.
If the propagator generates multiple output states for each input state, the source state can only be removed if all candidate extensions are exhausted. Thus, **pruning needs to be triggered from within the individual stages.**
  - Connect stages, on the other hand, must not remove these states, but just disable them (temporarily). If later new input states arrive, they can get reactivated.

So far, I considered pruning for `Connecting` stages only. Here, I stumbled over a symmetry issue: In the initial situation shown in the leftmost figure, we have several successful solutions (green) and some pending connect pairs (grey). CON2 fails to connect the red pair. This will disable (yellow) the solution trees originating from both ends as shown in the next figure.
If a new input arrives from the bottom to CON2, this should be connected to the disabled branch, thus re-enabling it.
On the other hand, if a new input arrives from the top for CON1, this must not be connected to the disabled branch!
The solution, I eventually found, is to remember on which end we originally failed and allow reconnecting at this side only.

image

### DIFF
--- a/core/include/moveit/task_constructor/storage.h
+++ b/core/include/moveit/task_constructor/storage.h
@@ -77,26 +77,33 @@ class InterfaceState
 	friend class SolutionBase;  // addIncoming() / addOutgoing() should be called only by SolutionBase
 	friend class Interface;  // allow Interface to set owner_ and priority_
 public:
+	enum Status
+	{
+		ENABLED,
+		DISABLED_START,  // state that was the root cause for disabling due to failure
+		DISABLED_END  // state on the other end of a disabled solution sequence
+	};
 	/** InterfaceStates are ordered according to two values:
 	 *  Depth of interlinked trajectory parts and accumulated trajectory costs along that path.
 	 *  Preference ordering considers high-depth first and within same depth, minimal cost paths.
 	 */
-	struct Priority : std::tuple<bool, unsigned int, double>
+	struct Priority : std::tuple<Status, unsigned int, double>
 	{
-		Priority(unsigned int depth, double cost, bool enabled = true)
-		  : std::tuple<bool, unsigned int, double>(enabled, depth, cost) {
+		Priority(unsigned int depth, double cost, Status status = ENABLED)
+		  : std::tuple<Status, unsigned int, double>(status, depth, cost) {
 			assert(std::isfinite(cost));
 		}
-		// Constructor copying depth and cost, but modifying enabled status
-		Priority(const Priority& other, bool enabled) : Priority(other.depth(), other.cost(), enabled) {}
+		// Constructor copying depth and cost, but modifying its status
+		Priority(const Priority& other, Status status) : Priority(other.depth(), other.cost(), status) {}
 
-		inline bool enabled() const { return std::get<0>(*this); }
+		inline Status status() const { return std::get<0>(*this); }
+		inline bool enabled() const { return std::get<0>(*this) == ENABLED; }
 		inline unsigned int depth() const { return std::get<1>(*this); }
 		inline double cost() const { return std::get<2>(*this); }
 
 		// add priorities
 		Priority operator+(const Priority& other) const {
-			return Priority(depth() + other.depth(), cost() + other.cost(), enabled() || other.enabled());
+			return Priority(depth() + other.depth(), cost() + other.cost(), std::min(status(), other.status()));
 		}
 		// comparison operators
 		bool operator<(const Priority& rhs) const;
--- a/core/include/moveit/task_constructor/storage.h
+++ b/core/include/moveit/task_constructor/storage.h
@@ -79,9 +79,9 @@ class InterfaceState
 public:
 	enum Status
 	{
-		ENABLED,
-		DISABLED_START,  // state that was the root cause for disabling due to failure
-		DISABLED_END  // state on the other end of a disabled solution sequence
+		ENABLED,  // state is actively considered during planning
+		DISABLED,  // state is disabled because a required connected state failed
+		DISABLED_FAILED,  // state that failed, causing the whole partial solution to be disabled
 	};
 	/** InterfaceStates are ordered according to two values:
 	 *  Depth of interlinked trajectory parts and accumulated trajectory costs along that path.
--- a/core/include/moveit/task_constructor/storage.h
+++ b/core/include/moveit/task_constructor/storage.h
@@ -390,6 +390,18 @@ template <>
 inline const InterfaceState::Solutions& trajectories<Interface::BACKWARD>(const InterfaceState* state) {
 	return state->incomingTrajectories();
 }
+
+/// Trait to retrieve opposite direction (FORWARD <-> BACKWARD)
+template <Interface::Direction dir>
+constexpr Interface::Direction opposite();
+template <>
+inline constexpr Interface::Direction opposite<Interface::FORWARD>() {
+	return Interface::BACKWARD;
+}
+template <>
+inline constexpr Interface::Direction opposite<Interface::BACKWARD>() {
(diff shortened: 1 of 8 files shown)


core/src/container.cpp
```diff
@@ -355,19 +395,51 @@ void SerialContainer::onNewSolution(const SolutionBase& current) {
...
+	// finally, store + announce new solutions to external interface
 	for (const auto& solution : sorted)
 		impl->liftSolution(solution, solution->internalStart(), solution->internalEnd());
 }
 
+void SerialContainer::onNewFailure(const Stage& child, const InterfaceState* from, const InterfaceState* to) {
+	switch (child.pimpl()->interfaceFlags()) {
+		case GENERATE:
+			// just ignore: the pair of (new) states isn't known to us anyway
+			// TODO: If child is a container, from and to might have associated solutions already!
```
v4hn, line 419: I'm not sure what you mean here. If the child is a generator container, there are still no neighboring states we could prune.
Do you want to prune potential alternative branches *inside* the container?
That might indeed make sense for structured serial containers. (e.g., pick-connect-place).

core/src/container.cpp
```diff
@@ -280,7 +281,32 @@ std::ostream& operator<<(std::ostream& os, const ContainerBase& container) {
...
+		if (!cimpl->starts() && !cimpl->ends())
+			os << "↕ " << std::endl;
+		if (cimpl->starts())
+			os << "↓ " << *child->pimpl()->starts() << std::endl;
+		if (cimpl->starts() && cimpl->ends())
+			os << std::setw(width) << "  ";
+		if (cimpl->ends())
+			os << "↑ " << *child->pimpl()->ends() << std::endl;
+	}
+}
```
v4hn, line 308: great help for debugging!

v4hn (commented):
Hi @rhaschke,

I just force-pushed a rebased branch and added a number of commits from my side.
The main changes/additions for today are

- rename `DISABLED_START / DISABLED_END` to `DISABLED_FAILED / DISABLED` to avoid overloaded terminology
- test and fix to support overlapping partial solution branches.
- disable pruning for propagators. As you stated yourself, this has to be implemented with support in the stages.

I want to address at least [the commented asserts](https://github.com/ros-planning/moveit_task_constructor/pull/221/files#diff-5b0930974b77eab73f48b74214a60ac0562cd2a110285279bb5ad1094b40a842R360-R361) before merge, would like to merge this feature soon.

core/src/container.cpp
rhaschke: Today, it's too late to think this through again...

core/include/moveit/task_constructor/storage.h
```diff
@@ -79,9 +79,9 @@ class InterfaceState
 public:
 	enum Status
 	{
-		ENABLED,
-		DISABLED_START,  // state that was the root cause for disabling due to failure
-		DISABLED_END  // state on the other end of a disabled solution sequence
+		ENABLED,  // state is actively considered during planning
+		DISABLED,  // state is disabled because a required connected state failed
+		DISABLED_FAILED,  // state that failed, causing the whole partial solution to be disabled
```
rhaschke, line 86: I like those new names.

core/src/container.cpp
```diff
@@ -412,11 +412,13 @@ void SerialContainer::onNewFailure(const Stage& child, const InterfaceState* fro
 			// TODO: If child is a container, from and to might have associated solutions already!
 			break;
 
+		// TODO(v4hn): Only disable partial incoming solution if there are no other solutions possible
+		// (Propagators might generate multiple solutions per incoming state)
```
rhaschke, line 423: I agree on this statement. But I would expect that a new feasible solution generated later will reactivate the whole solution branch again. Thus, we disable only temporarily. Doesn't that work?

core/test/test_serial.cpp
```diff
@@ -176,6 +183,28 @@ TEST(ConnectConnect, PruningForward) {
...
+	t.add(Stage::pointer(new Connect()));
+	t.add(Stage::pointer(new BackwardMockup()));
+	/* TODO(v4hn):
+	 * t.add(Stage::pointer(new GeneratorMockup()));
+	 * should be enough instead of the above, but propagators do not even respect ENABLED/DISABLED yet
+	 */
+
+	t.add(Stage::pointer(new GeneratorMockup()));
+	t.add(Stage::pointer(new ForwardMockup({ 0, 0 }, 2)));  // spawn two solutions for the only incoming state
+	t.add(Stage::pointer(new ForwardMockup({ 0, inf })));
```
rhaschke, line 201: This looks overly complex. Could you add a short comment describing the exact test case.

rhaschke (commented):
Thanks for your review and addons. I slightly clarified the comments for me and force-pushed the new history. Shall we have a zoom call to discuss open issues and next steps?

core/src/container.cpp
v4hn: Yes and no.

The other way around is the crucial failure: a new failure might disable a partial solution even if there is already valid alternatives for the branch. So if the last propagated solution for an incoming state is a failure all others will also stay disabled for no reason. In practice, a successful extension in the next stage might reactive it, but we can't guarantee the next stage can compute before we compute the disabled branch. That would be hell of a bug to search for.

core/src/container.cpp
v4hn: Nobody except yourself expects you to respond right away :smile:

core/src/container.cpp
rhaschke: I agree, but didn't you fix this in 2b759570a6edc48462b0930418c7999e6793fefd?

rhaschke:
> I just force-pushed a rebased branch and added a number of commits from my side.

I didn't notice the first time. But how did you manage to **force-push** to our organization's branch `ubi-agni:pruning`?
As far as I remember, pushing to organization branches is forbidden and force-pushing to non-organization repos is also not allowed.

core/test/test_serial.cpp
v4hn: You are one to talk, the other pruning tests feel just as complex to me. :D

I will try to address the TODO here soon and thus simplify the test. Also I just noticed the comment is wrong and should read `t.add(Stage::pointer(new BackwardMockup()));`

rhaschke: > You are one to talk, the other pruning tests feel just as complex to me.

Touche.

core/src/container.cpp
v4hn: You are right. There was one other issue, but I added a separate test (and fix) for that now.

v4hn:
> I didn't notice the first time. But how did you manage to force-push to our organization's branch ubi-agni:pruning?
As far as I remember, pushing to organization branches is forbidden and force-pushing to non-organization repos is also not allowed.

I have no idea, but apparently I'm still allowed to do so. :sunglasses:

v4hn:
Assuming CI passes, I consider this ready to merge by merge-commit.

rhaschke:
Thanks a lot for your contributions, Michael. Now we just need to tackle pruning for containers as well.

Issue #222, opened 2020-12-15
Demo broken


gautz:
When running ``rosrun moveit_task_constructor_demo scanning``
My computer name has some ``-``sign, I get following error:
```
terminate called after throwing an instance of 'ros::InvalidNameException'
  what():  Character [-] at element [16] is not valid in Graph Resource Name [get_solution_ComputeNameWith-].  Valid characters are a-z, A-Z, 0-9, / and _.
Aborted (core dumped)
```
I can traceback the regression to some time after commit be270cb57466d456ede2131d6cfaa8257b8428d2

JafarAbdi:
@gautz This's should be fixed in https://github.com/ros-planning/moveit_task_constructor/pull/223

gautz:
thanks for the quick fix, execution works! I am still getting an error that may be unrelated.
``Unregistered type: St3mapINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEdSt4lessIS5_ESaISt4pairIKS5_dEEE``

JafarAbdi:
> thanks for the quick fix, execution works! I am still getting an error that may be unrelated.
> `Unregistered type: St3mapINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEdSt4lessIS5_ESaISt4pairIKS5_dEEE`

This corresponds to the following type `std::map<std::string, double>` are you trying to set a property's value to `std::map` .?

gautz:
yes, in cartesian.cpp on line 106:
``std::map<std::string, double> offsets = { { "panda_joint1", M_PI / 6. }, { "panda_joint3", -M_PI / 6 } };``

rhaschke:
You can safely ignore this warning. It just means that you cannot inspect the value of this property in rviz.

PR #223, opened 2020-12-15
introspection: Fix hostname having  as character

JafarAbdi:
Fix: https://github.com/ros-planning/moveit_task_constructor/issues/222

### DIFF
--- a/core/src/introspection.cpp
+++ b/core/src/introspection.cpp
@@ -57,6 +57,8 @@ std::string getTaskId(const TaskPrivate* task) {
 	std::ostringstream oss;
 	char our_hostname[256] = { 0 };
 	gethostname(our_hostname, sizeof(our_hostname) - 1);
+	// Hostname could have `-` as a character but this is an invalid character in ROS so we replace it with `_`
+	std::replace(std::begin(our_hostname), std::end(our_hostname), '-', '_');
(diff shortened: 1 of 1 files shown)


JafarAbdi:
This's the Linux hostname specification https://man7.org/linux/man-pages/man7/hostname.7.html

felixvd (approved):
LGTM, although I haven't tested. Can't imagine many failure modes though.

mamoll (approved):
LGTM. This seems like a straightforward fix.

core/src/introspection.cpp
```diff
@@ -57,6 +57,8 @@ std::string getTaskId(const TaskPrivate* task) {
 	std::ostringstream oss;
 	char our_hostname[256] = { 0 };
 	gethostname(our_hostname, sizeof(our_hostname) - 1);
+	// Hostname could have `-` as a character but this is an invalid character in ROS so we replace it with `_`
+	std::replace(std::begin(our_hostname), std::end(our_hostname), '-', '_');
```
JafarAbdi, line 61: Or maybe replacing these lines with

```cpp
static const auto our_hostname = [] {
   char our_hostname[256] = { 0 };
   gethostname(our_hostname, sizeof(our_hostname) - 1);
   std::replace(std::begin(our_hostname), std::end(our_hostname), '-', '_');
   return std::string(our_hostname);
}();
```

rhaschke (approved):

PR #228, opened 2021-01-22
demo: Update mtc.rviz to use fixed robot color + alpha

EricCousineau-TRI:
Just a minor tweak:

Before:
> image

After:
> image

### DIFF
--- a/demo/config/mtc.rviz
+++ b/demo/config/mtc.rviz
@@ -104,9 +104,10 @@ Visualization Manager:
           Expand Tree: false
           Link Tree Style: Links in Alphabetic Order
         Robot Alpha: 1
+        Robot Alpha: 0.5
         Show Robot Collision: false
         Show Robot Visual: true
-        Use Fixed Robot Color: false
+        Use Fixed Robot Color: true
         Value: ""
       Robot Description: robot_description
       Scene:


rhaschke (approved):
It's a matter of taste, but yes, this better highlights the difference between the MTC robot state and the real one.

rhaschke:
Maybe we should introduce this setting as a default for the MTC display?

EricCousineau-TRI:
Yeah, that'd be awesome! Where does that come in?

rhaschke:
Here: https://github.com/ros-planning/moveit_task_constructor/blob/f99d03386ad8751fb0ade71c09a9810019a950ad/visualization/visualization_tools/src/task_solution_visualization.cpp#L115
and here: https://github.com/ros-planning/moveit_task_constructor/blob/f99d03386ad8751fb0ade71c09a9810019a950ad/visualization/visualization_tools/src/task_solution_visualization.cpp#L124

rhaschke (commented):
The reason, I don't fully like this PR yet, is that the robot will remain purple-only and transparent when moving. This makes it hard to recognize details. Actually, it would be nice to just tint the original meshes.
I will leave this open to be resolved later.

rhaschke:
Closing due to inactivity

PR #231, opened 2021-01-29
demo: switch to moveit_resources_panda_moveit_config

rhaschke:
In Noetic, the tutorial demos don't work anymore, because Franka has changed the collision model of the robot.
As a quick hack for this, this PR switches to the old Panda config, available in `moveit_resources`.


JafarAbdi (approved):

v4hn:
That's what moveit_resources is for, so YES. Thank you :)
Did you migrate any physical system to noetic?

PR #237, opened 2021-02-22
remove roscpp as an export dependency for the core package

v4hn:
We do use ROS in the background. But there is no need for a public export dependency on it.

This patch also resolves the following catkin_lint issue:

    moveit_task_constructor_core: CMakeLists.txt(17): error: package 'roscpp' must be in CATKIN_DEPENDS in catkin_package()

I also see this other warning which we might want to address, but I'm not convinced it's worth the additional include directives in dependent targets.

    moveit_task_constructor_core: src/CMakeLists.txt(37): warning: target_include_directories() exports external PUBLIC path '${catkin_INCLUDE_DIRS}'

### DIFF
--- a/core/include/moveit/task_constructor/stage_p.h
+++ b/core/include/moveit/task_constructor/stage_p.h
@@ -43,8 +43,6 @@
 #include <moveit/task_constructor/cost_terms.h>
 #include <moveit/task_constructor/cost_queue.h>
 
-#include <ros/ros.h>
-
 #include <ostream>
 #include <chrono>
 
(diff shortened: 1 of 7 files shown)


core/include/moveit/task_constructor/stage_p.h
```diff
@@ -116,7 +114,7 @@ class StagePrivate
 	/// enforce only one parent exists
 	inline bool setParent(ContainerBase* parent) {
 		if (parent_) {
-			ROS_ERROR_STREAM("Tried to add stage '" << name() << "' to two parents");
```
rhaschke, line 119: For consistency, you should also remove the return value. Throwing _and_ returning false doesn't make sense.

rhaschke (changes requested):
I'm fine with this change. Please, also remove the return value.

v4hn:
true. done.

rhaschke (approved):
I thought of a more fundamental change: removing the return value from the function. As you decided to use an exception instead of a return value, that's the logical conclusion for me. I pushed a corresponding change myself.

v4hn:
> removing the return value from the function.

Makes sense. Most examples use `ContainerBase::add` anyway, and we don't change this one.
Could you please fix the unit tests that broke because of the API change?

rhaschke:
> Could you please fix the unit tests that broke because of the API change?

Sure. Didn't compile the tests before.

PR #170, opened 2020-06-03
Port MTC to ROS2

JafarAbdi:
This shouldn't be against master branch

Initial effort for porting MTC to ROS2, next step is to port rviz_marker_tools after that continue porting mtc core and its tests

PR #240, opened 2021-03-08
ROS2: re-enable the singleton of the task panel

JafarAbdi:
Depends on https://github.com/ros-planning/moveit_task_constructor/pull/170 & https://github.com/ros2/rviz/pull/660


JafarAbdi:
No longer needed since it's part of the main PR

PR #246, opened 2021-03-22
explicitly instantiate template for required values

v4hn:
Fixup for 10f6b7a58ed6063771eb7b9cbe9931fc71497ca8 .

On my release workspace I see
/usr/bin/ld: libmoveit_task_constructor_core_stages.so: undefined reference to `
void moveit::task_constructor::PropagatingEitherWay::send<(moveit::task_constructor::Interface::Direction)0>(moveit::task_constructor::InterfaceState const&, moveit::task_constructor::InterfaceState&&, moveit::task_constructor::SubTrajectory&&)'
/usr/bin/ld: libmoveit_task_constructor_core_stages.so: undefined reference to `
void moveit::task_constructor::PropagatingEitherWay::send<(moveit::task_constructor::Interface::Direction)1>(moveit::task_constructor::InterfaceState const&, moveit::task_constructor::InterfaceState&&, moveit::task_constructor::SubTrajectory&&)'

and indeed the methods are not explicitly instantiated and not used in the compile unit.

### DIFF
--- a/core/src/stage.cpp
+++ b/core/src/stage.cpp
@@ -578,6 +578,11 @@ template <Interface::Direction dir>
 void PropagatingEitherWay::send(const InterfaceState& start, InterfaceState&& end, SubTrajectory&& trajectory) {
 	pimpl()->send<dir>(start, std::move(end), std::make_shared<SubTrajectory>(std::move(trajectory)));
 }
+// Explicit template instantiation is required. The compiler, otherwise, might just inline them.
+template void PropagatingEitherWay::send<Interface::FORWARD>(const InterfaceState& start, InterfaceState&& end,
+                                                             SubTrajectory&& trajectory);
+template void PropagatingEitherWay::send<Interface::BACKWARD>(const InterfaceState& start, InterfaceState&& end,
+                                                              SubTrajectory&& trajectory);
 
 template <Interface::Direction dir>
 void PropagatingEitherWay::computeGeneric(const InterfaceState& start) {


rhaschke (commented):
This is weird. Actually, these functions are used (and thus instantiated) in `computeGeneric()`:
https://github.com/ros-planning/moveit_task_constructor/blob/b638c503903ab2bf91753c6e8fcc6cf2240d31a4/core/src/stage.cpp#L586-L595
which in turn is instantiated here:
https://github.com/ros-planning/moveit_task_constructor/blob/10f6b7a58ed6063771eb7b9cbe9931fc71497ca8/core/include/moveit/task_constructor/stage.h#L294-L295

I didn't have issues running this code from `devel` space. Do you observe this in `install` space only?
In any case, we should better instantiate `computeGeneric()`, shouldn't we?

v4hn:
It's no install workspace, just a `-DCMAKE_BUILD_TYPE=Release` build using gcc 10.

Building the same sources with `-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang`
(my usual development workspace), the `send` symbols are around so I did not notice it before.

You are right, the symbol is used and `computeGeneric` is also present in the so.
Apparently the compiler decides to inline the calls in the unit and drops the public symbol, though I fail to see why.
I thought this might relate to https://github.com/ros-planning/moveit_task_constructor/pull/147 ,
but I can reproduce it without the flag as well.

I filed this as a pull-request to let you have a look, but I can't find a different patch at the moment to fix the missing symbols.

rhaschke:
Thanks for the clarification. I suggest to prepend the instantiation with a comment like this:
`// Explicit template instantiation is required. The compiler, otherwise, might just inline them.`

I guess you have other stages (than MoveTo/MoveRel) using sendForward/sendBackward?

v4hn:
> Thanks for the clarification. I suggest to prepend the instantiation with a comment like this:
> `// Explicit template instantiation is required. The compiler, otherwise, might just inline them.`

added

> I guess you have other stages (than MoveTo/MoveRel) using sendForward/sendBackward?

I do, but the error already manifests with MoveTo/MoveRel because they are compiled to a separate library `libmoveit_task_constructor_core_stages`

rhaschke (approved):
Thanks

PR #249, opened 2021-03-26
rework PipelinePlanner creation

v4hn:
https://github.com/ros-planning/moveit/pull/2127 merged support
for multiple planner plugins. As a result, the planner specification
should include a pipeline name from now on.

Support the new schema and provide a fallback for older configurations.

While doing so, I moved the create methods to the class they actually belong to
and introduced a clean options parameter

This change is somewhat urgent because the pickplace demo is broken since moveit/2127 was merged.

### DIFF
--- a/core/include/moveit/task_constructor/solvers/pipeline_planner.h
+++ b/core/include/moveit/task_constructor/solvers/pipeline_planner.h
@@ -55,7 +55,21 @@ MOVEIT_CLASS_FORWARD(PipelinePlanner)
 class PipelinePlanner : public PlannerInterface
 {
 public:
-	PipelinePlanner();
+	struct Specification
+	{
+		moveit::core::RobotModelConstPtr model;
+		std::string ns{ "move_group" };
+		std::string pipeline{ "ompl" };
+		std::string adapter_param{ "request_adapters" };
+	};
+
+	static planning_pipeline::PlanningPipelinePtr create(const moveit::core::RobotModelConstPtr& model) {
+		return create(Specification{ .model = model });
+	}
+
+	static planning_pipeline::PlanningPipelinePtr create(const Specification& spec);
+
+	PipelinePlanner(const std::string& pipeline = "ompl");
 
 	PipelinePlanner(const planning_pipeline::PlanningPipelinePtr& planning_pipeline);
 
@@ -73,6 +87,7 @@ public:
 	          const moveit_msgs::Constraints& path_constraints = moveit_msgs::Constraints()) override;
 
 protected:
+	std::string pipeline_name_;
 	planning_pipeline::PlanningPipelinePtr planner_;
 };
 }  // namespace solvers
--- a/core/include/moveit/task_constructor/solvers/pipeline_planner.h
+++ b/core/include/moveit/task_constructor/solvers/pipeline_planner.h
@@ -64,7 +64,9 @@ public:
 	};
 
 	static planning_pipeline::PlanningPipelinePtr create(const moveit::core::RobotModelConstPtr& model) {
-		return create(Specification{ .model = model });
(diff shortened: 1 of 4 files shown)


core/src/solvers/pipeline_planner.cpp
```diff
@@ -48,7 +48,62 @@ namespace moveit {
...
+			model_it = cache_.insert(cache_.begin(), std::make_pair(model, PlannerMap()));
+
+		return model_it->second.insert(std::make_pair(id, PlannerMap::mapped_type())).first->second;
+	}
+};
+
+planning_pipeline::PlanningPipelinePtr PipelinePlanner::create(const PipelinePlanner::Specification& spec) {
+	static PlannerCache cache;
+
+	constexpr char PLUGIN_PARAMETER_NAME[]{ "planning_plugin" };
```
v4hn, line 80: I dropped support for custom plugin parameter names on purpose because the new parameter scheme is a much cleaner replacement for non-standard parameter names.

I'm ok with leaving the parameter configurable as well or hard-code the `adapter_param`.
I believe there might be sufficient use for a configurable adapter parameter though.

v4hn:
```
Errors     << moveit_task_constructor_core
In member function ‘virtual void moveit::task_constructor::solvers::PipelinePlanner::init(const RobotModelConstPtr&)’:
sorry, unimplemented: non-trivial designated initializers not supported
   planner_ = create(Specification{ .model = robot_model, .pipeline = pipeline_name_ });
```

Wow, feels like we are back in the old days...
Turns out this is actually a C++20 feature that's active by default in both current g++ and clang++ :open_mouth:

I pushed a patch without the language feature...

core/include/moveit/task_constructor/solvers/pipeline_planner.h
```diff
@@ -55,7 +55,21 @@ MOVEIT_CLASS_FORWARD(PipelinePlanner)
...
+	struct Specification
+	{
+		moveit::core::RobotModelConstPtr model;
+		std::string ns{ "move_group" };
+		std::string pipeline{ "ompl" };
+		std::string adapter_param{ "request_adapters" };
+	};
+
+	static planning_pipeline::PlanningPipelinePtr create(const moveit::core::RobotModelConstPtr& model) {
+		return create(Specification{ .model = model });
```
rhaschke, line 67: Isn't this line using the C++20 feature as well? (https://github.com/ros-planning/moveit_task_constructor/pull/249#issuecomment-808472365)

rhaschke (approved):
I pushed two more commits:
- I think we can directly remove Task::createPlanner (and not only deprecate it)
- I'm not a fan of brace initializers.

core/include/moveit/task_constructor/solvers/pipeline_planner.h
v4hn: Yes and no. It's the same feature, but apparently the version of gcc in CI *does* implicitly support it as long as all initialized members are *in order*, even if some in the end are not specified. The other use skipped some members and triggered the "sorry unimplemented" error.

I rewrote it here as well.

v4hn:
> I'm not a fan of the brace initializers. They can be confused easily with initializer lists.

I do prefer and use uniform initialization a lot these days and I do think they solve actual problems in the language. 
With your line of argumentation `A a = <foo>;` can be (and often is) misunderstood as invoking `operator=()`, when it's really invoking `A(<foo>)`. If code uses `initializer_list`, you have to be very aware of the constructor anyway because they likely take priority over all other constructors anyway.

I'll not touch it again because you went through the trouble of changing it, but I do prefer to avoid printf-style and non-uniform initializers.

That aside, this should be good to merge now.

Issue #247, opened 2021-03-23
Pruning across container boundaries


v4hn:
In #221 we implemented correct pruning of infeasible candidate solutions inside SerialContainers.

However, this does not work across container boundaries yet because Containers currently act as opaque stages in their parent.

Consider this example structure:

![rect2643](https://user-images.githubusercontent.com/680358/112162753-a44d3c80-8bec-11eb-9a04-50be71d6ae7b.png)

Assuming InterfaceState `a` fails to extend in Stage `Back`, `Fwd2` does not have to compute  InterfaceState `b` and the state can be disabled.
The same is true for the inverted case.

Parallel containers can result in similar situations but have to consider alternative paths through the different children as well.

I believe this is what you meant in https://github.com/ros-planning/moveit_task_constructor/pull/221#issuecomment-804847964 @rhaschke ?

rhaschke:
Yes, that's definitely an open issue. I think, the other way around, i.e. a failing InterfaceState inside der container isn't yet propagated to the outside either.

rhaschke:
I think we need to rethink the propagation of the `Interface::Status` for pruning once again: Currently, propagation occurs along connected solutions, neglecting the type of stage owning the `Interface`. Thus, currently, we cannot propagate through `Connect` stages, if their solutions are not yet connected. However, here pruning would be particularly beneficial, because we can rule out many opposite InterfaceStates. I think, status propagation should be handled by the (private?) stage classes.

Also, in order to allow propagation from inside containers to the outside, containers need to hook into their children's Interface to get notified on Status changes. This is currently not possible either.

PR #250, opened 2021-03-27
prune infeasible partial solutions inside container from the outside

v4hn:
Partial fix for #247 , concretely the case where (a) fails and prunes (b) in the example there.
Note that I had to rework the container slightly in the test because the trivial scheduling would always schedule (b) before (a).

I initially fixed the inverted lookup by iterating over the map, but also set up a bimap as a cleaner alternative in a second commit.

I'm uncertain which implementation we want to prefer here, but all in all it's probably premature-optimization to think about performance and I prefer the bimap.


core/src/container.cpp
```diff
@@ -521,8 +529,9 @@ void SerialContainerPrivate::resolveInterface(InterfaceFlags expected) {
 		validateInterface<END_IF_MASK>(*last.pimpl(), expected);
 		// connect last child's (end) pull interface
 		if (const InterfacePtr& target = last.pimpl()->ends())
-			ends_.reset(new Interface(
-			    [this, target](Interface::iterator it, bool updated) { this->copyState(it, target, updated); }));
+			ends_.reset(new Interface([this, target](Interface::iterator it, bool updated) {
+				this->copyState<Interface::FORWARD>(it, target, updated);
```
rhaschke: Copy-Paste-error? Shouldn't this read `Interface::BACKWARD`?

core/include/moveit/task_constructor/container_p.h
```diff
@@ -151,7 +158,9 @@ class ContainerBasePrivate : public StagePrivate
 	container_type children_;
 
 	// map start/end states of children (internal) to corresponding states in our external interfaces
-	std::map<const InterfaceState*, InterfaceState*> internal_to_external_;
+	boost::bimap<boost::bimaps::unordered_set_of<const InterfaceState*>,
```
rhaschke, line 161: Why using a set of internal states here? Each internal state still has a unique external counterpart.
However, we have multiple internal stages mapping to the same external one in case of parallel containers, where an external stages is copied to all children.

rhaschke (commented):
I'm not yet convinced of the bimap. Otherwise I approve. Note, that I pushed two intermediate commits. I was able to force-push to your branch as well. I consider this a security issue of github...
What is missing for inside -> outside pruning?

core/src/container.cpp
v4hn: Thanks, obviously a mistake.

v4hn (commented):
> I'm not yet convinced of the bimap.

I'm not sure how to argue for it any further, unless you only mean the inline question.
The alternative is to iterate over the map items for lookup or modify the Interface to store a list of "children copies" with each external state.

> I was able to force-push to your branch as well. I consider this a security issue of github...

In my case that's ok. It's my pull-request and the branch belongs to my personal account (not a github group).

>  What is missing for inside -> outside pruning?

Probably not a lot, I'm on it (if my family lets me).

core/include/moveit/task_constructor/container_p.h
rhaschke: Sorry, I should have had a deeper look into the bimap doc before. So far, I only knew about `boost::bimap<A, B>`, but didn't have seen the collection-type specifications.

core/src/container.cpp
```diff
@@ -109,20 +109,18 @@ void ContainerBasePrivate::compute() {
 template <Interface::Direction dir>
 void ContainerBasePrivate::copyState(Interface::iterator external, const InterfacePtr& target, bool updated) {
 	if (updated) {
-		// TODO(v4hn): This is inefficient, consider storing inverse mapping for each start/end pair in children
-		// for parallel containers there will be multiple internal states
-		std::for_each(internal_to_external_.begin(), internal_to_external_.end(), [&external](auto& p) {
-			if (p.second == &*external)
-				setStatus<dir>(p.first, external->priority().status());
-		});
+		auto internals{ externalToInternalMap().equal_range(&*external) };
```
rhaschke, line 112: Doesn't `equal_range` require _sorted_ keys?

v4hn: No, [unordered_multimap::equal_range](https://en.cppreference.com/w/cpp/container/unordered_multimap/equal_range) is defined. [It does not keep the buckets sorted but keeps equal elements together](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/hashtable.h#L1806).

v4hn:
>  What is missing for inside -> outside pruning?

Turns out it's some more unrelated structural changes, so I'll create another pull-request for it when I'm done.

This PR is ready to fast-forward (after CI is done).

Issue #78, opened 2019-02-13
tf2 transition


rhaschke:

v4hn:
This is mostly done ([except for backward compatibility](https://github.com/ros-planning/moveit_task_constructor/blob/master/visualization/visualization_tools/src/marker_visualization.cpp#L75)), but to close this issue `grep -r tf::` should return only this occurrance.
In order to get there, all use of `eigen_conversions` should be reworked to use `tf2::convert` and the `tf2_eigen` package.

PR #255, opened 2021-04-15
improve visualization for MoveRelative stage

j-kuehn:
Implementation for this [TODO](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/move_relative.cpp#L273)

The `MoveRelative` stage can now be visualized as a red-green arrow, where the red part shows the invalid part and the green part shows the valid part of the plan. The construction of the arrow is different for forward and backward propagators.

For forward propagators, it looks like this:

https://user-images.githubusercontent.com/57101356/114857654-cfd8d680-9de8-11eb-8a97-ac9b20e298d0.mp4

and for backward propagators like this:

https://user-images.githubusercontent.com/57101356/114857852-0d3d6400-9de9-11eb-9021-32ebb64ec019.mp4

Unfortunately, for forward propagators with only short invalid parts, the arrow tip looks weird: 

https://user-images.githubusercontent.com/57101356/114857959-2ba35f80-9de9-11eb-8e70-b97b590edb32.mp4

I couldn't think of a reasonably easy way around this (except for removing the tip entirely), so I left it like this.

I changed the `rviz_marker_tools::makeArrow` signature to explicitly include the diameter, so that the diameter is not a function of the length anymore. I also removed the default value, as none of the other `make*` functions have one and I found no reference of the default being used.

Also, I was curious why the plan should only be visualized when the planned trajectory is longer than `1e-3`. I left it in there for now, but think this restriction could also be removed.

### DIFF
--- a/core/src/stages/move_relative.cpp
+++ b/core/src/stages/move_relative.cpp
@@ -104,6 +104,63 @@ static bool getJointStateFromOffset(const boost::any& direction, const moveit::c
 	return false;
 }
 
+static void visualizePlan(std::deque<visualization_msgs::Marker>& markers, Interface::Direction dir, bool success,
+                          const std::string& ns, const std::string& frame_id, const Eigen::Isometry3d& link_pose,
+                          const Eigen::Isometry3d& reached_pose, const Eigen::Vector3d& linear, double distance) {
+	double linear_norm = linear.norm();
+
+	// rotation of the target direction and for the cylinder marker
+	auto quat_target = Eigen::Quaterniond::FromTwoVectors(Eigen::Vector3d::UnitX(), linear);
+	auto quat_cylinder = quat_target * Eigen::AngleAxisd(0.5 * M_PI, Eigen::Vector3d::UnitY());
+
+	// link position before planning; reached link position after planning; target link position
+	Eigen::Vector3d pos_link = link_pose.translation();
+	Eigen::Vector3d pos_reached = reached_pose.translation();
+	Eigen::Vector3d pos_target = pos_reached + quat_target * Eigen::Vector3d(linear_norm - distance, 0, 0);
+
+	visualization_msgs::Marker m;
+	m.ns = ns;
+	m.header.frame_id = frame_id;
+	if (dir == Interface::FORWARD) {
+		if (success) {
+			// valid part: green arrow
+			rviz_marker_tools::makeArrow(m, pos_link, pos_reached, 0.1 * linear_norm);
+			rviz_marker_tools::setColor(m.color, rviz_marker_tools::LIME_GREEN);
+			markers.push_back(m);
+		} else {
+			// invalid part: red arrow
+			// set head length to keep default shaft:head proportion of 1:0.3 as defined in
+			// rviz/default_plugin/markers/arrow_marker.cpp#L105
+			rviz_marker_tools::makeArrow(m, pos_reached, pos_target, 0.1 * linear_norm, 0.23 * linear_norm);
+			rviz_marker_tools::setColor(m.color, rviz_marker_tools::RED);
+			markers.push_back(m);
+
+			// valid part: green cylinder
+			rviz_marker_tools::makeCylinder(m, 0.1 * linear_norm, distance);
+			rviz_marker_tools::setColor(m.color, rviz_marker_tools::LIME_GREEN);
+			// position half-way between pos_link and pos_reached
+			tf::pointEigenToMsg(0.5 * (pos_link + pos_reached), m.pose.position);
+			tf::quaternionEigenToMsg(quat_cylinder, m.pose.orientation);
+			markers.push_back(m);
+		}
+	} else {
+		// valid part: green arrow
+		// head length according to above comment
+		rviz_marker_tools::makeArrow(m, pos_reached, pos_link, 0.1 * linear_norm, 0.23 * linear_norm);
+		rviz_marker_tools::setColor(m.color, rviz_marker_tools::LIME_GREEN);
+		markers.push_back(m);
+		if (!success) {
+			// invalid part: red cylinder
+			rviz_marker_tools::makeCylinder(m, 0.1 * linear_norm, linear_norm - distance);
+			rviz_marker_tools::setColor(m.color, rviz_marker_tools::RED);
+			// position half-way between pos_reached and pos_target
+			tf::pointEigenToMsg(0.5 * (pos_reached + pos_target), m.pose.position);
+			tf::quaternionEigenToMsg(quat_cylinder, m.pose.orientation);
+			markers.push_back(m);
+		}
+	}
+}
+
 bool MoveRelative::compute(const InterfaceState& state, planning_scene::PlanningScenePtr& scene,
                            SubTrajectory& solution, Interface::Direction dir) {
 	scene = state.scene()->diff();
@@ -240,19 +297,21 @@ bool MoveRelative::compute(const InterfaceState& state, planning_scene::Planning
 
 		success = planner_->plan(state.scene(), *link, target_eigen, jmg, timeout, robot_trajectory, path_constraints);
 
+		robot_state::RobotStatePtr& reached_state = robot_trajectory->getLastWayPointPtr();
+		reached_state->updateLinkTransforms();
+		const Eigen::Isometry3d& reached_pose = reached_state->getGlobalLinkTransform(link);
+
+		double distance = 0.0;
+		if (robot_trajectory && robot_trajectory->getWayPointCount() > 0) {
+			if (use_rotation_distance) {
+				Eigen::AngleAxisd rotation(reached_pose.linear() * link_pose.linear().transpose());
+				distance = rotation.angle();
+			} else
+				distance = (reached_pose.translation() - link_pose.translation()).norm();
+		}
+
 		// min_distance reached?
 		if (min_distance > 0.0) {
-			double distance = 0.0;
-			if (robot_trajectory && robot_trajectory->getWayPointCount() > 0) {
-				robot_state::RobotStatePtr& reached_state = robot_trajectory->getLastWayPointPtr();
-				reached_state->updateLinkTransforms();
-				const Eigen::Isometry3d& reached_pose = reached_state->getGlobalLinkTransform(link);
-				if (use_rotation_distance) {
-					Eigen::AngleAxisd rotation(reached_pose.linear() * link_pose.linear().transpose());
-					distance = rotation.angle();
-				} else
-					distance = (reached_pose.translation() - link_pose.translation()).norm();
-			}
 			success = distance >= min_distance;
 			if (!success) {
 				char msg[100];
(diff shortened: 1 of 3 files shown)


rviz_marker_tools/include/rviz_marker_tools/marker_creation.h
```diff
@@ -66,7 +66,8 @@ inline visualization_msgs::Marker& makeMesh(visualization_msgs::Marker& m, const
 }
 
 /// create an arrow along x-axis
-visualization_msgs::Marker& makeArrow(visualization_msgs::Marker& m, double scale = 1.0, bool tip_at_origin = false);
+visualization_msgs::Marker& makeArrow(visualization_msgs::Marker& m, double diameter, double length,
+                                      bool tip_at_origin = false);
```
rhaschke, line 70: Please keep the old API and forward to the new function as `makeArrow(m, 0.1*scale, scale, tip_at_origin)`

rhaschke (commented):
The videos look good. Did you also consider the case of a pure rotation? In this case, the _linear_ arrow would be zero-length, although the `distance` would be non-zero. This was protected via the `1e-3` threshold before. Now, you seem to mix these two.
To improve on too short arrowheads: What about using an arrow tip without any shaft (of the same color). The shaft would be the cylinder of the other color.

j-kuehn:
I had not considered the case of pure rotation, thanks for the hint. But since I left the protection in there, it still works as expected: no visualization for a pure rotation, the `distance` is only used if `linear_norm > 1e-3`. Wouldn't it make more sense to use the `use_rotation_distance` variable for the decision to visualize or not, though? It also decides whether the `distance` is calculated as linear or angular motion, so it seems more consistent to me than the somewhat arbitrary `1e-3`.

Only using an arrowhead without a shaft for short invalid parts would require defining the arrow with start/end points instead of position/orientation (see [here](http://wiki.ros.org/rviz/DisplayTypes/Marker#Arrow_.28ARROW.3D0.29)). With the available interface for creating markers, there would always be a 'cut' in the visualization at the point where the red invalid part switches from only being an arrowhead to having the normal arrow proportions.

TL;DR Imho the tradeoff between the additional code complexity a case distinction would bring and the improved visualization for short invalid parts is not worth it and I'd stick with the short arrowheads for the sake of simplicity. If you think the benefit justifies an increased code complexity, I'd be happy to try and implement it, though.

rhaschke:
Yes, the distance definitely should consider the linear norm only. Did you try a real twist composed of rotation + translation along some common axis?
Regarding the arrowheads: I agree that extra complexity should be avoided. However, what about a slightly thinner arrow surrounded by cylinder occluded the longer arrow shaft?

core/src/stages/move_relative.cpp
```diff
@@ -104,6 +104,59 @@ static bool getJointStateFromOffset(const boost::any& direction, const moveit::c
 	return false;
 }
 
+static void shapeMarker(visualization_msgs::Marker& m, uint32_t shape, rviz_marker_tools::Color color_id,
+                        const Eigen::Vector3d& pos, const Eigen::Quaterniond& quat, double width, double length) {
+	rviz_marker_tools::setColor(m.color, color_id);
```
v4hn, line 109: Why do you set color before "making" the marker? I had to check to see the color's not actually reset below.

v4hn (commented):
> Wouldn't it make more sense to use linear_norm > 0 for the decision to visualize or not, though?

I agree. Especially because [the same check is already used in the class](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/move_relative.cpp#L177) to differentiate between pure rotational "mixed" modes.

> Imho the tradeoff between the additional code complexity a case distinction would bring and the improved visualization for short invalid parts is not worth it and I'd stick with the short arrowheads for the sake of simplicity

I'm unsure about this one. It would be nice to get a "normal" full arrow-head [with factor 0.23 of the whole arrow](https://github.com/ros-visualization/rviz/blob/noetic-devel/src/rviz/default_plugin/markers/arrow_marker.cpp#L105) of one color, even if the actual unreachable part of the trajectory is very short.
But then again, it's more accurate to show the translation that really failed and not exaggerate.

j-kuehn:
The arrow creation is now implemented with passing start and end points. I first reverted to the old API, then overloaded `makeArrow` to allow this alternative of defining the arrow.

The benefit of this is that the `scale.z` now allows to set the length of the arrowhead. Accordingly, I set this to `0.23 * linear_norm` to keep the default proportions of arrows:

backward-invalid
forward-invalid

If the invalid part (or the valid part for backward direction) is too short to have the default proportions, [this](https://github.com/ros-visualization/rviz/blob/noetic-devel/src/rviz/default_plugin/markers/arrow_marker.cpp#L110) ensures that the arrow only consists of an arrowhead without a shaft:
forward-invalid-short

For long valid parts and really short invalid parts it looks like this now:

https://user-images.githubusercontent.com/57101356/116075855-c3c80100-a693-11eb-988c-8101a5e6504a.mp4

Backward direction accordingly:
backward-invalid-short

Imho this visualization is good enough.

> Did you try a real twist composed of rotation + translation along some common axis?

I did test multiple twist messages and this works as expected. If the target direction only consists of a rotation, no visualization is computed. If the target direction consists of both a rotation and translation, the visualization is computed along the translational part.

core/src/stages/move_relative.cpp
```diff
@@ -104,6 +104,59 @@ static bool getJointStateFromOffset(const boost::any& direction, const moveit::c
...
+	} else if (shape == visualization_msgs::Marker::CYLINDER) {
+		rviz_marker_tools::makeCylinder(m, width, length);
+	}
+	tf::pointEigenToMsg(pos, m.pose.position);
+	tf::quaternionEigenToMsg(quat, m.pose.orientation);
+}
+
+static void visualizePlan(Interface::Direction dir, bool success, SubTrajectory& solution, const std::string& ns,
+                          const std::string& frame_id, Eigen::Quaterniond& quat, Eigen::Vector3d& pos,
+                          double linear_norm, double distance) {
```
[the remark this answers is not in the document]
j-kuehn: I am now passing the `reached_pose` and calculate the required rotation and translations inside the method. I am also passing `link_pose` to get the translation before planning. I could also compute this from the `reached_pose` (or go from `link_pose` to `reached_pose`) inside the method and have a parameter less, but it seemed more robust and concise to me this way.

core/src/stages/move_relative.cpp
```diff
@@ -104,6 +104,63 @@ static bool getJointStateFromOffset(const boost::any& direction, const moveit::c
...
+	auto linear_norm = linear.norm();
+
+	// rotation of the target direction and for the cylidner marker
+	auto quat_target = Eigen::Quaterniond::FromTwoVectors(Eigen::Vector3d::UnitX(), linear);
+	auto quat_cylinder = quat_target * Eigen::AngleAxisd(0.5 * M_PI, Eigen::Vector3d::UnitY());
+
+	// link position before planning; reached link position after planning; target link position
+	Eigen::Vector3d pos_link = link_pose.translation();
+	Eigen::Vector3d pos_reached = reached_pose.translation();
+	Eigen::Vector3d pos_target = pos_reached + quat_target * Eigen::Vector3d(linear_norm - distance, 0, 0);
```
v4hn, line 119:
```suggestion
	auto& pos_link = link_pose.translation();
	auto& pos_reached = reached_pose.translation();
	auto pos_target = pos_reached + quat_target * Eigen::Vector3d(linear_norm - distance, 0, 0);
```
... is what I would write, because the explicit types will actually evaluate Eigen's templated lazy evaluation before it's actually needed (the auto-deduced types are not Vector3d, but some `EigenBase<...>` datatypes). It won't make a difference here though, so feel free to leave it as is. See for example [here](https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html) for something to start reading about the subject.

rviz_marker_tools/src/marker_creation.cpp
```diff
@@ -276,6 +276,22 @@ vm::Marker& makeMesh(vm::Marker& m, const std::string& filename, double sx, doub
 	return m;
 }
 
+vm::Marker& makeArrow(vm::Marker& m, const Eigen::Vector3d& start_point, const Eigen::Vector3d& end_point, double diameter, double head_length) {
+	// scale.y is set according to default proportions in rviz/default_plugin/markers/arrow_marker.cpp#L61
+	// for the default head_length=0, the head length will keep the default proportion defined in arrow_marker.cpp#L106
+	m.scale.x = diameter;
+	m.scale.y = 2 * diameter;
+	m.scale.z = head_length;
+	prepareMarker(m, vm::Marker::ARROW);
```
v4hn, line 285:
```suggestion
	prepareMarker(m, vm::Marker::ARROW);
	// scale.y is set according to default proportions in rviz/default_plugin/markers/arrow_marker.cpp#L61
	// for the default head_length=0, the head length will keep the default proportion defined in arrow_marker.cpp#L106
	m.scale.x = diameter;
	m.scale.y = 2 * diameter;
	m.scale.z = head_length;

Comment (j-kuehn) [reply rviz_marker_tools/src/marker_creation.cpp:285]: I put `prepareMarker` after the `scale` assignments, because it ensures a valid scale. Currently, this is not necessary, as `diameter` and `head_length` are derived from `linear_norm`, which is verified to be non-zero for the visualization.

However, if this should ever change in the future, I thought it's better to have that scale check, especially since all other methods call `prepareMarker` after the `scale` assignments to ensure a valid scale.

File: core/src/stages/move_relative.cpp (hunk 4 above)
Comment (j-kuehn) [reply core/src/stages/move_relative.cpp:119]: I am unsure why this doesn't work, but I tried it with `auto` and `auto&` before and just did again and it breaks the visualization. Since I simply pass the variables to `tf::pointEigenToMsg`, it has to be something in there, but I didn't do any further troubleshooting.

**Review by v4hn — APPROVED** [review]

File: rviz_marker_tools/src/marker_creation.cpp [hunk 6]
```diff
@@ -276,6 +276,20 @@ vm::Marker& makeMesh(vm::Marker& m, const std::string& filename, double sx, doub
...
+                      double diameter, double head_length) {
+	// scale.y is set according to default proportions in rviz/default_plugin/markers/arrow_marker.cpp#L61
+	// for the default head_length=0, the head length will keep the default proportion defined in arrow_marker.cpp#L106
+	m.scale.x = diameter;
+	m.scale.y = 2 * diameter;
+	m.scale.z = head_length;
+	prepareMarker(m, vm::Marker::ARROW);
+	m.points.resize(2);
+	tf::pointEigenToMsg(start_point, m.points[0]);
+	tf::pointEigenToMsg(end_point, m.points[1]);
```
rhaschke, line 289: It would have been better to directly use tf2 functions. See https://github.com/ros-planning/moveit_task_constructor/issues/78#issuecomment-826788870. But I guess, Michael will give you this as a next  task anyway :wink:

rhaschke (approved):

rhaschke:
Very nice solution and visualization!

PR #264, opened 2021-05-10
Add visualization for Point goals

j-kuehn:
Implementation for this [TODO](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/move_to.cpp#L182).

The `MoveTo` stage can now visualize both Point and Pose goals. The frame for the target is now derived from `target_eigen` instead of the target message itself, because `target_eigen` is computed in both `getPoseGoal` and `getPointGoal`.


rhaschke:
> I wish this was something the planners did in MoveIt by default.

There isn't even a topic to publish such markers.

felixvd:
A sad state of affairs indeed. I think it should just go to `failed_plans` with namespaces. I wanted to add it to a `PlanningPipeline` recently, but then I remembered that the planner usually calls IK for pose goals and should publish it at that point instead.

v4hn:
@felixvd just use MTC :clownface:

PR #266, opened 2021-05-18
[WIP] Add pick-place capability

karolyartur:
This PR is to continue the work from PR #111

The aim is to provide a Pick-Place planning pipeline based on MTC.

During the design, I tried to keep the key concepts and architecture mentioned in Issue #112


felixvd (commented):
Thanks a lot for this big chunk of work! Guess I will have to get the `move_group_interface` side of this ready as promised.

I was testing this on our system right now and noticed that the applicability test stage should return `true` if the `object_name` of the plan request is empty. That would allow planning pick/place motions for a given set of grasps while ignoring the collision objects in the scene, and who are we to tell the user that they can't do that? Flexibility is good, and it reduces the obstacles until they can start planning. 

Apart from that, I went through and left some comments, but I've seen this code before, so it's nothing new.

bryceikeda:
Hi, I'm wondering what the status is of the PR?

karolyartur:
Hi @bryceikeda!

Unfortunately, I didn't have the time to follow up on this PR. I have a student currently who would pick this up from here, so there might be some updates in a few months.

PR #270, opened 2021-05-25
Allow to consider specific joints only in PathLength

j-kuehn:
Implements [this](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/include/moveit/task_constructor/cost_terms.h#L118) TODO.

By passing a vector of joint names to the constructor of `PathLength`, the cost of the path is now computed by only considering these joints. If no vector is passed, all joints are considered.


v4hn (approved):

v4hn:
Thanks! :-)

PR #268, opened 2021-05-19
GHA: Address Sanitizer

rhaschke:
Looking into the issues described in https://github.com/ros-planning/moveit_task_constructor/pull/254#pullrequestreview-662459305, I enabled ASan and stumbled over this other issue:

```
==10713==ERROR: AddressSanitizer: heap-use-after-free on address 0x6120000035bc at pc 0x7fda140a81a4 bp 0x7ffeed3fc060 sp 0x7ffeed3fc050
    READ of size 4 at 0x6120000035bc thread T0
        #0 0x7fda140a81a3 in moveit::task_constructor::InterfaceState::Priority::operator<(moveit::task_constructor::InterfaceState::Priority const&) const (/home/runner/work/target_ws/build/moveit_task_constructor_core/devel/lib/libmoveit_task_constructor_core.so+0x1ca1a3)
        #1 0x7fda140b7971 in std::_List_iterator<moveit::task_constructor::InterfaceState*> std::__upper_bound<std::_List_iterator<moveit::task_constructor::InterfaceState*>, moveit::task_constructor::InterfaceState*, __gnu_cxx::__ops::_Val_comp_iter<ValueOrPointeeLess<moveit::task_constructor::InterfaceState*, bool> > >(std::_List_iterator<moveit::task_constructor::InterfaceState*>, std::_List_iterator<moveit::task_constructor::InterfaceState*>, moveit::task_constructor::InterfaceState* const&, __gnu_cxx::__ops::_Val_comp_iter<ValueOrPointeeLess<moveit::task_constructor::InterfaceState*, bool> >) (/home/runner/work/target_ws/build/moveit_task_constructor_core/devel/lib/libmoveit_task_constructor_core.so+0x1d9971)
        #2 0x7fda140a9a1c in moveit::task_constructor::Interface::add(moveit::task_constructor::InterfaceState&) (/home/runner/work/target_ws/build/moveit_task_constructor_core/devel/lib/libmoveit_task_constructor_core.so+0x1cba1c)
        #3 0x7fda13f68dc7 in moveit::task_constructor::ContainerBasePrivate::liftSolution(std::shared_ptr<moveit::task_constructor::SolutionBase> const&, moveit::task_constructor::InterfaceState const*, moveit::task_constructor::InterfaceState const*) (/home/runner/work/target_ws/build/moveit_task_constructor_core/devel/lib/libmoveit_task_constructor_core.so+0x8adc7)
[...]

PR #254, opened 2021-04-14
run full demo as a test

v4hn:
Running that would have helped multiple times before.

Also simplify/streamline code in various places.

PR #271, opened 2021-05-27
Fix some memory issues pointed out by sanitizer

v4hn:
Related to #268

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -18,6 +18,7 @@ catkin_package(
 	LIBRARIES
 		${PROJECT_NAME}
 		${PROJECT_NAME}_stages
+		${PROJECT_NAME}_stage_plugins
 	INCLUDE_DIRS
 		include
 	CATKIN_DEPENDS
--- a/core/motion_planning_stages_plugin_description.xml
+++ b/core/motion_planning_stages_plugin_description.xml
@@ -1,4 +1,4 @@
-<library path="libmoveit_task_constructor_core_stages">
+<library path="libmoveit_task_constructor_core_stage_plugins">
   <class name="moveit_task_constructor/Current State"
          type="moveit::task_constructor::stages::CurrentState"
          base_class_type="moveit::task_constructor::Stage">
--- a/core/src/stages/CMakeLists.txt
+++ b/core/src/stages/CMakeLists.txt
@@ -41,6 +41,11 @@ add_library(${PROJECT_NAME}_stages
 )
 target_link_libraries(${PROJECT_NAME}_stages ${PROJECT_NAME} ${catkin_LIBRARIES})
 
-install(TARGETS ${PROJECT_NAME}_stages
+add_library(${PROJECT_NAME}_stage_plugins
+   plugins.cpp
+)
+target_link_libraries(${PROJECT_NAME}_stage_plugins ${PROJECT_NAME}_stages ${catkin_LIBRARIES})
+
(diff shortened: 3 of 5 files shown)


rhaschke (commented):
I approve the 1st and 3rd commits but don't understand why the 2nd one is necessary and why only for CurrentStage? Can you explain?

rhaschke:
Thanks for the explanation. You are right, I wanted to enable GUI-based assembly of tasks in the past...
I think, locally, I'm using a suppression file to ignore the known ClassLoader issues. However, interestingly, GHA didn't complain either, which made me wonder, why these changes are necessary.
https://github.com/ros-planning/moveit_task_constructor/actions/runs/883749251

Actually, I don't like this workaround very much but suggest using a suppression file instead. There is an explicit reminder that suppressions were used so that the issue isn't silently hidden.

But there is another issue: As I merged #254 _after_ having merged the (originally successful) asan checks, the build fails now, because `move_group` wasn't compiled against libasan.so, but loads a plugin that was:
https://github.com/ros-planning/moveit_task_constructor/runs/2689556141

A possible workaround is to `LD_PRELOAD=libasan.so.x` via an `<env>` setting in the `.test` file.
However, this is fragile, because the version of libasan will change across different OSes. Removing the asan check, isn't a nice option either.

v4hn:
> However, interestingly, GHA didn't complain either

I was wondering why that did not come up, yes.

> Actually, I don't like this workaround very much but suggest using a suppression file instead. There is an explicit reminder that suppressions were used so that the issue isn't silently hidden.

This is not a workaround at all in my opinion but simple cleanup. Using/linking a regular C++ class should not invoke magic code to add entries to unused(!) global datastructures that even leak memory. I was surprised to find that you added the macro in the regular cpp file of the class.

Incidentally I just followed [a suggestion in a potential warning of class loader](https://github.com/ros/class_loader/blob/melodic-devel/include/class_loader/class_loader_core.hpp#L222-L225). I've never seen the warning, but it looks like there can(could?) actually be issues if you link a library with class-loadable datastructures.

It would still be nice to suppress the same warnings in the new library if we ever write tests for it.

> because move_group wasn't compiled against libasan.so, but loads a plugin that was
> A possible workaround is to LD_PRELOAD=libasan.so.x via an <env> setting in the .test file.

Yes, I've noticed similar issues locally (same with RViz and the MTC display/panel, though these are not relevant for CI tests).
I did not know about the workaround though.
We could disable the flag for plugins (though that's likely somewhat cumbersome).
If the workaround will not result in any misbehavior (I guess it might be problematic if a plugin exchanges ownership of allocated memory with the main node?) couldn't we just add the environment flag to the `asan` job?
Alternatively we could of course also add another environment flag to detect in the respective `.test` files.
I don't see much of a problem if we can define the library version in the `ci.yaml`.

JafarAbdi:
> You are right, I wanted to enable GUI-based assembly of tasks in the past...

Are you still planning to do so.? I'm asking because one thing I've been working on during the last few lockdowns is to create a similar application, this's the current status

https://user-images.githubusercontent.com/16278108/123803939-69e2f200-d8f5-11eb-8831-9a44fe877caa.mp4

As you can see it's not ready yet, but I do still slowly work on it when I have free time, 

I wonder if you have an architect document that may help me during the development

rhaschke:
@JafarAbdi: See https://github.com/ros-planning/moveit_task_constructor/issues/278.

PR #273, opened 2021-06-03
make CI jobs more picky & fix pointed out bugs

v4hn:
And add warnings for gcc AND clang builds (as they can differ slightly).

Clang does not support -Wused-but-set-parameter yet, though
support is [on the way](https://reviews.llvm.org/D100581).
So this gets rid of this warning we see in CI:

> error: unknown warning option '-Wno-unused-but-set-parameter'; did you mean '-Wno-unused-parameter'?

Also address various new warnings that are produced now.


v4hn:
I'm down to a single issue which is that OGRE should be included as `SYSTEM`. But it's added by RViz.
Can you have a look please @rhaschke?

rhaschke:
> I'm down to a single issue which is that OGRE should be included as `SYSTEM`. But it's added by RViz.

Yeah, that's a severe limitation of the catkin build system: Individual `catkin_include_DIRS` cannot be declared as `SYSTEM`. Usually, I solve such issues by including all external stuff as SYSTEM. But this has other side effects, if not done consistently.
I pushed a corresponding commit that fixes the open issue. Additionally, I had to limit the strict compile flags to the target workspace. Otherwise, some [issues in `mtc_pour` in the downstream workspace break the build](https://github.com/ubi-agni/moveit_task_constructor/runs/2762551937?check_suite_focus=true).

rhaschke (approved):

v4hn:
> Individual catkin_include_DIRS cannot be declared as SYSTEM.

Sure they can, if you can identify them in the list.
Including the whole ROS ecosystem this way is completely wrong to me and I dislike doing so a lot - it probably works though as long as *everything* is included as SYSTEM.
I just created #274 with the alternative (and from my perspective preferable) approach to work around the catkin limitation.

>  I pushed a corresponding commit that fixes the open issue. Additionally, I had to limit the strict compile flags to the target workspace. Otherwise, some issues in mtc_pour in the downstream workspace break the build.

Yeah, I was just following up on this when I saw your commits and hopefully just fixed the downstream issues as well.
Given our current downstream is completely under our control I would like to leave the flags enabled for it as well.

rhaschke:
> Including the whole ROS ecosystem this way is completely wrong to me and I dislike doing so a lot

Why? Everything external to the source in question should be considered as external and thus as `SYSTEM`.

> it probably works though as long as _everything_ is included as SYSTEM.

If you mean with _everything_ everything external, I agree.

v4hn:
merged #274 instead.

PR #274, opened 2021-06-07
Stricter CI / alternative Ogre handling

v4hn:
An alternative for #273 where @rhaschke proposed to include everything as `SYSTEM` includes.

I also fixed the mtc_pour issues pointed out by the stricter CI, so there should not be any reason to restrict the flags to the target workspace now.


visualization/CMakeLists.txt
```diff
@@ -10,6 +10,13 @@ find_package(catkin REQUIRED COMPONENTS
 	rviz
 )
 
+# rviz transitively includes OGRE headers which break with `-Wall -Werror`
+# so isolate this include dir and add it as SYSTEM include where needed.
+find_path(OGRE_INCLUDE_DIR OgreRoot.h
```
rhaschke, line 15: Note, that rviz currently includes two OGRE folders: `/usr/include/OGRE/Overlay` and `/usr/include/OGRE`.
I think you should handle both in the same fashion. At first sight, I considered this approach elegant.
However, it is also brittle, because any change of OGRE include folders in rviz, will probably break here.

visualization/motion_planning_tasks/src/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ target_include_directories(${MOVEIT_LIB_NAME}
 	PUBLIC $<TARGET_PROPERTY:motion_planning_tasks_properties,INTERFACE_INCLUDE_DIRECTORIES>
 	PUBLIC $<TARGET_PROPERTY:moveit_task_visualization_tools,INTERFACE_INCLUDE_DIRECTORIES>
 	PUBLIC ${catkin_INCLUDE_DIRS}
+	PUBLIC SYSTEM ${OGRE_INCLUDE_DIR}
```
rhaschke, line 41: This syntax is not possible. You need to add another `target_include_directories` command with the `SYSTEM` keyword present directly after the target name. I ran into the same issue...

visualization/visualization_tools/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ target_link_libraries(${MOVEIT_LIB_NAME}
 target_include_directories(${MOVEIT_LIB_NAME}
   PUBLIC include
   PRIVATE ${catkin_INCLUDE_DIRS}
+  PRIVATE SYSTEM ${OGRE_INCLUDE_DIR}
```
rhaschke, line 37: Same here.

rhaschke (commented):
I'm not sure that this approach is better than the one proposed in #273. See comments below.
If you get it working, I'm fine to merge though. It's rather improbable that rviz will change its OGRE include folders.

visualization/motion_planning_tasks/src/CMakeLists.txt
v4hn: I thought I tested this locally, but apparently it doesn't even work in cmake 3.19 over here...
Fixing it.

PR #275, opened 2021-06-12
Implement missing pregrasp property for RobotState.msg 

captain-yoshi:
The `pregrasp` property segfaults when a RobotState.msg is passed to the GenerateGraspPose stage like so:

```cpp
moveit_msgs::RobotState robot_state;
robot_state.is_diff = true;
robot_state.joint_state.name.resize(1);
robot_state.joint_state.name[0] = "panda_finger_joint1";
robot_state.joint_state.position.resize(1);
robot_state.joint_state.position[0] = 0.0;

{
	// Sample grasp pose
	auto stage = std::make_unique<stages::GenerateGraspPose>("generate grasp pose");
	stage->properties().configureInitFrom(Stage::PARENT);
	stage->properties().set("marker_ns", "grasp_pose");
	stage->setPreGraspPose(robot_state);
        ...
}
```
This PR may not be the best way to report the bad_any_cast. Also the robot state could fail silently in the `compute` section but is dealt with in the init section. Not sure if a property can be change after the init section.

### DIFF
--- a/core/src/stages/generate_grasp_pose.cpp
+++ b/core/src/stages/generate_grasp_pose.cpp
@@ -126,14 +126,7 @@ void GenerateGraspPose::onNewSolution(const SolutionBase& s) {
 	const std::string& object = props.get<std::string>("object");
 	if (!scene->knowsFrameTransform(object)) {
 		const std::string msg = "object '" + object + "' not in scene";
-		if (storeFailures()) {
-			InterfaceState state(scene);
-			SubTrajectory solution;
-			solution.markAsFailure();
-			solution.setComment(msg);
-			spawn(std::move(state), std::move(solution));
-		} else
-			ROS_WARN_STREAM_NAMED("GenerateGraspPose", msg);
+		spawn(InterfaceState{ scene }, SubTrajectory::failure(msg));
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


core/src/stages/generate_grasp_pose.cpp
```diff
@@ -123,7 +151,17 @@ void GenerateGraspPose::compute() {
...
-	robot_state.setToDefaultValues(jmg, props.get<std::string>("pregrasp"));
+	try {
+		// try named joint pose
+		robot_state.setToDefaultValues(jmg, props.get<std::string>("pregrasp"));
+	} catch (const boost::bad_any_cast&) {
+	}
+	try {
+		// try RobotState
+		const moveit_msgs::RobotState& robot_state_msg = props.get<moveit_msgs::RobotState>("pregrasp");
+		robotStateMsgToRobotState(robot_state_msg, robot_state);
```
captain-yoshi, line 162: Set `copy_attached_bodies = false` instead of the default which is true?

v4hn:
I just pushed some improvements. Please have a look @captain-yoshi .
From my perspective this can be merged once CI suceeds.

captain-yoshi:
It works on my end with named pose and RobotState message. Your solution is much cleaner!

core/src/stages/generate_grasp_pose.cpp
```diff
@@ -58,6 +59,30 @@ GenerateGraspPose::GenerateGraspPose(const std::string& name) : GeneratePose(nam
...
+		}
+		return;
+	} catch (const boost::bad_any_cast&) {
+	}
+
+	try {
+		// try RobotState
+		const moveit_msgs::RobotState& robot_state_msg = boost::any_cast<moveit_msgs::RobotState>(diff_property.value());
+		// might throw on invalid message
+		robotStateMsgToRobotState(robot_state_msg, state);
```
rhaschke, line 78: Should we ensure here that only joints of `jmg` are changed?

rhaschke (approved):
I approve. However, please see the question below.

rhaschke (approved):

Issue #276, opened 2021-06-23
Set desired/max Cartesian end-effector speed


gautz:
[``setMaxCartesianEndEffectorSpeed()`` should soon be available through the MoveGroup interface](https://github.com/ros-planning/moveit/pull/2674). Which would be the proper way to get a similar functionality in e.g. a moveTo stage? Is there some plan on integrating this?

This is following my issue on using MTC for welding #251

rhaschke:
Velocity-related stuff is not configured in individual stages, but via the properties of a `PlannerInterface`:
https://github.com/ros-planning/moveit_task_constructor/blob/ee6c50ad31402b00fc4b8a0d8fcd62b1c85c27b7/core/src/solvers/planner_interface.cpp#L47-L48

These are finally used to configure the `MotionPlanRequest`:
https://github.com/ros-planning/moveit_task_constructor/blob/4fa8c10f4456e893e09212b51a06b3d2ec4e9266/core/src/solvers/pipeline_planner.cpp#L150-L151

However, as these settings are only passed to pipeline planners, they will not become effective for Cartesian motions!

v4hn:
> However, as these settings are only passed to pipeline planners, they will not become effective for the CartesianPath planner!

So basically, the max Cartesian speed parameter can be added to the `PipelinePlanner` straight away and *all* standard velocity parameters *could* be added to the other solvers as well. Adding the EEF Cartesian velocity to `CartesianPath` is quite a good idea I think.

Also, I would provide a `WrapperBase` container that can adapt all three time parameters for arbitrary stages.

Issue #210, opened 2020-09-23
moveit default_velocity_scaling_factor is ignored


JonasTietz:
Hello,
I noticed that the default_velocity_scaling_factor is ignored when planning trajectories with the pipeline planner, which is suprising at least. Especially if you tuned the joint_limits.yaml to be save at the default_velocity_scaling_factor.

rhaschke:
Configurable scaling factors were introduced in https://github.com/ros-planning/moveit/pull/1890 after the corresponding MTC code was written. In MTC, the scaling factors are configured as (constant) properties here:
https://github.com/ros-planning/moveit_task_constructor/blob/d3b878a31c63a5ceaead481b9c536cff929a9941/core/src/solvers/planner_interface.cpp#L45-L49

A pull request replacing this by a lookup of default scaling factors from ROS parameters is highly welcome.
I'm not yet sure how to pass the move_group's namespace to access the correct parameters though.

v4hn:
I discussed this with @JonasTietz yesterday and was wondering whether we want to support this in MTC at all.
You are right, we can add support for this to the PlannerInterface classes, but doing so will ignore all custom stages and that might be *more* surprising than leaving it out everywhere.

We could also tackle this by making time parameterization an explicit parameter in *each stage*, additionally supporting different algorithms for that. I'm not sure I like the associated overhead though and it would add more parameters if you want to auto compute a "correct" dynamics profile for some stages, but allow the stage to compute their own profile if they know better (pouring is a good example for this :))

Lastly, we could also add default properties for this similar to the `marker_ns` property.

rhaschke:
Actually, I like the idea of adding time parameterization as an explicit, configurable property of stages. Some stages, e.g. the Merger or merger Connect stages, don't have time parameterization (or it is hardcoded for a specific method).
But I agree, that only configuring this for `PlannerInterface` classes, might be not sufficient.

PR #277, opened 2021-06-25
Set max cartesian end-effector speed in planning pipeline

gautz:
Enable trajectory post-processing in planning pipeline to set maximum cartesian end-effector speed.
Requires following to be merged:
- https://github.com/ros-planning/moveit/pull/2674/
- https://github.com/ros-planning/moveit_msgs/pull/113

How to use? Instead of ``setProperty("max_velocity_scaling_factor", velocity);``, use:
```
pipeline_planner->setProperty("max_cartesian_speed", velocity); // cartesian_velocity
pipeline_planner->setProperty("cartesian_speed_end_effector_link", welding_tcp_frame_);
```
I will also implement it in the cartesian demo.

Solves #276

### DIFF
--- a/core/src/solvers/pipeline_planner.cpp
+++ b/core/src/solvers/pipeline_planner.cpp
@@ -40,6 +40,7 @@
 #include <moveit/task_constructor/task.h>
 #include <moveit/planning_scene/planning_scene.h>
 #include <moveit/planning_pipeline/planning_pipeline.h>
+#include <moveit/trajectory_processing/cartesian_speed.h>
 #include <moveit_msgs/MotionPlanRequest.h>
 #include <moveit/kinematic_constraints/utils.h>
 #include <eigen_conversions/eigen_msg.h>
@@ -149,6 +150,8 @@ void initMotionPlanRequest(moveit_msgs::MotionPlanRequest& req, const PropertyMa
 	req.num_planning_attempts = p.get<uint>("num_planning_attempts");
 	req.max_velocity_scaling_factor = p.get<double>("max_velocity_scaling_factor");
 	req.max_acceleration_scaling_factor = p.get<double>("max_acceleration_scaling_factor");
+	req.max_cartesian_speed = p.get<double>("max_cartesian_speed");
+	req.cartesian_speed_end_effector_link = p.get<std::string>("cartesian_speed_end_effector_link");
 	req.workspace_parameters = p.get<moveit_msgs::WorkspaceParameters>("workspace_parameters");
 }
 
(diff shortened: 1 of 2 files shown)


gautz:
I tested this successfully in the ``pick_place_task`` to set velocity of the hand when moving the ``panda_arm`` with the pipeline planner. For instance I can set:
- Move-to pre-grasp with max ``free_space_velocity_=1`` [m/s]
- Move-to place with max ``approach_velocity_=0.1`` [m/s] 

Do you see a better pipeline planner demo where to set the max cartesian velocity?

v4hn:
> Do you see a better pipeline planner demo where to set the max cartesian velocity?

Instead of squeezing *all* features into the one demo, I would encourage you to just create a new demo, possibly something that resembles a simple welding trajectory.

gautz:
> > Do you see a better pipeline planner demo where to set the max cartesian velocity?
> 
> Instead of squeezing _all_ features into the one demo, I would encourage you to just create a new demo, possibly something that resembles a simple welding trajectory.

Makes sense, we will probably propose a welding demo, but rather in middle-long term. 

In shorter term I could propose to adapt the cartesian Demo to use the pipeline planner (Pilz and/or OMPL constrained planning) and set different cartesian velocities in there.

core/src/solvers/pipeline_planner.cpp
```diff
@@ -168,6 +171,11 @@ bool PipelinePlanner::plan(const planning_scene::PlanningSceneConstPtr& from,
 	::planning_interface::MotionPlanResponse res;
 	bool success = planner_->generatePlan(from, req, res);
 	result = res.trajectory_;
+	// optionally compute timing to move the eef with constant speed
+	if (req.max_cartesian_speed > 0.0) {
+		trajectory_processing::setMaxCartesianEndEffectorSpeed(*result, req.max_cartesian_speed,
+		                                                       req.cartesian_speed_end_effector_link);
+	}
```
rhaschke, line 178: As I understood https://github.com/ros-planning/moveit/pull/2674, you implemented a planning adapter. Hence, there is no reason to explicitly call this function here (and below). It's automatically handled by the adapter - if configured.
The `CartesianPath` planner on the other hand could benefit from this explicit call :wink:

rhaschke (changes requested):
In principle looks good to me. See the remark below.

core/src/solvers/pipeline_planner.cpp
gautz: max cartesian speed can now be set after configuring the adapter https://github.com/ros-planning/moveit_resources/pull/88

gautz:
I tried to add the explicit call in cartesian_path but I do not manage to get the velocity set in the cartesian demo.

gautz:
@rhaschke when using the ``cartesian_path`` planner, neither setting ``max_velocity_scaling_factor`` nor  ``max_cartesian_speed`` results in a change in robot velocity. Could it be somehow related to the discussions in #210 ?

PR #279, opened 2021-07-04
Merge different mockup implementations

v4hn:
Makes a lot of sense in general and simplifies independent future tests.

I guess we could unify more of the infrastructure
(TestBase, Standalone<>, append...)
but this is the most important part to have around.

Also fix an unnoticed bug that was triggered during development on the side.

Thanks to @j-kuehn for the initial effort.

### DIFF
--- a/core/test/CMakeLists.txt
+++ b/core/test/CMakeLists.txt
@@ -6,7 +6,7 @@
 if (CATKIN_ENABLE_TESTING)
 	find_package(rostest REQUIRED)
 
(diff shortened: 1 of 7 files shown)


rhaschke (commented):
Thanks a lot for this cleanup. I was thinking of that several times before, but never found time for it.

v4hn:
> I was thinking of that several times before, but never found time for it.

Same here... Giving things to a student who struggles with it for a while is a good incentive. ;)

PR #280, opened 2021-07-12
Test cases for Fallbacks container

j-kuehn:
I added three test cases for the `Fallbacks` container:

- `ConnectStageInsideFallbacks` shows that adding a `Connect` stage to the `Fallbacks` container seems to be broken, as no solution can be computed for the task currently.
- `ComputeFirstSuccessfulStageOnly` suggests that solutions inside the `Fallbacks` container should not be computed after one solution has been found. I'm not sure how the container was originally intended to work, so this might not be a bug.
- `ActiveChildReset` reveals the failure of the container to reset the [`active_child_`](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/include/moveit/task_constructor/container.h#L161) for every spawned solution from a generator. In this test case, the first `ForwardMockup` does not become the `active_child_` for the third spawned solution from the `GeneratorMockup`.

### DIFF
--- a/core/test/test_container.cpp
+++ b/core/test/test_container.cpp
@@ -725,13 +725,13 @@ TEST(Fallback, ActiveChildReset) {
 	Task t;
 	t.setRobotModel(getModel());
 
-	t.add(std::make_unique<GeneratorMockup>(std::initializer_list<double>({ 0.0, 0.0 })));
+	t.add(std::make_unique<GeneratorMockup>(PredefinedCosts{ { 0.0, INF, 0.0 } }));
 
 	auto fallbacks = std::make_unique<Fallbacks>("Fallbacks");
-	fallbacks->add(std::make_unique<ForwardMockup>(std::initializer_list<double>({ INF, 0.0 })));
-	fallbacks->add(std::make_unique<ForwardMockup>(std::initializer_list<double>({ 0.0, INF })));
+	fallbacks->add(std::make_unique<ForwardMockup>(PredefinedCosts::constant(0.0)));
+	fallbacks->add(std::make_unique<ForwardMockup>(PredefinedCosts::constant(0.0)));
 	t.add(std::move(fallbacks));
 
 	EXPECT_TRUE(t.plan());
-	EXPECT_EQ(t.numSolutions(), 2u);
+	EXPECT_EQ(t.numSolutions(), 4u);
 }
--- a/core/test/test_container.cpp
+++ b/core/test/test_container.cpp
@@ -44,6 +44,7 @@ void append(ContainerBase& c, const std::initializer_list<StageType>& types) {
 		}
 	}
 }
+constexpr double INF = std::numeric_limits<double>::infinity();
 
 class NamedStage : public GeneratorMockup
 {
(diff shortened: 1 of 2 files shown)


core/test/test_container.cpp
```diff
@@ -685,3 +686,52 @@ TEST(Fallback, failing) {
...
+
+	t.add(std::make_unique<GeneratorMockup>(PredefinedCosts{ { 0.0, INF, 0.0 } }));
+
+	auto fallbacks = std::make_unique<Fallbacks>("Fallbacks");
+	fallbacks->add(std::make_unique<ForwardMockup>(PredefinedCosts::constant(0.0)));
+	fallbacks->add(std::make_unique<ForwardMockup>(PredefinedCosts::constant(0.0)));
+	t.add(std::move(fallbacks));
+
+	EXPECT_TRUE(t.plan());
+	EXPECT_EQ(t.numSolutions(), 4u);
```
v4hn, line 736: Depending on the preferred semantics (whether fallback alternatives should be computed at all), the test is still not correct: If no alternatives should be considered you would have to check for *two* solutions where both should come from the first `ForwardMockup`.

Let's wait for Robert's response before modifying this.

v4hn (commented):
@j-kuehn Thank you for the discussed tests! I would appreciate if you could provide a fix for `ActiveChildReset` as well. :cat2: 

@rhaschke I would appreciate your opinion on the second issue. I strictly prefer the behavior in the test (don't run alternatives if a solution succeeds) but if you prefer we can add an option to toggle behaviors instead.

core/test/test_container.cpp
rhaschke: I don't yet fully understand the purpose of this test. I would expect two solutions here, contributed by the _first_ child. However, this was not yet tested here. I pushed a corresponding fixup.

core/test/test_container.cpp
```diff
@@ -685,3 +686,52 @@ TEST(Fallback, failing) {
 	EXPECT_FALSE(t.plan());
 	EXPECT_EQ(t.solutions().size(), 0u);
 }
+
+TEST(Fallback, ConnectStageInsideFallbacks) {
```
rhaschke, line 690: This fails for the following reason:
1. The `Generator` spawns a solution, which is correctly forwarded to the start pull interface of `Connect`.
2. However, because there was no state pushed into the end pull interface of `Connect`, `canCompute()` returns false
and the `Fallbacks` container forwards to the next child (which doesn't exist).

This can only be solved, if we drop the simple linear execution order and switch to topological execution order.

rhaschke (approved):
Thanks for providing these tests. See detailed comments below.
I suggest to disable/skip `ConnectStageInsideFallbacks` for now as it won't easily fix.
@v4hn: Do you already have concrete plans, how to implement a topological processing order?

core/test/test_container.cpp
rhaschke: There should be another test case, where the external generator is failing first before succeeding.
Currently, I would expect this test to fail (with a solution of the wrong, second child).

rhaschke:
Ping @v4hn, @j-kuehn. Any thoughts on my feedback?

core/test/test_container.cpp
j-kuehn: The test suggested that after computing solutions for the second spawned solution from the generator, the `active_child_` would not be reset to the first child for computing the third solution. Before your two commits, this led to the first child being ignored for the third spawned generator solution.

As @v4hn pointed out, this test was not consistent with the proposed functionality in `ComputeFirstSuccessfulStageOnly`, but rather tried to test the status quo. But even after your fix for the `ActiveChildReset` test, I would have expected the test to fail, because as I understand it, the `active_child_` is pointing to the second child after the second spawned generator solution. So for the third solution, I expected the container to use the second child for computation. However, the test passes, so it is actually using the first child. I'm not sure how your commit fixed this.

I just tried adding another test as you suggested, where the external generator is failing first before succeeding:
```cpp
std::make_unique<GeneratorMockup>(PredefinedCosts{ { INF, 0.0 } }));
```
However, this yields 0 solutions and I'm not sure why.

core/test/test_container.cpp
j-kuehn: Sounds reasonable to disable this test for now.

v4hn (commented):
Thanks for the feedback. I wanted to leave it to @j-kuehn to implement a fix for at least the simple test, but it looks like we disagree on what exactly should happen? Please give your reasoning to my comment below.

v4hn:
superseded by #287

PR #286, opened 2021-08-21
Fix some issues with ComputeIK

### LINKED ISSUE #209: ComputeIk sometimes fails to check collision
Hello,
I have noticed that sometimes the solutions of the ComputeIk node are definitely in collision but get reported as success.

ik_bug

For me it is not a big problem right now, because the collision gets detected once a trajectory is planned to that pose.

v4hn:
Fixes #209

### DIFF
--- a/core/src/stages/compute_ik.cpp
+++ b/core/src/stages/compute_ik.cpp
@@ -37,6 +37,7 @@
 #include <moveit/task_constructor/stages/compute_ik.h>
 #include <moveit/task_constructor/storage.h>
 #include <moveit/task_constructor/marker_tools.h>
+#include <moveit/task_constructor/moveit_compat.h>
 
 #include <moveit/planning_scene/planning_scene.h>
 #include <moveit/robot_state/conversions.h>
@@ -295,7 +296,12 @@ void ComputeIK::compute() {
 				ROS_WARN_STREAM_NAMED("ComputeIK", "Unknown frame: " << ik_pose_msg.header.frame_id);
 				return;
 			}
-			const EigenSTL::vector_Isometry3d& tf = attached->getFixedTransforms();
+			const EigenSTL::vector_Isometry3d& tf =
+#if MOVEIT_HAS_OBJECT_POSE
+			    attached->getShapePosesInLinkFrame();
+#else
+			    attached->getFixedTransforms();
+#endif
 			if (tf.empty()) {
 				ROS_WARN_STREAM_NAMED("ComputeIK", "Attached body doesn't have shapes.");
 				return;
--- a/core/src/stages/compute_ik.cpp
+++ b/core/src/stages/compute_ik.cpp
@@ -405,7 +405,7 @@ void ComputeIK::compute() {
 
 			// set scene's robot state
 			robot_state::RobotState& solution_state = solution_scene->getCurrentStateNonConst();
-			solution_state.setJointGroupPositions(jmg, ik_solutions.back().data());
+			solution_state.setJointGroupPositions(jmg, ik_solutions[i].data());
 			solution_state.update();
 
 			InterfaceState state(solution_scene);
--- a/core/src/stages/compute_ik.cpp
+++ b/core/src/stages/compute_ik.cpp
@@ -89,22 +89,21 @@ namespace {
 
 // ??? TODO: provide callback methods in PlanningScene class / probably not very useful here though...
 // TODO: move into MoveIt core, lift active_components_only_ from fcl to common interface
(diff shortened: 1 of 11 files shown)


core/src/stages/compute_ik.cpp
```diff
@@ -231,10 +230,10 @@ void ComputeIK::compute() {
 	properties().performInitFrom(INTERFACE, s.start()->properties());
 	const auto& props = properties();
 
-	planning_scene::PlanningScenePtr sandbox_scene = s.start()->scene()->diff();
+	const planning_scene::PlanningSceneConstPtr& scene{ s.start()->scene() };
```
rhaschke, line 233: Can you explain why it wasn't sufficient to get a new scene diff?

core/src/stages/compute_ik.cpp
```diff
@@ -405,7 +405,7 @@ void ComputeIK::compute() {
 
 			// set scene's robot state
 			robot_state::RobotState& solution_state = solution_scene->getCurrentStateNonConst();
-			solution_state.setJointGroupPositions(jmg, ik_solutions.back().data());
+			solution_state.setJointGroupPositions(jmg, ik_solutions[i].data());
```
rhaschke, line 408: Good catch!

rhaschke (commented):
I don't understand the failure of #209 yet. Can you explain more, please?

v4hn:
I would prefer to fix the CI failures via version checks and going away from our `MOVEIT_MASTER` detection. But that depends on https://github.com/ros-planning/moveit/issues/2831 .

core/src/stages/compute_ik.cpp
v4hn: The main culprit of the bug was that `isTargetPoseColliding` [changes the ACM](https://github.com/ros-planning/moveit_task_constructor/pull/286/files#diff-eb4791e6fc52e29c2b9fb96a606928a1bf587989c5ebf29af186a50137ecc159L117) but the scene was reused later on.
The rest of the patch changes the logic a bit to use what is actually needed.

rhaschke: Now, I see. The `sandbox_scene` was reused for later collision checking:
https://github.com/ros-planning/moveit_task_constructor/blob/84ccd679d980e30b5365fa542f1fd0a72598b79b/core/src/stages/compute_ik.cpp#L368

rhaschke (approved):
I approve. Thanks a lot for hunting this down!
I guess you are already working on fixing the nasty API changes.

rhaschke (approved):
Approving. CI is still failing, because MoveIt's `noetic-devel` and `docker-master` branches were not forwarded and thus didn't have the version 1.1.6 yet. I did that just now. dockerhub is building the images now...

v4hn:
@rhaschke Do you know how long it takes/should take to rebuild the MoveIt containers after merging a change? CI still uses a version before https://github.com/ros-planning/moveit/pull/2832 (merged 13 hours ago).
@henningkayser mentioned last week that our docker images might not build at all right now (which would be why PickNik pushed forward with the GHA integration)

I could of course write proper feature detection in CMake logic, but I would prefer to keep things simple and refer to the release versions that include each change.

v4hn:
> CI is still failing, because MoveIt's `noetic-devel` and `docker-master` branches were not forwarded and thus didn't have the version 1.1.6 yet. I did that just now.

Thanks. Since when do we have a `docker-master` branch and where's the discussion on why it is needed?
I could have figured out the `noetic-devel` branch myself though. :confused:

rhaschke:
I guess we missed each other writing our comments. 
Regarding, Henning's comment: I don't know about any dockerhub issues. All images build fine - at least for MoveIt1.
For MoveIt2, dockerhub tries to build the new images we switched to ghcr.io. However, one cannot access those w/o logging in first. Hence, I suggest to switch back to the dockerhub repo (but still build images on GHA). @vatanaksoytezer, what do you think?

rhaschke:
CI succeeded compiling but fails on the `pick_place_test` because the `world` frame is not defined anymore (see https://github.com/ros-planning/moveit_resources/issues/89).
I already prepared a fix a while ago: https://github.com/ros-planning/moveit_resources/pull/92, which is still pending.

v4hn:
> Since when do we have a `docker-master` branch and where's the discussion on why it is needed?

@rhaschke did you miss my question or do you not have an answer right away?

> I already prepared a fix a while ago: ros-planning/moveit_resources#92, which is still pending.

I'll have another look. it's a much bigger change than just adding a frame back though.

rhaschke:
> Since when do we have a `docker-master` branch and where's the discussion on why it is needed?

Since roughly a year. See https://github.com/ros-planning/moveit/issues/2299.

PR #287, opened 2021-09-01
Fallbacks per solution

v4hn:
Here is my recent effort to get the tests proposed in #280 to pass by [what I consider correct behavior](https://github.com/ros-planning/moveit_task_constructor/pull/280#discussion_r676565782) for `Fallbacks`.

I also added further tests for various edge-cases.
Please have a look [at the set of tests](https://github.com/v4hn/moveit_task_constructor/blob/dc1c9ca04a1a2bffd5d47adadabe08b8c53b0040/core/test/test_fallback.cpp) first.

It turned out to be quite a bit more work than I anticipated, especially with Pruning involved...
@j-kuehn you might want to have a look to see what I ended up with. :-)


core/src/container.cpp
```diff
@@ -806,46 +814,144 @@ void Alternatives::onNewSolution(const SolutionBase& s) {
...
+	          (*spec.stage)->pimpl()->pullInterface(spec.dir),
+	          false);
+
+	const auto& stage { (*spec.stage)->pimpl() };
+
+	try {
+		// run ALL possible computations (on new state)
+		// this is needed to decide whether it should be passed to the next child too
+		while (stage->canCompute()){
+			stage->runCompute();
```
rhaschke, line 945: Same here: We should never run an exhaustive search on a single stage only.
Similar to generators, you could just keep track of the active child for each external states and only forward to the next child, if it returns `canCompute() = false`.

rhaschke (commented):
Thanks a lot for pushing this forward! I generally agree with the tests. 
However, I need to add some test cases to point out the undesired behaviour I noticed (see below).
I pushed some minor cosmetic changes as well.

core/test/stage_mockups.h
```diff
@@ -128,5 +129,13 @@ struct TaskTestBase : public testing::Test
...
 };
 
+#define EXPECT_COSTS(value, matcher)                                           \
+	{                                                                           \
+		std ::vector<double> costs;                                              \
+		std::transform(value.begin(), value.end(), std::back_inserter(costs),    \
+		               [](const SolutionBaseConstPtr& s) { return s->cost(); }); \
+		EXPECT_THAT(costs, matcher);                                             \
+	}
+
```
v4hn, line 139: Nice simplification, though I'm no big friend of preprocessor macros... Do functions hide the line from gtests output?

core/include/moveit/task_constructor/storage.h
```diff
@@ -82,8 +82,8 @@ class InterfaceState
 	enum Status
 	{
 		ENABLED,  // state is actively considered during planning
+		FAILED,  // state that failed, causing the whole partial solution to be disabled
 		DISABLED,  // state is disabled because a required connected state failed
```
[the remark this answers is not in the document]
v4hn: I had to reorder this because of [early stopping in `hasPendingOpposites`](https://github.com/ros-planning/moveit_task_constructor/pull/287/files#diff-b284e09d92fea0042ecfec277496e7e6fdccc48459e7a2d008c8be34bb33ad57R749).
It's a matter of perspective which one is "more severe" here. `FAILED` might be retried in `Connecting` stages, but `DISABLED` will never be looked at unless it becomes active again.

core/test/test_fallback.cpp
```diff
@@ -0,0 +1,143 @@
...
+	auto fwd1 = fallbacks->findChild("FWD1");
+	auto fwd2 = fallbacks->findChild("FWD2");
+	t.add(std::move(fallbacks));
+
+	EXPECT_TRUE(t.plan());
+	EXPECT_EQ(t.numSolutions(), 2u);
+	ASSERT_EQ(fwd1->solutions().size(), 1u);
+	EXPECT_EQ(fwd1->solutions().front()->start()->priority().cost(), 1.0);
+	ASSERT_EQ(fwd2->solutions().size(), 1u);
+	EXPECT_EQ(fwd2->solutions().front()->start()->priority().cost(), 0.0);
```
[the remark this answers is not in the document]
v4hn: yes. I believe that's the only intuitive interpretation for actual use-cases.

core/src/container.cpp
```diff
@@ -806,46 +814,144 @@ void Alternatives::onNewSolution(const SolutionBase& s) {
...
+	// otherwise there might still be a feasible solution
+	if(&child == &*children().back())
+		ContainerBasePrivate::onNewFailure(child, from, to);
+}
+
+void FallbacksPrivate::computeGenerate() {
+	if(solutions_.empty())
+		// move to first generator that can run
+		while(current_generator_ != children().end() && !(*current_generator_)->pimpl()->canCompute())
+			++current_generator_;
```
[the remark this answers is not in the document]
v4hn: Yes, this Fallback container must not be used with `MonitoringGenerator`, the previous implementation had the same limitation.
I plan to have different logic (possible even a `MonitoringFallbacks` container) for this case, because you would expect all Monitors in there to monitor the same Stage and the `Fallbacks` container to pass on solutions individually as well.

I agree about better scheduling required, but I have to tackle one task after the other and `Fallbacks` was much higher up on my list right now and my family life reduces my effective work time even more than usual at the moment...

core/test/stage_mockups.h
rhaschke: Yes. That's the reason gtest and gmock use macros everywhere. I share your dislike of macros :wink:

core/include/moveit/task_constructor/storage.h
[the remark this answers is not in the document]
v4hn: https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/storage.cpp#L74-L75 :cat2:

core/include/moveit/task_constructor/storage.h
```diff
@@ -100,6 +100,8 @@ class InterfaceState
 
 		inline Status status() const { return std::get<0>(*this); }
 		inline bool enabled() const { return std::get<0>(*this) == ENABLED; }
+		inline bool failed() const { return std::get<0>(*this) == FAILED; }
+		inline bool pruned() const { return std::get<0>(*this) == DISABLED; }
```
rhaschke, line 104: Shouldn't we use consistent naming here, i.e. either pruned or disabled?

core/include/moveit/task_constructor/storage.h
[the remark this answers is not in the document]
rhaschke: Thanks. I'm surprised that changing the order back (4b4f8f0b6db7f554b3c8a95a311adf98714de36b) doesn't have any impact on any of our tests!
Shouldn't at least the pruning tests be affected?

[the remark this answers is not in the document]
v4hn: No, because I also [adapted the internal logic testing for `pruned()` instead of `!enabled()`](https://github.com/ros-planning/moveit_task_constructor/pull/287/commits/767d410203c60b59b4e2bce86be875ec7b0caf5b)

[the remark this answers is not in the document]
v4hn: Ah, sorry I guess I misunderstood. Catching the early return breaking requires a rather sophisticated test with multiple solution candidates partially getting disabled. I'm not surprised we do not have a test for that condition, though the pick-place demo could hit it - it's just hard to spot and not reliable in that context.
I encourage you to write more tests for pruning of various situations (and possibly put it in a separate test file...). :-)

core/src/container.cpp
```diff
@@ -806,46 +814,144 @@ void Alternatives::onNewSolution(const SolutionBase& s) {
...
+	if(current_generator_ == children().end())
 		return;
 
 	try {
-		active_child_->pimpl()->runCompute();
+		// run ALL possible computations (on new state)
+		// this is needed to decide whether it should be passed to the next child too
+		while ((*current_generator_)->pimpl()->canCompute()){
+			(*current_generator_)->pimpl()->runCompute();
+		}
```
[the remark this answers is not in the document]
v4hn: For the generator case it seems I simply forgot to take this loop out again. You are perfectly right in this case.
That aside we have to make a clear distinction between generators and monitoring generators anyway (see my other comment). For plain generators we are always in a pinch if one returns `canCompute()==false`, but will later on "magically" become available, so I would like to put that aside for now.

core/src/container.cpp
```diff
@@ -806,46 +814,144 @@ void Alternatives::onNewSolution(const SolutionBase& s) {
...
+
+	// we've got to set this somewhere once the interface flags are known.
+	// so we might as well do it here
+	if(expected == GENERATE)
+		current_generator_ = children().begin();
+}
+
+void FallbacksPrivate::onNewFailure(const Stage& child, const InterfaceState* from, const InterfaceState* to) {
+	// only react to failure if it's the last possible candidate failing
+	// otherwise there might still be a feasible solution
```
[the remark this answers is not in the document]
v4hn: >  However, do you really need a call to onNewFailure() here, when you already have the one in computeFromExternal()?

True, I removed it here and left the empty override.

> If you remove it here, you probably need to add a similar one in computeGenerate() as well.

`onNewFailure` ignores GENERATE stages, so I did not add it there.

core/src/container.cpp
v4hn: I did not bother to include the logic for this before because it is impossible with the current interfaces to get rid of the exhaustive search altogether. Once we pass an external state to a child, Fallbacks *has* to commit to run that child until it's exhausted. I just pushed [an implementation to run only a single job per compute](https://github.com/ros-planning/moveit_task_constructor/pull/287/commits/2bac879c1bcc6946fc8cc874c61e22597a75bd95) nonetheless, because `Task` depends on `compute()` to return periodically to publish introspection.

core/include/moveit/task_constructor/storage.h
v4hn: True, I propose to rename `DISABLED` to `PRUNED` for clarity. There is currently no other way to DISABLE a state.

v4hn:
Good to see CI succeed for the first time in a while ;)

rhaschke:
I pushed a cleanup for `computeGenerate()`.

core/src/container.cpp
```diff
@@ -790,60 +805,144 @@ bool Alternatives::canCompute() const {
...
+		ends().reset(new Interface([this](Interface::iterator external, bool updated) {
+		                this->onNewExternalState<Interface::BACKWARD>(external, updated);
+		             }));
+}
+
+void FallbacksPrivate::onNewFailure(const Stage& /*child*/, const InterfaceState* /*from*/, const InterfaceState* /*to*/) {
+	// This override is deliberately empty.
+	// The method prunes solution paths when a child failed to find a valid solution for it,
+	// but in Fallbacks the next child might still yield a successful solution
+	// Thus pruning must only occur once the last child is exhausted (inside computeFromExternal)
```
rhaschke, line 882: In general, I agree with you. We do need a mechanism for a stage to claim that it is _exhausted_. `onNewFailure()` doesn't perfectly fit that purpose, because it's called on _any_ failure:
https://github.com/ros-planning/moveit_task_constructor/blob/f51f6eb982f90f3c31d9adfe1fc6e041d467c5bb/core/src/stage.cpp#L140-L143

I think we should rename the function to `onExhausted` and only call it when the solutions are actually exhausted. For basic `Propagator` and `Connect` stages that's always the case if they fail. However, `Generator` stages and containers will need special handling.
Actually, for `Generator` stages _exhausted_ is closely related to `!canCompute()`. Not sure that assumption holds for generator-like containers as well.

I any case, we might need to distinguish between "currently exhausted" and "finally exhausted". For example, `MonitoringGenerators` might be currently not able to compute something, but later, if they received new input, they can. The same holds for connect-like containers, where we already have seen that we can re-enable previously pruned solution branches.

Given this new semantics of the function, the Fallbacks container can easily check whether `child` actually exhausted its solutions and thus propagate this info upwards if needed.

core/src/container.cpp
```diff
@@ -790,60 +805,144 @@ bool Alternatives::canCompute() const {
...
+	(*current_generator_)->pimpl()->runCompute();
+}
+
+template <typename Interface::Direction dir>
+void FallbacksPrivate::onNewExternalState(Interface::iterator external, bool updated) {
+	// TODO(v4hn): updated is not implemented
+	if(updated){
+		ROS_DEBUG_NAMED("Fallbacks", "updating external states is not supported in Fallbacks");
+		return;
+	}
```
rhaschke, line 896: I'm wondering, why you didn't use [`ordered<ExternalState>`](https://github.com/ros-planning/moveit_task_constructor/blob/0aff5d56dd45b91b7c6c7c63c9674eec84e49aed/core/include/moveit/task_constructor/cost_queue.h#L28-L129). This way, incoming states would be ordered automatically and of course you could forward updated costs from the external state to the internal ones.

core/src/container.cpp
```diff
@@ -790,60 +805,144 @@ bool Alternatives::canCompute() const {
...
+
+	pending_states_.push_back(ExternalState(external, dir, children().begin()));
+}
+
+void FallbacksPrivate::computeFromExternal(){
+	assert(!pending_states_.empty());
+
+	if(!current_stage_) {
```
rhaschke, line 904: I'm not sure why you need a `current_stage_`. The current child for each pending external state is stored in `pending_states_` anyway. IMO, you just need to take the ordered list of pending_states_, fetch the most promising first, and try to compute on its child stage. If that fails (and there wasn't a solution computed before), you forward to the next stage but keep this pending state.
Did I miss something? If so, let's have a phone chat.

rhaschke (commented):
I think the function `computeFromExternal()` only handles propagator stages. Connect-like stages require their own set of functions, because you actually want to consider each _pair_ of (compatible) solutions separately - much like you consider each incoming external state of a propagator child separately now.

core/src/container.cpp
v4hn: I just pushed a change to `ordered<...>`, as proposed. I didn't manage to implement the corresponding cost updates yet though. Additionally I just noticed Fallbacks should not push pruned states to children either.

My intuition for the cyclic queue was that I would expect early stages in a `Fallbacks` container to be cheaper to compute, so computing the cheaper stages for all states instead of continuing the best-first-search through all children might be more economical. Either way should work reasonably well though.

rhaschke:
I pushed a bunch of new tests pointing out that priority propagation from external to internal interfaces of containers [doesn't work yet](https://github.com/ros-planning/moveit_task_constructor/pull/287/commits/bfe9bee9f04c67133f0dbfb0bb50229a39ebe80a#diff-5b0930974b77eab73f48b74214a60ac0562cd2a110285279bb5ad1094b40a842R201).

rhaschke:
Sorry for the mess. It's too late...

rhaschke:
I couldn't put the topic aside yet and pushed some more commits. Although the current tests work, the fundamental problem remains that a (fallback) container cannot decide yet whether a child is exhausted on a specific external state (a single one for propagating, a pair for connecting interfaces).
However, I'm loosely convinced that we can achieve this rather easily: We need to change the semantics of `onNewFailure()` to `onExhausted()` such that a child can explicitly state that it became exhausted on a specific input state (or pair).
If we can trust children to report that correctly, this info can be escalated to the parents. This might make `canCompute()` superfluous.

If you want to keep (for now) the current implementation fixing the _active_ pending state, I have this implemented as an alternative as well in [my repo](https://github.com/ubi-agni/moveit_task_constructor/tree/mtc-fallback-reworked): https://github.com/v4hn/moveit_task_constructor/compare/mtc-fallback-reworked...ubi-agni:mtc-fallback-reworked
However, this makes the new tests fail...

v4hn:
> I couldn't put the topic aside yet

Weren't you the one who said we should *coordinate* our efforts? :anger: 
I wrote 230 lines of patches for this branch yesterday afternoon that were not ready for pushing yet and conflict now.

I have a feverish kid sitting on my lap, so I don't the nerves right now to either fixup&polish my local patches or review your commits from the last 17 hours. Let's see whether I can keep my eyes open for long enough this evening to look into it.

rhaschke:
Sorry to hear that. If it helps you, just force-push your changes. I'm curious what they address. But, as I said, I need to turn to other stuff for next week...

rhaschke:
I think we should split this PR.
- There is whole bunch of commits adding tests. I think we agree on those and should commit them, [disabling](https://github.com/google/googletest/blob/master/docs/advanced.md#temporarily-disabling-tests) them if needed.
- There are a few commits working on pruning / un-pruning (1292d3ae3206c6074d1e141457113d43cce6119d, 1292d3ae3206c6074d1e141457113d43cce6119d, b14acfc66bce0cbd298b5b31d8f13e589cb8d4a1). These are the most controversial in my opinion. Would be good if you could provide failing tests for them or a sketch as in #221 for illustration.
- commits actually improving the `Fallbacks` container
- generic improvements (b14acfc66bce0cbd298b5b31d8f13e589cb8d4a1, 8b49d861de8eafc2214e08893709c257867217de, ffd884d028164ff123f3b6d251ee0fdf9584450d)

v4hn:
I finally finished splitting&rebasing the branch.
I included most of your commits in one of the three PRs.

I explicitly excluded https://github.com/v4hn/moveit_task_constructor/commit/ece9cd018db55b78b6cd0bf38ad3b88def691ebf because I changed the logic quite a lot in my own commit and I believe your patch can result in pushing multiple states to the same child and thus mess with pruning because there is nothing connecting the entry in pending_ to the corresponding call to `runCompute`. It's quite entangled and I did not attempt to write a test.

v4hn:
@rhaschke if you should have time after summer school, iros and DGR days, this should be ready for merge (once more).

rhaschke (commented):
It took me quite a while to understand your refactorings as you provided only very sparse documentation, both in commit comments and in code :disappointed:
The logic of switching between start and end queue is probably motivated by this commit comment of https://github.com/ros-planning/moveit_task_constructor/pull/287/commits/23f32d72d492048bc89d0d68e520d66f47ba1936:
> with a single ordered queue, jobs from one direction were always preferred

However, I don't see this imbalance (yet). I suggest having a chat these days to clarify this.
Considering a propagator stage, I would expect that a single queue will order states from both ends by their priority, which is exactly what I would expect. I think we dropped the stage, which could propagate from either side anyway, didn't we?

For CONNECT-like interfaces, I think we need a different approach, namely a job queue of `(start, end)` pairs, much like in the `Connecting` stage.

Hence we will have 3 implementations: for generator, propagator, and connect stages. 
To simplify the distinction between those, I suggest having 3 `FallbacksPrivate` classes deriving from a common base class. When initializing the interface in `initializeExternalInterfaces()` (which would be implemented by the base class), the actually required class type could be instantiated. This separation would nicely separate data structures required for these variants and avoid the interface-type-based switches. What do you think?

Most review comments I directly implemented in some additional commits. Please have a look. I'm going to merge with #309 as well and check a disabled unit test there...

rhaschke:
Superseded by #311.

PR #292, opened 2021-09-15
Update main scene node's pose on updates of robot's model frame

### LINKED ISSUE #291: Motion Planning tasks introspection is wrong when using map as fixed frame
Hi,

There seems to be a problem when trying to review the planned task stages on robot that is localized into a map. 

problem_mtc_map

As you can see on the above picture, I want to review the "grasp pose IK" stage, when  I click on it, another robot with it's environment appears at the origin of the map (in the foreground of the above picture). 
The actual robot position is at the background of the above picture.

It seems also to affects the view of the poses generated by moveit task constructor. 

The expected behavior is that the stage review should show the robot at it's actual localized position

Here is what happen in a video, when I want to review the solution found, another robot appear at the origin of the map instead of the actual position of the robot : 

https://user-images.githubusercontent.com/59170639/133257627-2aeeab73-32ec-45bd-9857-499ff894a21b.mp4

Feel free to ask if something is unclear.

rhaschke:
That's the same fix as https://github.com/ros-planning/moveit/pull/2876.
This fixes #291 in a naive way: 
If the map changes over time, I would actually expect the solution display to use the robot frame transform at the time of planning - not the most recent one as implemented here. However, this timestamp isn't conveyed with our messages yet.
@v4hn @YannickRiou any opinions?

### DIFF
--- a/visualization/motion_planning_tasks/src/task_display.cpp
+++ b/visualization/motion_planning_tasks/src/task_display.cpp
@@ -179,6 +179,7 @@ void TaskDisplay::calculateOffsetPosition() {
 void TaskDisplay::update(float wall_dt, float ros_dt) {
 	requestPanel();
 	Display::update(wall_dt, ros_dt);
+	calculateOffsetPosition();
 	trajectory_visual_->update(wall_dt, ros_dt);
 }
 


v4hn:
> However, this timestamp isn't conveyed with our messages yet.

If your model frame is not the map but the robot base link, MoveIt has no way of knowing the transform, true.
But there could be a very long delay between acquisition of the PlanningScene and visualization of the planner result,
so I would not want to rely on TF to provide the transforms based on a time stamp.
The PlanningScene can have a fixed transform to the map frame and should already have it when you add a mobile virtual frame (using the map frame as planning frame).
If it doesn't include the transform in the case where the robot base frame is used as planning frame, we could possibly add it explicitly? Adding the whole known TF tree seems somewhat exaggerated, though it would solve the problem. Alternatively we could have an explicit option in the CurrentState/FixedState stages to add a target frame transform.

PR #293, opened 2021-09-16
Improve readability of internal-external bimap using tags

rhaschke:
I had a hard time reading through this code. I hope, using these tags, it's slightly better now.

### DIFF
--- a/core/include/moveit/task_constructor/container_p.h
+++ b/core/include/moveit/task_constructor/container_p.h
@@ -121,9 +121,14 @@ public:
 	InterfacePtr pendingBackward() const { return pending_backward_; }
 	InterfacePtr pendingForward() const { return pending_forward_; }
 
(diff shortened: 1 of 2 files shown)


v4hn (approved):
I don't see any change in readability really, but if it helps you that's enough here.

The other cleanup also makes sense.

rhaschke:
> I don't see any change in readability really, but if it helps you that's enough here.

It's more explicitly stated whether we refer to an internal or external state (instead of using first and second).

v4hn:
> It's more explicitly stated whether we refer to an internal or external state (instead of using first and second).

I read the patch before merging. :cat:
It's a reasonable feature I did not know about before. I just wanted to state that `first` and `second` were clear enough for me when you get the pair from `**internal**To**External**()` or vice versa. Using a new feature on the other hand requires knowing about this additional feature.

PR #296, opened 2021-09-23
setIKFrame() supports subframes

gautz:
Currently only Robot links can be set as IKFrame for planning with a `moveTo` stage. 
This PR allows to use the function `setIKFrame()` with a subframe of an attached collision object. 
This allows to support [this idea](https://github.com/ros-planning/moveit/issues/2866)

TODO: 
- [ ] Be able to set the IKframe with an offset from the subframe.
- [ ] Discuss if having something similar to `move_group.setEndEffectorLink()` would be possible. Setting path (or goal) constraints with this approach based on transforming the target pose is still difficult.

### DIFF
--- a/core/src/stages/move_to.cpp
+++ b/core/src/stages/move_to.cpp
@@ -220,6 +220,21 @@ bool MoveTo::compute(const InterfaceState& state, planning_scene::PlanningSceneP
 			ik_pose_msg.pose.orientation.w = 1.0;
 		} else {
 			ik_pose_msg = boost::any_cast<geometry_msgs::PoseStamped>(value);
+			if (scene->knowsFrameTransform(ik_pose_msg.header.frame_id)) {
+				// get IK frame in planning frame
+				Eigen::Isometry3d ik_frame_eigen_world = scene->getFrameTransform(ik_pose_msg.header.frame_id);
+				// determine IK link from group
+				if (!(link = jmg->getOnlyOneEndEffectorTip())) {
+					solution.markAsFailure("missing ik_frame");
+					return false;
+				}
+				ik_pose_msg.header.frame_id = link->getName();
+				// get planning group tip frame in planning frame
+				Eigen::Isometry3d link_pose_eigen = scene->getCurrentState().getGlobalLinkTransform(link);
+				// get IK frame in planning group tip frame
+				Eigen::Isometry3d ik_pose_eigen = link_pose_eigen.inverse() * ik_frame_eigen_world;
+				tf::poseEigenToMsg(ik_pose_eigen, ik_pose_msg.pose);
+			}
 			if (!(link = robot_model->getLinkModel(ik_pose_msg.header.frame_id))) {
 				solution.markAsFailure("unknown link for ik_frame: " + ik_pose_msg.header.frame_id);
 				return false;


core/src/stages/move_to.cpp
```diff
@@ -220,6 +220,24 @@ bool MoveTo::compute(const InterfaceState& state, planning_scene::PlanningSceneP
...
+					return false;
+				}
+				ik_pose_msg.header.frame_id = link->getName();
+				// get planning group tip frame in planning frame
+				Eigen::Isometry3d link_pose_eigen = scene->getCurrentState().getGlobalLinkTransform(link);
+				// get IK pose in planning group tip frame
+				ik_pose_eigen = link_pose_eigen.inverse() * ik_frame_eigen_world * ik_pose_eigen;
+				tf::poseEigenToMsg(ik_pose_eigen, ik_pose_msg.pose);
+			}
 			if (!(link = robot_model->getLinkModel(ik_pose_msg.header.frame_id))) {
```
JafarAbdi, line 241: Wouldn't it always fail here in the case of a subframe .?

core/src/stages/move_to.cpp
```diff
@@ -220,6 +220,24 @@ bool MoveTo::compute(const InterfaceState& state, planning_scene::PlanningSceneP
...
 			ik_pose_msg = boost::any_cast<geometry_msgs::PoseStamped>(value);
+			if (scene->knowsFrameTransform(ik_pose_msg.header.frame_id)) {
+				// get IK frame in planning frame
+				Eigen::Isometry3d ik_frame_eigen_world = scene->getFrameTransform(ik_pose_msg.header.frame_id);
+				// get IK pose in IK frame
+				Eigen::Isometry3d ik_pose_eigen;
+				tf::poseMsgToEigen(ik_pose_msg.pose, ik_pose_eigen);
+				// determine IK link from group
+				if (!(link = jmg->getOnlyOneEndEffectorTip())) {
+					solution.markAsFailure("missing ik_frame");
```
JafarAbdi, line 231: The ik_frame isn't missing here, so the failure message should be updated

gautz: Then I guess also line 216 should be updated, e.g. with "No unique end effector tip"

core/src/stages/move_to.cpp
```diff
@@ -213,14 +213,31 @@ bool MoveTo::compute(const InterfaceState& state, planning_scene::PlanningSceneP
 		if (value.empty()) {  // property undefined
 			// determine IK link from group
 			if (!(link = jmg->getOnlyOneEndEffectorTip())) {
-				solution.markAsFailure("missing ik_frame");
+				solution.markAsFailure("No unique end effector tip");
```
JafarAbdi, line 216: I think the previous error code is correct and shouldn't be removed, otherwise, it may confusing if a user forgot to add the ik_frame

gautz: The user may not want to set `ik_frame` and just use the IK link from group, no failure or error message in this case. The error occurs when the the user did not set `ik_frame` AND no or several end effector tips are defined in the JMG. So I still think this new failure message is clearer to the end.

Or did I misunderstand something?

JafarAbdi: Yes, you're right, but the requirement for having a unique end-effector tip is a fallback for not specifying an `ik_frame` so the solution for the user is to add an `ik_frame` which is what the previous failure message is saying

gautz:
@rhaschke @v4hn any thought about this PR?

v4hn:
I'm writing [tests for `MoveTo` right now](https://github.com/v4hn/moveit_task_constructor/tree/pr-move-to-tests) and will probably include a version of your patch there in a bit. Sorry to leave you hanging here with no reply :pleading_face:

PR #304, opened 2021-10-24
Add MoveTo tests & make them pass

v4hn:
- implement a simple IK interface for `JointInterpolationSolver` Cartesian goals
- fix some bugs
- include attached objects/subframe fixes similar to #296

### DIFF
--- a/core/src/stages/move_to.cpp
+++ b/core/src/stages/move_to.cpp
@@ -43,6 +43,7 @@
 #include <moveit/task_constructor/stages/move_to.h>
 #include <moveit/task_constructor/cost_terms.h>
 #include <moveit/task_constructor/moveit_compat.h>
+#include <moveit/task_constructor/utils.h>
 
 #include <rviz_marker_tools/marker_creation.h>
 
@@ -262,13 +263,8 @@ bool MoveTo::compute(const InterfaceState& state, planning_scene::PlanningSceneP
 		add_frame(target, "target frame");
 		add_frame(ik_pose_world, "ik frame");
 
-		const moveit::core::LinkModel* parent {
-#if MOVEIT_HAS_STATE_RIGID_PARENT_LINK
-			scene->getCurrentState().getRigidlyConnectedParentLinkModel(ik_pose_msg.header.frame_id)
-#else
-			getRigidlyConnectedParentLinkModel(scene->getCurrentState(), ik_pose_msg.header.frame_id)
-#endif
-		};
+		const moveit::core::LinkModel* parent{ utils::getRigidlyConnectedParentLinkModel(scene->getCurrentState(),
+			                                                                              ik_pose_msg.header.frame_id) };
 
 		// transform target pose such that ik frame will reach there if link does
 		Eigen::Isometry3d ik_pose;
--- a/core/src/stages/move_to.cpp
+++ b/core/src/stages/move_to.cpp
@@ -208,44 +208,10 @@ bool MoveTo::compute(const InterfaceState& state, planning_scene::PlanningSceneP
 		// What frame+offset in the robot should go there?
 		geometry_msgs::PoseStamped ik_pose_msg;
 
-		const boost::any& value = props.get("ik_frame");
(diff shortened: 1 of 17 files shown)


rhaschke (approved):
Looks good. Sorry for the long delay in reviewing this. I focussed on the MoveIt release first.
Note the minor fixup commits. They should be squashed into their base commit before merging (I guess you want to keep history).

rhaschke:
Sorry, I didn't test on Melodic... Please, just drop my latest commit (8408fba17d75efbf7cd5716ba51899ff96cd612b).

rhaschke:
Hm. https://github.com/ros-planning/moveit_resources/pull/92 broke the pick+place test. Should be fixed now.

core/src/stages/move_to.cpp
```diff
@@ -204,47 +202,81 @@ bool MoveTo::compute(const InterfaceState& state, planning_scene::PlanningSceneP
...
+			if (ik_pose_msg.header.frame_id.empty() && !get_tip(ik_pose_msg.header.frame_id)) {
+				solution.markAsFailure("frame_id of ik_frame is empty and no unique group tip was found");
+				return false;
+			} else if (!scene->knowsFrameTransform(ik_pose_msg.header.frame_id)) {
+				std::stringstream ss;
+				ss << "ik_frame specified in unknown frame '" << ik_pose_msg << "'";
+				solution.markAsFailure(ss.str());
 				return false;
 			}
 		}
```
rhaschke, line 240: This block of code is required 1:1 for `MoveRelative` as well:
https://github.com/ros-planning/moveit_task_constructor/blob/7ec874572c3b35f1cdd7a71c5e6b94c749b3093b/core/src/stages/move_relative.cpp#L199-L213

We should factor that out into a function...

core/test/test_move_to.cpp
```diff
@@ -70,29 +72,87 @@ TEST_F(PandaMoveTo, stateTarget) {
...
+	aco.object.header.frame_id = aco.link_name;
+	aco.object.operation = aco.object.ADD;
+	aco.object.id = id;
+	aco.object.primitives.resize(1, []() {
+		shape_msgs::SolidPrimitive p;
+		p.type = p.SPHERE;
+		p.dimensions.resize(1);
+		p.dimensions[p.SPHERE_RADIUS] = 0.01;
+		return p;
+	}());
```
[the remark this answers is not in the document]
v4hn: ROS messages do not have useful constructors, so it's pretty much a replacement for that.
The main advantage is that scope does not leak. In this case the variable `p` does not stay available outside the argument. I use it as a more concise version of extra braces to restrict scope `{ ... }`.

The real advantage of [IIFE](https://rigtorp.se/iife/) (I'm not sure I used that anywhere) is that you can initialize const objects that way too.

> I guess they are inlined automatically by the compiler?

[Yes](https://godbolt.org/z/Krjvrqvnj).

core/src/stages/move_to.cpp
v4hn: please have another look. I'm happy with the code again (assuming CI is) and I addressed this too.

core/test/test_move_to.cpp
```diff
@@ -62,7 +62,7 @@ TEST_F(PandaMoveTo, mapTarget) {
 }
 
 TEST_F(PandaMoveTo, stateTarget) {
-	move_to->setGoal([]() {
+	move_to->setGoal([] {
```
rhaschke: Interesting. Haven't seen this syntax before :wink:

rhaschke (approved):
Thanks for the cleanup. LGTM. Note, that I didn't test myself - I assume you did that :wink:

core/test/test_move_to.cpp
v4hn: Me neither, but it's nice. The downside is that for some reason `[] -> Type {}` is not supported syntax though and requires `[]() -> Type {}` (which reads so much... better?)

PR #306, opened 2021-11-08
Improve failure reporting

rhaschke:
I had to debug a `MoveRelative` stage that was failing for no visible reason. We should improve failure reporting by:
- showing the failing pose (this is currently pruned away in Cartesian and joint interpolation planners)
- naming and highlighting collision pairs if a collision is the reason for a failure
- naming and highlighting joint limit violations

This PR is just a (tiny) starting point into that direction. It (partially) served me for debugging my problem. However, this is not generally applicable yet: The `achieved_fraction` returned by the Cartesian interpolation methods is not the fraction we actually need to evaluate for the stage... I'm tempted to drop the `min_fraction` parameter from the planners as this should be a stage property only.


rhaschke:
> I'm tempted to drop the `min_fraction` parameter from the planners as this should be a stage property only.

@v4hn, any comments on this?

PR #309, opened 2021-11-19
Fix Pruning

### LINKED ISSUE #308: Bug in pruning
I have a very simple task, that fails to find a solution because it's not trying to connect certain pairs, I created a [simple test](https://github.com/JafarAbdi/moveit_task_constructor/tree/pr-pruning_bug) the trigger it, I have spent so much time trying to debug it without any hope :(

```
Current State
Connect
ComputeIK
MoveRelative
```

I'll keep posting my findings, I would really appreciate any help

rhaschke:
Building on some new (failing) tests introduced by @JafarAbdi and @v4hn, this PR aims to improve (and eventually fix) pruning.
This follows these general ideas:

- If a planning stage fails in a linear pipeline, we can skip all remaining planning stages as well (cols of _left-most_ block)
- However, we need to keep the states in the interfaces as we might re-enable a path later on (left-most cols of _middle_ block)
- A new CONNECT state becomes disabled too (together with its connected branch) (right col of _middle_ block)
- We shouldn‘t prune if there are alternative solutions (_right_ block)

image

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,7 +32,7 @@ jobs:
             DOCKER_RUN_OPTS: >-
               -e PRELOAD=libasan.so.5
               -e LSAN_OPTIONS="suppressions=$PWD/.github/workflows/lsan.suppressions"
-            TARGET_CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -O1"
+            TARGET_CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -O1 -g"
 
     env:
       CATKIN_LINT: true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@
 repos:
   # Standard hooks
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v4.0.1
     hooks:
       - id: check-added-large-files
       - id: check-case-conflict
@@ -29,7 +29,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 21.11b1
     hooks:
       - id: black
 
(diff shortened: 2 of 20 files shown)


rhaschke:
Finally fixed #308. The culprit was that `setStatus()` was propagating the `FAILED` status unconditionally through existing `CONNECT` solutions. However, in this case `hasPendingOpposites()` needs to be used to stop this propagation if other solution pairs are still feasible.

rhaschke:
## Reenabling pruned solution branches
We need to distinguish the _origin_ of a pruned branch:
- If the origin is _not_ a `Connecting` stage, a branch can only become re-enabled when a new solution appears within this branch.
  Newly appearing `CONNECT` states should get disabled as well. For this mode, I suggest the `Status` name: `PRUNED`.
- If the origin is a `Connecting` stage, we need to distinguish the two different situations already mentioned in #221, namely:
  - A newly appearing `CONNECT` state partnering an _initially failed_ `CONNECT` state. In this case, the branch can be re-enabled immediately (_right-most_ block). Name suggestion for `Status`: `ARMED` (for re-enabling).
  - If the new `CONNECT` state appears in another `Connecting` stage, the pruned solution branch must not get re-enabled.
    Rather the new state (and its connected branch) should become pruned themselves. (left col of _middle_ block). 
    Only if a new state partners the _initially failed_ state, the branch will become re-enabled (right col of _middle_ block).

Summarizing, I suggest the following `Status` enums:
- `ENABLED`
- `ARMED`: states of a failed `Connecting` stage that are eligible for re-enabling the branch
- `PRUNED`

image

rhaschke:
## TODOs

- [ ] Rename Interface::Status `FAILED` -> `ARMED`
- [ ] _Recursively_ re-enable states when matching an `ARMED` state
- [ ] Recursively prune new `Connecting` state if there is no enabled opposite
- [ ] Propagate status updates across `Connecting` stage

rhaschke:
I merged #311 here, resolving all conflicts. This finalizes this PR (and #311), proving that they nicely play together.
@v4hn, ideally merge by fast-forwarding to [`ubi-agni@master`](https://github.com/ubi-agni/moveit_task_constructor/tree/master) as I already continued development from there.

PR #310, opened 2021-11-23
Stage::reset() should reset total_compute_time_

rhaschke:
@JafarAbdi, I think you introduced `total_compute_time_`. Did you omit to reset `total_compute_time_` on purpose or was it just missed? I would expect that the total compute time count restarts after a `reset()`.


JafarAbdi (approved):
> Did you omit to reset total_compute_time_ on purpose or was it just missed? I would expect that the total compute time count restarts after a reset().

I missed it, thank you for the fix

PR #311, opened 2021-11-25
Fallbacks reworked: Factorize FallbacksPrivate into 3 separate classes

rhaschke:
This is a new attempt to correctly implement `Fallbacks`. The key insight is that we need to implement different behavior based on the actual interface type (`GENERATE`, `PROPAGATE_*`, `CONNECT`). These types differ on what they consider a `job` that a child stage should run on till exhaustion before advancing to the next child.
- `GENERATE`: A job is essentially identical to the child itself. The generator child runs until it is exhausted. This is the only operation mode I had in mind when originally implementing the `Fallbacks` stage.
- `PROPAGATE_*`: A job is an external input state. As [pointed out by @v4hn](https://github.com/ros-planning/moveit_task_constructor/pull/280#discussion_r676565782) for each new input state, the Fallbacks container should start over from its first child stage. This behavior was added in #287 (but also mistakenly applied for `CONNECT` interfaces).
- `CONNECT`: A job is a single pair of (start, end) states. As we cannot withdraw a state once it was forwarded to a child, we need to disable (and later re-enable) states in the child `starts`/`ends` interfaces to ensure that exactly one specific (start, end) pair is active at a time.

To correctly implement these three behaviors, this PR essentially realizes three different implementations for `FallbacksPrivate`.
Common behavior between these classes is implemented in `FallbacksPrivate`. The actual interface-specific class is instantiated in `initializeExternalInterfaces()` resp. `Fallbacks::replaceImpl()` when the actual interface is known.

### TODOs:
- [x] Simplify implementation based on some more ideas from #287 
- [ ] Implement `FallbacksPrivateConnect` (**WIP**)
  - [x] Basic version available, but with similar bug as before: `CONNECT` pairs cannot easily be treated correctly as separate jobs, causing the issue described in https://github.com/ros-planning/moveit_task_constructor/pull/311#issuecomment-979766156. 
    My plan to solve that issue was to disable inactive pairs (or their solutions).
    However, I now have a coarse idea that Michael's approach, feeding start and end interfaces in an alternating fashion, might work as well. We should definitely talk, Michael :wink:
     EDIT: In the [given example](https://github.com/ros-planning/moveit_task_constructor/pull/311#issuecomment-979766156), the pair `3-4` that failed for the first connecting child, will not be passed on to the next child.
  - [ ] Merge #309 and fix remaining pruning issues with containers
    - [ ] If a single _pair_ of external states receives multiple solutions, only the best should be kept and all worse (partial) solutions should be pruned. Need to sketch a unit test for this...

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,7 +32,7 @@ jobs:
             DOCKER_RUN_OPTS: >-
               -e PRELOAD=libasan.so.5
               -e LSAN_OPTIONS="suppressions=$PWD/.github/workflows/lsan.suppressions"
-            TARGET_CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -O1"
+            TARGET_CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -O1 -g"
 
     env:
       CATKIN_LINT: true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@
 repos:
   # Standard hooks
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v4.0.1
     hooks:
       - id: check-added-large-files
       - id: check-case-conflict
@@ -29,7 +29,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 21.11b1
     hooks:
       - id: black
 
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -159,6 +159,8 @@ class FallbacksPrivate;
  */
 class Fallbacks : public ParallelContainerBase
 {
+	inline void replaceImpl();
+
 public:
 	PRIVATE_CLASS(Fallbacks);
 	Fallbacks(const std::string& name = "fallbacks");
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -167,12 +167,15 @@ public:
 
 	void reset() override;
 	void init(const moveit::core::RobotModelConstPtr& robot_model) override;
-	bool canCompute() const override;
-	void compute() override;
 
 protected:
 	Fallbacks(FallbacksPrivate* impl);
 	void onNewSolution(const SolutionBase& s) override;
+
+private:
+	// not needed, we directly use corresponding virtual methods of FallbacksPrivate
+	bool canCompute() const final { return false; }
+	void compute() final {}
 };
 
 class MergerPrivate;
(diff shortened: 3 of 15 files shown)


rhaschke:
Thinking more about CONNECT-like stages, I don't think that Michael's approach (#287) is feasible. Let's consider the following simple situation, where generators produce solutions in the given order.
```
GEN           1 3
Fallbacks     |X
GEN           2 4
```
When passing state 4 to the Fallbacks' connector, it forms pending pairs with both 1 and 3. Thus, the last pair is `3-4` is computed without the fallback container being aware of it. Actually, the container cannot (immediately) know whether `1-4` or `3-4` failed or succeeded.

rhaschke:
### Solution proposal for https://github.com/ros-planning/moveit_task_constructor/pull/311#issuecomment-979766156
The main idea is to give up on tracking the current job. Just monitor solutions and failures reported by the current stage.
Assuming that `onNewFailure` reports _exhaustive_ failures for the given state pair (which is not yet the case), we can do this:
- Maintain _separate_ `pending`, `succeeded`, and `failed` lists for each child. (Each list comprises state pairs.)
  Jobs received externally, enter the _first_ child's `pending` list.
- Jobs entering a `pending` list are _immediately_ propagated to the owning child as well.
- Succeeded jobs (noticed via `onNewSolution`) are moved from `pending` to `succeeded` indicating a success.
- Failed jobs (noticed via `onNewFailure`) are moved to `failed`. 
- The `failed` list of a child is just a pointer to the `pending` list of the next child. Thus, moving a job to `failed`, makes it `pending` for the next child, which thus becomes computable.
- If there is no next child, the job is moved a specific `failed` list of the container, indicating ultimate failure. The state pair can be reported as an _exhaustive_ failure in this case (employing the list's `notify()` callback).
- Children can run in parallel if they have a non-empty `pending` list.
  Scheduling could be proportional to the number of pending jobs per child.

As `onNewFailure` doesn't (yet) report an _exhaustive_ failure for jobs, we need a slightly different approach meanwhile:
- There can be only a single active child. Only if that child cannot compute anymore, we move on to the next child, processing its `pending` queue. This approach clearly prefers early children more than later ones!
  **Open issue**: New external solutions can still arrive _after_ the child was marked exhausted, can't they?
  If so, the first child should become _active_ again.
  So, we definitely need to switch to the above scheme to ultimately solve the Fallbacks issue.
- Succeeded jobs need to be moved from `failed` to `succeeded`, thus removing them from the next child's `pending` list.
  This situation cannot occur when failure means exhaustive failure. Thus, the above scheme doesn't actually need `succeeded`.

rhaschke:
In a phone call with Michael, we agreed to limit the `CONNECT`-like interface of the Fallbacks container to pure `Connecting` children. This will allow using the `pending` queues of the `Connecting` stages directly and calling their `compute(from, to)` method instead of the generic `compute()`.

rhaschke:
I eventually found some time to implement https://github.com/ros-planning/moveit_task_constructor/pull/311#issuecomment-984540252. It resolves https://github.com/ros-planning/moveit_task_constructor/pull/311#issuecomment-979766156 as expected.
Please don't merge yet as this will create conflicts with #309. I'm working on a merge of this PR into #309.

PR #318, opened 2021-12-12
MoveRelative: Fix bug in calculation of target pose

JafarAbdi:
https://github.com/ros-planning/moveit_task_constructor/commit/3b835986e34e09edf1e2f93680f0b6a5d4ace38b# introduced a bug where `MoveRelative` tries to move the link's origin to `ik_frame` despite `link_T_ik_frame` isn't the identity matrix, the problem is in the calculation we assume `target_eigen` to be `world_T_ik_frame_new_pose` which isn't the case (see [move_relative.cpp#L270-L273](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/move_relative.cpp#L270-L273) and [move_relative.cpp#L243-L249](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/move_relative.cpp#L243-L249))

I think the bug existed even before the refactor, it's just that we didn't have a case where `link_T_ik_frame` isn't the identity matrix

### DIFF
--- a/core/src/stages/move_relative.cpp
+++ b/core/src/stages/move_relative.cpp
@@ -204,6 +204,7 @@ bool MoveRelative::compute(const InterfaceState& state, planning_scene::Planning
 		Eigen::Vector3d angular;  // angular rotation
 		double linear_norm = 0.0, angular_norm = 0.0;
 
+		// Target pose such that ik frame will reach there if link does
 		Eigen::Isometry3d target_eigen;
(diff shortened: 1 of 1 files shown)


v4hn:
This patch is still broken because the direction property will be computed based on `link_pose` instead of `ik_pose_world`. I will provide a proper fix + test in a bit.

PR #319, opened 2021-12-13
Return error code from task::plan

JafarAbdi:
Sometimes it's really useful to know whether the plan failed due to timeouts, preempting, or task can no longer run compute

Another solution would be to have something similar to [MoveItErrorCode](https://github.com/ros-planning/moveit/blob/928afabe6cd857e15ab9e5b1c35f621c3f586ae6/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h#L61-L90)


core/test/test_container.cpp
```diff
@@ -611,7 +611,7 @@ TEST(Task, reuse) {
 
 	try {
 		configure(t);
-		EXPECT_TRUE(t.plan(1));
+		EXPECT_TRUE(t.plan(1).val == moveit_msgs::MoveItErrorCodes::SUCCESS);
```
v4hn, line 614: I would strongly prefer an error code wrapper class to keep the current boolean check for success.

Ideally we would migrate the `MoveItErrorCode` class you refer to to `moveit_core` (where it should have been to start with...). Alternatively we could copy the class here.

core/include/moveit/task_constructor/task.h
```diff
@@ -47,6 +47,26 @@
...
+public:
+	MoveItErrorCode() { val = 0; }
+	MoveItErrorCode(int code) { val = code; }
+	MoveItErrorCode(const moveit_msgs::MoveItErrorCodes& code) { val = code.val; }
+	explicit operator bool() const { return val == moveit_msgs::MoveItErrorCodes::SUCCESS; }
+	bool operator==(const int c) const { return val == c; }
+	bool operator!=(const int c) const { return val != c; }
+};
+}  // namespace core
+}  // namespace moveit
```
rhaschke, line 68: I'm hesitant to include a redefinition of MoveItErrorCode. I would prefer to backport: https://github.com/ros-planning/moveit/pull/3015
If that's not desired, just reuse an existing definition via:
`using MoveItErrorCode = moveit::planning_interface::MoveItErrorCode;`

core/test/test_container.cpp
```diff
@@ -649,23 +649,23 @@ TEST(Task, timeout) {
...
 	// no timeout, but limited number of solutions
-	EXPECT_TRUE(t.plan(3).val == moveit_msgs::MoveItErrorCodes::SUCCESS);
+	EXPECT_TRUE(t.plan(3));
 	EXPECT_EQ(t.solutions().size(), 3u);
 
 	// zero timeout fails
 	t.reset();
 	t.setTimeout(0.0);
-	EXPECT_TRUE(t.plan().val == moveit_msgs::MoveItErrorCodes::TIMED_OUT);
+	EXPECT_FALSE(t.plan());
```
rhaschke, line 658: Here, the original check was more specific! Please revert.

rhaschke (commented):
~~As far as I can see, this just adapts the code to https://github.com/ros-planning/moveit/pull/3009.
Your original comment suggests that there is some new functionality introduced. Did I miss something? Could you please clarify.~~
Sorry, I reviewed a single commit only.

core/include/moveit/task_constructor/task.h
```diff
@@ -117,11 +137,11 @@ class Task : protected WrapperBase
 	void init();
 
 	/// reset, init scene (if not yet done), and init all stages, then start planning
-	bool plan(size_t max_solutions = 0);
+	moveit::core::MoveItErrorCode plan(size_t error_code = 0);
```
rhaschke, line 140: The argument name `max_solutions` should stay:
```suggestion
	moveit::core::MoveItErrorCode plan(size_t max_solutions = 0);
```

demo/src/pick_place_task.cpp
```diff
@@ -490,7 +490,7 @@ bool PickPlaceTask::plan() {
 	ROS_INFO_NAMED(LOGNAME, "Start searching for task solutions");
 	int max_solutions = pnh_.param<int>("max_solutions", 10);
 
-	return task_->plan(max_solutions);
+	return task_->plan(max_solutions) == moveit_msgs::MoveItErrorCodes::SUCCESS;
```
rhaschke, line 493: Can't you stick with the old code here, because the result is cast to bool automatically?

rhaschke (commented):
Looks like I review a single commit only. Approving with the suggested changes.

demo/src/pick_place_task.cpp
JafarAbdi: For some reason I'm getting `No viable conversion from returned value of type 'moveit::core::MoveItErrorCode' to function return type 'bool' explicit conversion function is not a candidate` with the old version

core/test/test_container.cpp
rhaschke: This isn't addressed yet, is it?

rhaschke: Hm. It's not reverted in https://github.com/ros-planning/moveit_task_constructor/pull/319/commits/92a6267b80f1eab68987e4ca028867278b03a7aa, but it is actually fixed.

core/include/moveit/task_constructor/moveit_compat.h
```diff
@@ -57,3 +57,6 @@
 #define MOVEIT_HAS_OBJECT_POSE MOVEIT_VERSION_GE(1, 1, 6)
 
 #define MOVEIT_HAS_STATE_RIGID_PARENT_LINK MOVEIT_VERSION_GE(1, 1, 6)
+
+// moveit_core has MoveItErrorCode class
+#define MOVEIT_CORE_HAS_ERROR_CODE MOVEIT_VERSION_GE(1, 1, 7)
```
rhaschke, line 62: I have released 1.1.7 for Noetic and we have back-ported #3009 to Melodic as well (and I plan to cut a Melodic release soon as well). Could we simply drop this?

demo/src/pick_place_task.cpp
rhaschke: Try with explicit conversion to bool then?
```suggestion
	return static_cast<bool>(task_->plan(max_solutions));
```

rhaschke (approved):

PR #321, opened 2021-12-21
execute_task_solution_capability: check for canceling request before canceling the goal handle

JafarAbdi:
Fix the following exception

```
[move_group-1] terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
[move_group-1]   what():  goal_handle attempted invalid transition from state EXECUTING with event CANCELED
```

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -119,7 +119,7 @@ void ExecuteTaskSolutionCapability::goalCallback(
 
 	if (result->error_code.val == moveit_msgs::msg::MoveItErrorCodes::SUCCESS)
 		goal_handle->succeed(result);
-	else if (result->error_code.val == moveit_msgs::msg::MoveItErrorCodes::PREEMPTED)
+	else if (result->error_code.val == moveit_msgs::msg::MoveItErrorCodes::PREEMPTED && goal_handle->is_canceling())
 		goal_handle->canceled(result);
 	else
 		goal_handle->abort(result);


stephanie-eng (approved):

JafarAbdi:
> If not yet canceling, shouldn't you explicitly cancel the goal?

Do you mean the `goal_handle->abort(result);` in line#125 ?

rhaschke:
> Do you mean the `goal_handle->abort(result);` in line#125 ?

No. According to the [design document](http://design.ros2.org/articles/actions.html), there should be a separate cancel_goal() method. However, I couldn't find a corresponding method. According to code comments, `canceled()` can be called in states `EXECUTING` and `PENDING`. However, the exception indicates that this is not the case. So, I am confused as well. Maybe, file an upstream issue? Are you sure, the execution can still be correctly canceled with this PR?

rhaschke:
Thanks for this clarification. I guess the Cancel request is accepted here then:
https://github.com/ros-planning/moveit_task_constructor/blob/3a30e498bf081c83e6d206b9c150001cfbd67ad7/capabilities/src/execute_task_solution_capability.cpp#L132

JafarAbdi:
Super sorry I forgot to submit my comment :grimacing: 

> I guess the Cancel request is accepted here then:

Yes, you're right [preemptCallback](https://github.com/ros-planning/moveit_task_constructor/blob/ros2/capabilities/src/execute_task_solution_capability.cpp#L128-L133) is used for acceptance

I should have added more info about what this fixes, as @v4hn said this only works when the `PREEMPT` isn't triggered by the cancel API

I just noticed that canceling the goal from the client will not work because for some reason ROS2 expects us to manage the thread to run the goal callback ourselves, otherwise, preemptCallback will never be called when executing goalCallback the [official examples](https://github.com/ros2/examples/blob/master/rclcpp/actions/minimal_action_server/member_functions.cpp#L110) uses `detach` :crying_cat_face:

rhaschke:
> I just noticed that canceling the goal from the client will not work because for some reason ROS2 expects us to manage the thread to run the goal callback ourselves, otherwise, preemptCallback will never be called when executing goalCallback the official examples uses `detach`

The source code you linked refers to goal acceptance, not cancel acceptance. Are you sure this is related at all?

JafarAbdi:
> The source code you linked refers to goal acceptance, not cancel acceptance. Are you sure this is related at all?

The one I linked is for goal execution (after the goal got accepted by [handle_goal](https://github.com/ros2/examples/blob/master/rclcpp/actions/minimal_action_server/member_functions.cpp#L49-L60)), and it's not possible to call any of these other callbacks while executing it which require us to have a different thread for goalCallback to enable client-side canceling, does this make sense?

PR #326, opened 2021-12-29
Use TOTG rather than IPTP

AndyZe:
IPTP was causing errors like this:

`Time between points 10 and 11 is not strictly increasing, it is 0.357595 and 0.357595 respectively`

Hat tip to @JafarAbdi for finding the cause.

The only concern I have with this PR is, does the robot ever start from non-zero velocity or acceleration? Because TOTG cannot handle that.

(Jafar mentioned loading a plugin to allow any of the time parameterization algorithms. In my opinion that is not necessary because TOTG is clearly the best option at the moment.)

### DIFF
--- a/core/src/merge.cpp
+++ b/core/src/merge.cpp
@@ -36,7 +36,7 @@
 /* Authors: Luca Lach, Robert Haschke */
 
 #include <moveit/task_constructor/merge.h>
-#include <moveit/trajectory_processing/iterative_time_parameterization.h>
+#include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
 
 #include <boost/range/adaptor/transformed.hpp>
 #include <boost/algorithm/string/join.hpp>
@@ -166,7 +166,7 @@ merge(const std::vector<robot_trajectory::RobotTrajectoryConstPtr>& sub_trajecto
 	}
 
 	// add timing
-	trajectory_processing::IterativeParabolicTimeParameterization timing;
+	trajectory_processing::TimeOptimalTrajectoryGeneration timing;
 	timing.computeTimeStamps(*merged_traj, 1.0, 1.0);
 	return merged_traj;
 }
--- a/core/src/solvers/cartesian_path.cpp
+++ b/core/src/solvers/cartesian_path.cpp
@@ -40,7 +40,7 @@
 #include <moveit/task_constructor/moveit_compat.h>
 
 #include <moveit/planning_scene/planning_scene.h>
-#include <moveit/trajectory_processing/iterative_time_parameterization.h>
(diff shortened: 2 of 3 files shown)


JafarAbdi (approved):
LGTM, we already have TOTG as the default parameterization algorithm for the pipeline planners

rhaschke (changes requested):
Instead of switching to another hard-coded time parameterization, I would like to switch to a plugin mechanism. The plugin to be used could be simply specified by a new property `time_parameterization`.

rhaschke:
@AndyZe, are you going to file a new PR implementing [my suggestion](https://github.com/ros-planning/moveit_task_constructor/pull/326#pullrequestreview-842234089)? Hopefully you do, but if not, please file an issue as a reminder. Both Michael and me are rather busy these days...

AndyZe:
I hope to. We shall see. I'm not rolling in free time, either.

JafarAbdi:
@rhaschke I'm working on this, but how do we want to handle having a different set of parameters for the initialize function of the plugin base?

- [IterativeParabolicTimeParameterization](https://github.com/ros-planning/moveit/blob/master/moveit_core/trajectory_processing/include/moveit/trajectory_processing/iterative_time_parameterization.h#L48)
- [IterativeSplineParameterization](https://github.com/ros-planning/moveit/blob/master/moveit_core/trajectory_processing/include/moveit/trajectory_processing/iterative_spline_parameterization.h#L73)
- [TimeOptimalTrajectoryGeneration](https://github.com/ros-planning/moveit/blob/master/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h#L165-L166)

rhaschke:
Great, thanks @JafarAbdi!
Don't handle the time-parameterization classes directly, but use them via a plugin, which then can read corresponding parameters from the ROS parameter server. Actually, I was assuming that all time-parameterization methods already have a corresponding plugin, but only [`IterativeParabolicTimeParameterization`](https://github.com/ros-planning/moveit/blob/8bdf9e57b5e8878f96c0457ab5f498cb06a5d883/moveit_ros/planning/planning_request_adapter_plugins/src/add_time_parameterization.cpp) has!

Thinking deeper, ROS parameters seem not very flexible - there will be only a single global parameter set possible.
Another approach would be to use a new TimeParameterization plugin base, which accepts an MTC `PropertyMap` for its `initialize()` function. Thus you could forward the properties from a stage or the MTC `PlanningInterface`.

We could even wrap those `TimeParameterization` plugins into the existing `AddTimeParameterization` planning adapter plugin to allow instantiating different methods based on corresponding ROS parameters.

Issue #330, opened 2022-01-04
Time between points is not strictly increasing


AndyZe:
Time parameterization occasionally causes an error like this:

`Time between points 10 and 11 is not strictly increasing, it is 0.357595 and 0.357595 respectively`

For reference, #326 proposed a simple solution.

PR #333, opened 2022-01-06
Add trajectory_parameterization parameter to CartesianPath & JointInterpolationPlanner

JafarAbdi:
Fix: https://github.com/ros-planning/moveit_task_constructor/issues/330

Depend on https://github.com/ros-planning/moveit/pull/3021

I think this's simpler than the plugin based solution

### DIFF
--- a/core/include/moveit/task_constructor/solvers/cartesian_path.h
+++ b/core/include/moveit/task_constructor/solvers/cartesian_path.h
@@ -39,7 +39,7 @@
 #pragma once
 
 #include <moveit/task_constructor/solvers/planner_interface.h>
-#include <moveit/trajectory_processing/iterative_time_parameterization.h>
+#include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
 
 namespace moveit {
 namespace task_constructor {
@@ -52,7 +52,7 @@ class CartesianPath : public PlannerInterface
 {
 public:
 	explicit CartesianPath(std::unique_ptr<trajectory_processing::TimeParameterization> trajectory_parameterization =
-	                           std::make_unique<trajectory_processing::IterativeParabolicTimeParameterization>());
+	                           std::make_unique<trajectory_processing::TimeOptimalTrajectoryGeneration>());
 
 	void setStepSize(double step_size) { setProperty("step_size", step_size); }
 	void setJumpThreshold(double jump_threshold) { setProperty("jump_threshold", jump_threshold); }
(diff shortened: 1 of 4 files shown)


rhaschke (commented):
> I think this's simpler than the plugin-based solution.

Sure, it's easier, but less flexible! A plugin-based solution would allow using the other time parameterization methods for MoveIt planning pipelines as well.

core/include/moveit/task_constructor/solvers/cartesian_path.h
```diff
@@ -50,7 +51,8 @@ MOVEIT_CLASS_FORWARD(CartesianPath);
 class CartesianPath : public PlannerInterface
 {
 public:
-	CartesianPath();
+	explicit CartesianPath(std::unique_ptr<trajectory_processing::TimeParameterization> trajectory_parameterization =
+	                           std::make_unique<trajectory_processing::IterativeParabolicTimeParameterization>());
```
AndyZe, line 55: No comment on the high-level strategy. But please make the default TOTG

JafarAbdi:
> A plugin-based solution would allow using the other time parameterization methods for MoveIt planning pipelines as well.

I thought this's already the case [TimeOptimalTrajectoryGeneration](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/planning_request_adapter_plugins/src/add_time_optimal_parameterization.cpp), [IterativeParabolicTimeParameterization](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/planning_request_adapter_plugins/src/add_time_parameterization.cpp), [IterativeSplineParameterization](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/planning_request_adapter_plugins/src/add_iterative_spline_parameterization.cpp) am I missing something ?

rhaschke (approved):
> I thought this's already the case TimeOptimalTrajectoryGeneration, IterativeParabolicTimeParameterization, IterativeSplineParameterization.

You are right. I was misled by the too generic name `AddTimeParameterization` used for the IPTP plugin...

Approving. Depends on https://github.com/ros-planning/moveit/pull/3021 and https://github.com/ros-planning/moveit/pull/3023.

AndyZe:
I've tested this. It works well.  :+1:

AndyZe (approved):

rhaschke (commented):
@JafarAbdi, may I ask you to change the API such that the time parameterization is specified as a property of those classes?
Time parameterization is also used by `merge(...)`:
https://github.com/ros-planning/moveit_task_constructor/blob/4fa8c10f4456e893e09212b51a06b3d2ec4e9266/core/src/merge.cpp#L168-L170
which is used by `Connect` and `Merger` stages. Do you mind adding this feature there as well?

henningkayser (approved):
This looks very straight-forward to me. I agree it makes much more sense than using plugins here. Do you think you could address Robert's comments so that this can get merged?

PR #335, opened 2022-02-01
move constructor deletes implicitly declared copy assignment operator

jrgnicho:
Not sure why a move constructor was explicitly declared but it has the side effect of declaring the implicitly-declared copy assignment operator as deleted.  See here for reference https://en.cppreference.com/w/cpp/language/copy_assignment

### DIFF
--- a/core/include/moveit/task_constructor/storage.h
+++ b/core/include/moveit/task_constructor/storage.h
@@ -127,7 +127,6 @@ public:
 
 	/// copy an existing InterfaceState, but not including incoming/outgoing trajectories
 	InterfaceState(const InterfaceState& other);
-	InterfaceState(InterfaceState&& other) = default;
 
 	inline const planning_scene::PlanningSceneConstPtr& scene() const { return scene_; }
 	inline const Solutions& incomingTrajectories() const { return incoming_trajectories_; }


rhaschke:
I'm not sure we should have an assignment operator. Can you provide an example use case, please?

jrgnicho:
Well I just updated a project to the most recent version of MTC and it broke my build.  Mainly I have a function where I pass an InterfaceState by reference as follows:
```
std::vector<robot_trajectory::RobotTrajectoryPtr> computeSolution(const moveit::task_constructor::InterfaceState& from,
                            moveit::task_constructor::InterfaceState& to,
                            EigenSTL::vector_Isometry3d& process_toolpath)
{
 // compute trajectories
  ...
// set final of planning scene after reaching last waypoint in trajectories
  planning_scene::PlanningScenePtr end_scene = ...
  end_scene->setCurrentState(cartesian_traj->getLastWayPoint());

// save scene into Interface state
  to = InterfaceState(end_scene);  // This fails to build when there's no assignment operator
}
```

This function built fine in an older implementation of MTC so the assignment operator was supported then

rhaschke:
Thanks. I decided to _explicitly_ enable both the move constructor and the copy assignment operator.

jrgnicho:
@rhaschke thanks, will this be added in another PR?

rhaschke:
It's already merged: ca38d11303ba8a0ea7da6c917fda011eb9cefd45

jrgnicho:
@rhaschke excellent, thank you

PR #339, opened 2022-03-03
Make TimeParamerization configurable

### LINKED ISSUE #330: Time between points is not strictly increasing

rhaschke:
Replaces #333, implementing https://github.com/ros-planning/moveit_task_constructor/pull/333#pullrequestreview-850576619. Fixes #330.
The TimeParameterization method becomes a property in corresponding stages and PlannerInterfaces.
This allows users to configure them easily.

### DIFF
--- a/core/include/moveit/task_constructor/moveit_compat.h
+++ b/core/include/moveit/task_constructor/moveit_compat.h
@@ -39,6 +39,7 @@
 #pragma once
 
 #include <moveit/version.h>
+#include <moveit/macros/class_forward.h>
 
 #define MOVEIT_VERSION_GE(major, minor, patch)                                                \
 	(MOVEIT_VERSION_MAJOR * 1'000'000 + MOVEIT_VERSION_MINOR * 1'000 + MOVEIT_VERSION_PATCH >= \
@@ -57,3 +58,9 @@
 #define MOVEIT_HAS_OBJECT_POSE MOVEIT_VERSION_GE(1, 1, 6)
 
 #define MOVEIT_HAS_STATE_RIGID_PARENT_LINK MOVEIT_VERSION_GE(1, 1, 6)
+
+#if !MOVEIT_VERSION_GE(1, 1, 9)
+namespace trajectory_processing {
+MOVEIT_CLASS_FORWARD(TimeParameterization);
+}
+#endif
--- a/core/include/moveit/task_constructor/moveit_compat.h
+++ b/core/include/moveit/task_constructor/moveit_compat.h
@@ -60,6 +60,7 @@
 #define MOVEIT_HAS_STATE_RIGID_PARENT_LINK MOVEIT_VERSION_GE(1, 1, 6)
 
 #if !MOVEIT_VERSION_GE(1, 1, 9)
+// the pointers are not yet available
 namespace trajectory_processing {
 MOVEIT_CLASS_FORWARD(TimeParameterization);
 }
--- a/core/src/merge.cpp
+++ b/core/src/merge.cpp
@@ -36,7 +36,6 @@
 /* Authors: Luca Lach, Robert Haschke */
 
 #include <moveit/task_constructor/merge.h>
-#include <moveit/trajectory_processing/iterative_time_parameterization.h>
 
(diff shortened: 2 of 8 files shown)


core/src/merge.cpp
```diff
@@ -166,8 +166,7 @@ merge(const std::vector<robot_trajectory::RobotTrajectoryConstPtr>& sub_trajecto
 	}
 
 	// add timing
-	trajectory_processing::IterativeParabolicTimeParameterization timing;
-	timing.computeTimeStamps(*merged_traj, 1.0, 1.0);
+	time_parameterization.computeTimeStamps(*merged_traj, 1.0, 1.0);
```
v4hn, line 169: This can (and could before) break a lower specified dynamics limit in one of the merged trajectories. The merged trajectory might suddenly be "inexplicably" faster.

But the real solution to this is not adding yet another set of parameters, but implement a "proper" merge, that does not need to compute a whole new timing for the trajectory, but resamples the trajectories as needed.
So I guess it's best to leave this as is until then.

core/include/moveit/task_constructor/moveit_compat.h
```diff
@@ -57,3 +58,10 @@
 #define MOVEIT_HAS_OBJECT_POSE MOVEIT_VERSION_GE(1, 1, 6)
 
 #define MOVEIT_HAS_STATE_RIGID_PARENT_LINK MOVEIT_VERSION_GE(1, 1, 6)
+
+#if !MOVEIT_VERSION_GE(1, 1, 9)
+// the pointers are not yet available
+namespace trajectory_processing {
+MOVEIT_CLASS_FORWARD(TimeParameterization);
+}
+#endif
```
v4hn, line 67: I just saw you [backported the polymorphic base class change to melodic](https://github.com/ros-planning/moveit/commit/393815e4872de6349ae3784360dbe276a9abda7a). So this should be unnecessary.
But Jafar and you both did not notice the missing pointers in the original patch, so I just filed [this PR](https://github.com/ros-planning/moveit/pull/3078). It might be worth creating a patch release for melodic with this addition as well. If you should do that, these lines can go away
```suggestion
```

rhaschke:
This commit finally breaks compatibility to older MoveIt releases, namely Melodic. How should we handle this? Drop Melodic? 
I'm fine with that.

v4hn (approved):
Mostly looks good. I'll leave it to you whether you want to deal with the missing pointers in upstream before merging.

I would be happier with this if ...
- the time parameterization were an actual plugin in MoveIt
- if the triple `<plugin, vel_scaling, acc_scaling>` would be somehow one property instead of three needed in multiple places
- if we could add additional string overloads to load these properties instead of having to pass shared_ptr.

core/src/stages/connect.cpp
```diff
@@ -54,6 +58,8 @@ Connect::Connect(const std::string& name, const GroupPlannerVector& planners) :
 	p.declare<MergeMode>("merge_mode", WAYPOINTS, "merge mode");
 	p.declare<moveit_msgs::Constraints>("path_constraints", moveit_msgs::Constraints(),
 	                                    "constraints to maintain during trajectory");
+	properties().declare<TimeParameterizationPtr>("time_parameterization",
```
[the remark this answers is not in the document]
rhaschke: Yeah, you are right.

core/src/merge.cpp
rhaschke: Please file an issue as a reminder...

core/include/moveit/task_constructor/moveit_compat.h
rhaschke: Until we have a new release of MoveIt, this addition is required. Do you propose a patch release of MoveIt before? 
I could do that over the weekend indeed...

v4hn: I guess you can remove this now. :)

JafarAbdi (approved):

Issue #350, opened 2022-04-02
ros2 pickplace.launch.py demo planning fails


Jeisongarcia9713:
Hello, I am trying to make work the pick and place demo in ros2 galactic, the demo.launch.py is throwing without any error even though when I run the pick and place launch and the plain fails but there is no too much detailed about the problem.

```
ros2 launch moveit_task_constructor_demo pickplace.launch.py
```
[INFO] [launch]: All log files can be found below /home/jeison/.ros/log/2022-04-01-20-27-21-530037-jeison-134961
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [pick_place_demo-1]: process started with pid [134963]
[pick_place_demo-1] [INFO] [1648862842.432076827] [moveit_task_constructor_demo]: Loading task parameters
[pick_place_demo-1] [INFO] [1648862842.432894505] [moveit_task_constructor_demo]: Initializing task pipeline
[pick_place_demo-1] [INFO] [1648862842.465088839] [moveit_rdf_loader.rdf_loader]: Loaded robot model in 0.00555339 seconds
[pick_place_demo-1] [INFO] [1648862842.465139221] [moveit_robot_model.robot_model]: Loading robot model 'panda'...
[pick_place_demo-1] Link panda_link1 had 1 children
[pick_place_demo-1] Link panda_link2 had 1 children
[pick_place_demo-1] Link panda_link3 had 1 children
[pick_place_demo-1] Link panda_link4 had 1 children
[pick_place_demo-1] Link panda_link5 had 1 children
[pick_place_demo-1] Link panda_link6 had 1 children
[pick_place_demo-1] Link panda_link7 had 1 children
[pick_place_demo-1] Link panda_link8 had 1 children
[pick_place_demo-1] Link panda_hand had 2 children
[pick_place_demo-1] Link panda_leftfinger had 0 children
[pick_place_demo-1] Link panda_rightfinger had 0 children
[pick_place_demo-1] [WARN] [1648862842.515616923] [moveit_task_constructor_demo]: Failed to find 'ompl.planning_plugin'. Attempting to load pipeline from old parameter structure. Please update your MoveIt config.
[pick_place_demo-1] [INFO] [1648862842.574433035] [moveit.ros_planning.planning_pipeline]: Using planning interface 'OMPL'
[...]

Issue #26, opened 2018-02-26
non-empty solution in `onNewSolution`


v4hn:
After longer execution (2-3 minutes) of my 10+ stages serial container and possibly close to the end of exhaustive planning I currently see the following quite often:

```
moveit_task_constructor/core/src/container.cpp:347: virtual void moveit::task_constructor::SerialContainer::onNewSolution(const moveit::task_constructor::SolutionBase&): Assertion `solution.empty()' failed.
```

I did not investigate further yet, because I guess you have some ideas what the problem might be @rhaschke  .

rhaschke:
For me the assertion never failed: As every solution should be moved into the sorted list, I cannot really see, why the assertion fails. However, I provided a quick fix in #27.
Maybe, you should provide your stage pipeline for further investigation.

v4hn:
hopefully obsolete, I haven't seen the problem in a long time now.

PR #355, opened 2022-04-20
[ros2] Add Stage property to assign a list of controllers to use when executing the planned trajectory

schornakj:
This is essentially a ROS2 version of the work in #123, plus some additions based on @henningkayser's review of that PR.

- Add a new `TrajectoryExecutionInfo` ROS message to hold info used when executing the trajectory. Right now it just contains a vector of strings to define controller names used with the trajectory. Add an instance of this message as a field in the `SubTrajectory` message.
- Add a new `TrajectoryExecutionInfo` struct to hold this info as an MTC property.
- Propagate the `"trajectory_execution_info"` property from each stage into the `SubTrajectory` messages output when the task is planned.
- Modify the `ExecuteTaskSolution` capability to copy the controller names into the `RobotTrajectory` prior to execution.

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -171,6 +171,12 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 		exec_traj.trajectory = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
 		exec_traj.trajectory->setRobotTrajectoryMsg(state, sub_traj.trajectory);
 
+		if (!sub_traj.controller_names.empty()) {
+			for (auto cn : sub_traj.controller_names) {
+				exec_traj.controller_names_.push_back(cn.data.data());
+			}
+		}
+
 		/* TODO add action feedback and markers */
 		exec_traj.effect_on_success = [this, sub_traj,
 		                               description](const plan_execution::ExecutableMotionPlan* /*plan*/) {
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -170,12 +170,7 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 		}
 		exec_traj.trajectory = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
 		exec_traj.trajectory->setRobotTrajectoryMsg(state, sub_traj.trajectory);
-
-		if (!sub_traj.controller_names.empty()) {
-			for (auto cn : sub_traj.controller_names) {
-				exec_traj.controller_names_.push_back(cn.data.data());
-			}
-		}
+		exec_traj.controller_name = sub_traj.execution_info.controller_names;
 
 		/* TODO add action feedback and markers */
 		exec_traj.effect_on_success = [this, sub_traj,
--- a/core/include/moveit/task_constructor/stage.h
+++ b/core/include/moveit/task_constructor/stage.h
@@ -201,6 +201,9 @@ public:
 	/// marker namespace of solution markers
 	const std::string& markerNS() { return properties().get<std::string>("marker_ns"); }
 
+	void setControllers(std::vector<std::string> controllers) { setProperty("controllers", controllers); }
+	std::vector<std::string> controllers() { return properties().get<std::vector<std::string>>("controllers"); }
+
 	/// forwarding of properties between interface states
 	void forwardProperties(const InterfaceState& source, InterfaceState& dest);
 	std::set<std::string> forwardedProperties() const {
--- a/core/include/moveit/task_constructor/stage.h
+++ b/core/include/moveit/task_constructor/stage.h
@@ -39,6 +39,7 @@
 
 #pragma once
 
+#include "trajectory_execution_info.h"
 #include "utils.h"
 #include <moveit/macros/class_forward.h>
 #include <moveit/task_constructor/storage.h>
@@ -201,8 +202,13 @@ public:
 	/// marker namespace of solution markers
 	const std::string& markerNS() { return properties().get<std::string>("marker_ns"); }
 
-	void setControllers(std::vector<std::string> controllers) { setProperty("controllers", controllers); }
-	std::vector<std::string> controllers() { return properties().get<std::vector<std::string>>("controllers"); }
+	/// Set and get info to use when executing the stage's trajectory
+	void setTrajectoryExecutionInfo(TrajectoryExecutionInfo trajectory_execution_info) {
+		setProperty("trajectory_execution_info", trajectory_execution_info);
+	}
+	TrajectoryExecutionInfo trajectoryExecutionInfo() {
+		return properties().get<TrajectoryExecutionInfo>("trajectory_execution_info");
+	}
 
 	/// forwarding of properties between interface states
 	void forwardProperties(const InterfaceState& source, InterfaceState& dest);
--- a/core/include/moveit/task_constructor/trajectory_execution_info.h
+++ b/core/include/moveit/task_constructor/trajectory_execution_info.h
@@ -1,0 +1,47 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2022, PickNik Inc.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of PickNik Inc. nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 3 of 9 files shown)


IgorUnlimited:
Pick/place still not work. Planning failed ... after running:

1) `ros2 launch moveit_task_constructor_demo demo.launch.py`
2) `ros2 launch moveit_task_constructor_demo pickplace.launch.py`

schornakj:
> Pick/place still not work. Planning failed ... after running:
> 
>     1. `ros2 launch moveit_task_constructor_demo demo.launch.py`
> 
>     2. `ros2 launch moveit_task_constructor_demo pickplace.launch.py`

Was your comment meant for #350? This PR wasn't intended as a fix for the issue described over there.

schornakj:
I'll tag @JafarAbdi here since we were talking about this PR earlier today.

sjahr:
@rhaschke @v4hn Other than the conflict, what is blocking this PR from being merged? I saw this comment in #26:
> Anyway, I was not happy with that approach because it focused only on controller names.

What else would you add? To move forward I think it would be good to review and merge this PR to make this usable and build on top of it. A nice follow-up PR could involve hiding the actual controller names behind more abstracted config names such as complaint trajectory execution (e.g. JTC + admittance control) and offer such an API in MoveIt itself

core/include/moveit/task_constructor/stage.h
```diff
@@ -201,6 +202,14 @@ class Stage
 	/// marker namespace of solution markers
 	const std::string& markerNS() { return properties().get<std::string>("marker_ns"); }
 
+	/// Set and get info to use when executing the stage's trajectory
```
sjahr, line 205: Nit: Document functions individually

core/include/moveit/task_constructor/trajectory_execution_info.h
```diff
@@ -0,0 +1,66 @@
...
+ *********************************************************************/
+
+/* Authors: Joe Schornak
+   Desc:    Define a struct to hold options used to configure trajectory execution
+*/
+
+#pragma once
+
+#include <moveit_task_constructor_msgs/msg/trajectory_execution_info.hpp>
+#include <std_msgs/msg/string.hpp>
```
sjahr, line 42: Is this used anywhere?

core/include/moveit/task_constructor/trajectory_execution_info.h
```diff
@@ -0,0 +1,66 @@
...
+#include <string>
+#include <vector>
+
+namespace {
+using TrajectoryExecutionInfoMsg = moveit_task_constructor_msgs::msg::TrajectoryExecutionInfo;
+}
+
+namespace moveit {
+namespace task_constructor {
+struct TrajectoryExecutionInfo
```
sjahr, line 52: Can you add documentation for this class and its members

sjahr (commented):
This is useful to pass information downwards, how a trajectory should be used. I'd be in favor of merging it once more documentation is added

core/src/container.cpp
```diff
@@ -283,6 +283,8 @@ void ContainerBase::insert(Stage::pointer&& stage, int before) {
 	if (!stage)
 		throw std::runtime_error(name() + ": received invalid stage pointer");
 
+	stage->setTrajectoryExecutionInfo(this->trajectoryExecutionInfo());
```
v4hn, line 286: This would overwrite whatever was specified for the stage, right?

core/src/storage.cpp
```diff
@@ -190,6 +190,10 @@ void SubTrajectory::fillMessage(moveit_task_constructor_msgs::msg::Solution& msg
 	moveit_task_constructor_msgs::msg::SubTrajectory& t = msg.sub_trajectory.back();
 	SolutionBase::fillInfo(t.info, introspection);
 
+	const auto trajectory_execution_info =
+	    creator()->properties().get<TrajectoryExecutionInfo>("trajectory_execution_info");
+	t.execution_info = trajectory_execution_info.toMsg();
```
v4hn, line 195: This way the execution info cannot be set *per solution*, but it is entirely valid that the same stage would generate solutions with, e.g., a left or a right arm. Accordingly, the controllers that you would need to execute the solution varies per solution.

v4hn (commented):
I still believe this is a special case of solution properties which [Robert summarized recently](https://github.com/ubi-agni/moveit_task_constructor/issues/26) and I think it doesn't make sense to merge partial support for this very specific use or modifying the messages for it.
For the execution through `MoveItCpp` there's not much missing to get the generic properties to work.

core/src/container.cpp
sjahr, line 286:
```suggestion
	if (stage->trajectoryExecutionInfo().controller_names.empty()) {
		stage->setTrajectoryExecutionInfo(this->trajectoryExecutionInfo());
	}
```

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -168,8 +168,9 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 				RCLCPP_DEBUG(LOGGER, "Using JointModelGroup '%s' for execution", group->getName().c_str());
 			}
 		}
-		exec_traj.trajectory = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
-		exec_traj.trajectory->setRobotTrajectoryMsg(state, sub_traj.trajectory);
+		exec_traj.trajectory_ = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
+		exec_traj.trajectory_->setRobotTrajectoryMsg(state, sub_traj.trajectory);
+		exec_traj.controller_names_ = sub_traj.execution_info.controller_names;
```
sjahr, line 173:
```suggestion
		exec_traj.trajectory = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
		exec_traj.trajectory->setRobotTrajectoryMsg(state, sub_traj.trajectory);
		exec_traj.controller_name = sub_traj.execution_info.controller_names;
```

core/include/moveit/task_constructor/trajectory_execution_info.h
sjahr, line 42:
```suggestion
```

core/src/storage.cpp
sjahr: Is there a solution available to pick the trajectory info per solution/ any need for it at the moment? In order to unblock this PR I think this could be addressed in a follow-up issue when there is the need to specify trajectory info per solution.

sjahr (approved):
@v4hn @rhaschke I think we should merge this. We've been successfully using these changes for ~1 year in our MTC fork. This PR might only cover a special case of solution properties but nobody has started working on #26 so far and merging this won't prevent anyone from refactoring and generalizing this.

core/src/storage.cpp
rhaschke: We would need to register custom user callbacks to fill those fields in a dynamic fashion.

rhaschke (commented):
Generally, I'm fine to merge this temporarily. But be prepared that we change/generalize the API in future. I have to admit that MTC wasn't very high on my prior list lately.

While you rely on #123, I'm surprised to not see a cherry-pick of these commits.
To honor @llach original work on this (and keep his authorship), I suggest to cherry-pick (and squash) #123 and then add your modificiations on top of it. Please also rebase to the ros2 branch, such that we can merge via a true merge-commit.

sjahr:
@rhaschke Done :heavy_check_mark:

core/include/moveit/task_constructor/trajectory_execution_info.h
rhaschke, line 51: Why don't you simply define a synonym:
`using TrajectoryExecutionInfo = moveit_task_constructor_msgs::msg::TrajectoryExecutionInfo`?

core/src/container.cpp
```diff
@@ -365,8 +365,13 @@ void ContainerBase::insert(Stage::pointer&& stage, int before) {
 	if (!stage)
 		throw std::runtime_error(name() + ": received invalid stage pointer");
 
+	if (stage->trajectoryExecutionInfo().controller_names.empty()) {
+		stage->setTrajectoryExecutionInfo(this->trajectoryExecutionInfo());
+	}
```
rhaschke, line 370: This is not the correct way of inheriting properties from parent to child (I think this is intended here).
Users should make this inheritance explicit _in their code_ via:
`Property::configureInitFrom(Stage::PARENT, "trajectory_execution_info");`

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -168,14 +168,9 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
...
-		exec_traj.trajectory = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
-		exec_traj.trajectory->setRobotTrajectoryMsg(state, sub_traj.trajectory);
-
-		if (!sub_traj.controller_names.empty()) {
-			for (auto cn : sub_traj.controller_names) {
-				exec_traj.controller_names_.push_back(cn.data.data());
-			}
-		}
+		exec_traj.trajectory_ = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
+		exec_traj.trajectory_->setRobotTrajectoryMsg(state, sub_traj.trajectory);
```
rhaschke: The names (`_`) are changed back and forth. Please further cleanup history to have minimal changes per commit. 
This also facilitates reviewing :wink:

rhaschke (commented):
Thanks for cleaning up the history. Please squash the `_` fixups into the first commit.
I think the 3rd commit should be dropped completely (see comment below).

core/src/container.cpp
sjahr: The intend here is to make inheriting TrajectoryExecutionInfo from the parent in the container default except, the stage has explicitly configured it's own trajectory execution info. If a user has already done this via configureInitFrom, nothing happens here. Do you think that is a bad default?

sjahr:
@rhaschke Can you give this a final review?

sea-bass (approved):
This has been working for us for a while, so I'm inclined to approve it so we can move away from our fork and use the upstream MTC.

@rhaschke if there is some concrete changes in this PR you would like to see, please let @sjahr know before he merges it... or if you prefer to revisit this later, that works too.

core/src/container.cpp
rhaschke: Yes, I do consider this a bad default, because it contradicts the philosophy of _explicit_ property inheritance, which @v4hn requested from the very beginning. We should not deviate from that philosophy. Please revert that part.

sjahr: https://github.com/ros-planning/moveit_task_constructor/pull/502

rhaschke:
Looks like this PR introduced a segfault in unit tests:
https://github.com/ros-planning/moveit_task_constructor/actions/runs/6730632997

Please investigate!

sjahr:
:thinking: Why wasn't it caught by CI?

rhaschke:
Not sure. Maybe it is just due to a recent change in MoveIt2?

sjahr:
I'll investigate it :+1:

PR #320, opened 2021-12-14
Fix using IKFrame with MoveRelative

v4hn:
`direction` was interpreted relative to the link even when an `ik_frame` was specified.

Solves #318 .

PR #357, opened 2022-04-21
Fix using IKFrame with MoveRelative in ROS2

swiz23:
@JafarAbdi,

This fix address the same issue as https://github.com/ros-planning/moveit_task_constructor/pull/320/ but for the ros2 branch. I've tested this by commanding both a VectorGoal and TwistGoal using the MoveRelative action type and it worked as expected.


JafarAbdi:
The fix is already a part of https://github.com/ros-planning/moveit_task_constructor/pull/322

PR #362, opened 2022-05-14
Fix Eigen build issues on some platforms

### LINKED ISSUE #361: Commit 36459e38 breaks build for Eigen3
On some platforms, `eigen3_cmake_module` is required to fix non-standard env vars in the default provided `FindEigen3`, as described [here](https://github.com/ros2/eigen3_cmake_module).  To work correctly, the package must call `find_package(eigen3_cmake_module)` *before* `find_package(Eigen3)`.

Commit 36459e38 reverses the order of these two calls in `rviz_marker_tools/CMakeLists.txt`, which causes its build to fail on platforms where this workaround is required.

Fix is to revert to the original ordering in `rviz_marker_tools/CMakeLsits.txt`.  `eigen3_cmake_module` before `Eigen3`.

AndyZe:
Fixes #361. @JeremyZoss please take a look

### DIFF
--- a/rviz_marker_tools/CMakeLists.txt
+++ b/rviz_marker_tools/CMakeLists.txt
@@ -4,8 +4,8 @@ project(rviz_marker_tools)
 find_package(ament_cmake REQUIRED)
 
 set(THIS_PACKAGE_INCLUDE_DEPENDS
-    Eigen3
     eigen3_cmake_module
+    Eigen3
     geometry_msgs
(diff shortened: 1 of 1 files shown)


JafarAbdi (approved):

JeremyZoss:
> Fixes #361. @JeremyZoss please take a look

Works for me.  Thanks for prompt fix.

PR #363, opened 2022-05-17
[ros2]: Fixed preempting of execution

JafarAbdi:
Same as https://github.com/JafarAbdi/moveit_task_constructor/pull/6


JafarAbdi:
@mechwiz Please review

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -95,7 +97,8 @@ void ExecuteTaskSolutionCapability::initialize() {
 	    ActionServerType::CancelCallback(
 	        std::bind(&ExecuteTaskSolutionCapability::preemptCallback, this, std::placeholders::_1)),
 	    ActionServerType::AcceptedCallback(
-	        std::bind(&ExecuteTaskSolutionCapability::goalCallback, this, std::placeholders::_1)));
+	        std::bind(&ExecuteTaskSolutionCapability::goalCallback, this, std::placeholders::_1)),
```
AndyZe, line 100: I believe the `std::bind` isn't needed here since 1 argument is expected

core/src/task.cpp
```diff
@@ -301,9 +301,17 @@ moveit_msgs::msg::MoveItErrorCodes Task::execute(const SolutionBase& s) {
...
+		if (pimpl()->preempt_requested_) {
+			auto cancel_future = ac->async_cancel_goal(goal_handle);
+			if (rclcpp::spin_until_future_complete(node, cancel_future) != rclcpp::FutureReturnCode::SUCCESS) {
+				RCLCPP_ERROR(node->get_logger(), "Could not preempt execution");
+				return error_code;
+			} else {
+				error_code.val = moveit_msgs::msg::MoveItErrorCodes::PREEMPTED;
+				return error_code;
+			}
+		}
```
v4hn, line 314: Is that really the best we can do for this kind of feature? 10ms waits are quite bad style :(

v4hn (commented):
preempt was added to allow the user to stop *planning*.
As *execution* is just a thin wrapper around the action server it seems like a clumsy interface to repurpose preempt for execution instead.

Wouldn't it be better and more flexible to expose the action client to the user to cancel execution there?

PR #366, opened 2022-05-25
Fix parameter in ik_clearance_cost demo

stephanie-eng:


JafarAbdi:
Nice catch, thanks! @stephanie-eng

PR #369, opened 2022-05-31
[ros2]: Launch files simplifications

JafarAbdi:
Depends on: https://github.com/ros-planning/moveit2/pull/1283 and https://github.com/ros-planning/moveit_resources/pull/139


rhaschke:
I prefer to keep the URDF definition explicit.

PR #374, opened 2022-06-15
Remove collision objects as final step

tylerjw:
Signed-off-by: Tyler Weaver <tyler@picknik.ai>

This is useful because if you write a stage that detaches and removes an object it does not perform as you'd expect.


rhaschke:
@tylerjw, could you please explain the issue solved by this PR in more detail? I don't yet see the actual difference to the previous code, except that objects are removed after ACM edits. However, I don't see a reason for this yet.

rhaschke:
> @tylerjw, could you please explain the issue solved by this PR in more detail?

Ping, @tylerjw.

tylerjw:
I'm sorry, I don't remember why I did this. I only remember doing this on a previous project.

rhaschke:
Hm. In that case, I suggest to close the PR for now...

PR #375, opened 2022-07-12
[ros2]: Add moveit cost function for IK solver as parameter for Cartesian path solver

wyattrees:
This incorporates a newly added MoveIt feature, kinematics cost functions, to pass to `CartesianInterpolator`.


rhaschke:
Obviously, the new feature isn't yet part of the `rolling-source` docker image.

wyattrees:
The docker images must not have been updated at the time that CI ran the first time, but all seems good now.

rhaschke (approved):

PR #376, opened 2022-07-21
Printing collision info for invalid IK solutions.

NFiedler:
Instead of just marking an invalid IK solution in red, a comment is added describing the (and only one) collision causing the problem. 
This was done with a lot of help from @v4hn .

### DIFF
--- a/core/src/stages/compute_ik.cpp
+++ b/core/src/stages/compute_ik.cpp
@@ -84,7 +84,14 @@ void ComputeIK::setTargetPose(const Eigen::Isometry3d& pose, const std::string& 
 }
 
 // found IK solutions
-using IKSolutions = std::vector<std::vector<double>>;
+
+struct IKSolution {
+	std::vector<double> joint_positions;
+	bool feasible;
+	collision_detection::Contact contact;
+};
+
+using IKSolutions = std::vector<IKSolution>;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


core/src/stages/compute_ik.cpp
```diff
@@ -353,14 +360,25 @@ void ComputeIK::compute() {
...
 		ik_solutions.emplace_back();
-		state->copyJointGroupPositions(jmg, ik_solutions.back());
+		auto& solution{ ik_solutions.back() };
+		state->copyJointGroupPositions(jmg, solution.joint_positions);
+		collision_detection::CollisionRequest req;
+		collision_detection::CollisionResult res;
+		req.contacts = true;
+		req.max_contacts = 1;
+		scene->checkCollision(req, res, *state);
+		solution.feasible = ignore_collisions || !res.collision;
```
rhaschke, line 375: When `ignore_collisions` is `true`, the collision checking should be skipped, shouldn't it?

core/src/stages/compute_ik.cpp
```diff
@@ -388,15 +406,17 @@ void ComputeIK::compute() {
 			solution.setComment(s.comment());
 			std::copy(frame_markers.begin(), frame_markers.end(), std::back_inserter(solution.markers()));
 
-			if (succeeded && i + 1 == ik_solutions.size())
+			if (ik_solutions[i].feasible)
```
rhaschke, line 409: I guess this new condition just aims to be more obvious.
Do we still need the variable `succeeded`?

rhaschke (changes requested):
Looks good in general. But there is one critical remark:

core/src/stages/compute_ik.cpp
v4hn: We discussed tha when writing the patch and I am in favor of *always* doing collisions checks here to provide the feedback even if the solution is still marked as feasible. I notice now that checking was indeed skipped before in this case (as obviously no feedback was generated from it), but I would still change it in favor of the additional feedback.

If you insist it's better to retain the current behavior here in favor of performance we can do it as well, but `ignore_collisions=true` was always more of a debugging flag for me.

core/src/stages/compute_ik.cpp
v4hn: It's [used below](https://github.com/ros-planning/moveit_task_constructor/pull/376/files#diff-eb4791e6fc52e29c2b9fb96a606928a1bf587989c5ebf29af186a50137ecc159R434).

core/src/stages/compute_ik.cpp
rhaschke: You are right. The default case is to perform collision checking anyway. I just wanted to point out the difference in behavior.

rhaschke (approved):

rhaschke:
Please fix the formatting issue before merging.

PR #380, opened 2022-08-28
Fix Cartesian interpolation

### LINKED ISSUE #359: Using MTC to Plan a Circular Path
@JafarAbdi,

I'm having issues trying to use the ros2 branch of MTC to plan a circular path around a non-TCP frame (that already includes PR #322). I wrote up a question at ROS Answers describing the problem. Instead of copying and pasting it, I'm linking to it [here](https://answers.ros.org/question/400041/using-mtc-to-plan-a-circular-path/). In a nutshell though the two main questions I have are:
- How to use MTC to plan a circular path around a non-TCP frame? Specifying the frame_id in a goal message to be the non-TCP frame has no impact (circular path is still generated only around the TCP frame).
- While rotating around the non-TCP frame, how to keep the orientation of the TCP constant?

rhaschke:
This fixes #359. There were actually two major issues:
1. #304 broke Cartesian interpolation, wrongly calculating the target frame when rotations were involved (see videos below, were the endeffector should move and rotate about the green axis)
2. MoveIt itself didn't yet support Cartesian interpolation for non-link frames. To fix this, I filed https://github.com/ros-planning/moveit/pull/3197.

old:

https://user-images.githubusercontent.com/5376030/187067390-dd6e7af7-80e9-4b2a-8623-e3b844185e89.mp4

new:

https://user-images.githubusercontent.com/5376030/187067403-528cd575-678a-44ef-8570-5c437a8e4520.mp4

### DIFF
--- a/core/include/moveit/task_constructor/solvers/cartesian_path.h
+++ b/core/include/moveit/task_constructor/solvers/cartesian_path.h
@@ -66,8 +66,8 @@ public:
 	          const moveit_msgs::Constraints& path_constraints = moveit_msgs::Constraints()) override;
 
 	bool plan(const planning_scene::PlanningSceneConstPtr& from, const moveit::core::LinkModel& link,
-	          const Eigen::Isometry3d& target, const moveit::core::JointModelGroup* jmg, double timeout,
-	          robot_trajectory::RobotTrajectoryPtr& result,
+	          const Eigen::Isometry3d& offset, const Eigen::Isometry3d& target, const moveit::core::JointModelGroup* jmg,
+	          double timeout, robot_trajectory::RobotTrajectoryPtr& result,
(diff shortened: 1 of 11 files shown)


rhaschke:
Closing and reopening to trigger CI.

rhaschke:
@v4hn, looks like C++14 standard is not default (or downgraded by other packages?) in Melodic.
Could this be related to https://github.com/ros-planning/moveit_tutorials/commit/cf79c874dc0d720aecb3bff7f8ba645ef542bce3?

v4hn:
> This actually fails with clang-6.0 on Bionic!

You lost me there, what?
Why did you cite the moveit_tutorials commit that succeeded on ci (including the `master-source` job on 18.04)?

#386 should consistently cleanup the standard specification now. I don't like the fact that we need the workaround for 5-year old build-system bugs, but at least the macro hides it consistently.

PR #382, opened 2022-09-06
[ros2] Fix Cartesian interpolation

### LINKED ISSUE #359: Using MTC to Plan a Circular Path

rhaschke:
Merge of #380 into ros2 branch (and resolving conflicts). This fixes #359.


rhaschke:
I merged the master branch, including #380, into ros2.

PR #386, opened 2022-10-22
set useful build-options from MoveIt also in MTC

v4hn:
This includes the workaround to specify c++14 explicitly for Ubuntu 18.04 (checking the gcc version is not a good way to do that that, but it works and getting it more accurate is a lot of work...)

This should implicitly address #385 and Robert's problem in https://github.com/ros-planning/moveit_task_constructor/pull/380#issuecomment-1286096646 . Sorry for bypassing CI for this, as always a bad idea.

The actual problem is that Ubuntu 18.04 comes with a cmake-config/qt5 mkspecs configuration that forces c++11 by default although it's compatible with 14, which breaks the fact that c++14 is the default in compilers in 18.04 anyway...

Why do we want melodic/Ubuntu 18.04 support in the master branches again?

Fixes #385 .

PR #387, opened 2022-10-23
Fix CI + build issues

rhaschke:
Cleanup of #386, split into two branches: source code changes (fixing warnings + tests) and CI fixups.
@v4hn, I suggest merging via (force-)push from cmd line (force-pushing because I dropped your two initial, but breaking commits).
We still have the Eigen warnings in clang builds. Hence the draft status. I will try a different order of includes...


rhaschke:
> We still have the Eigen warnings in clang builds. Hence the draft status. I will try a different order of includes...

Didn't work :disappointed:

v4hn:
I believe I just pushed a commit that moves eigen3 to `-isystem` includes and thus fixes the warning from eigen, but adding `-Werror` back triggers `-Wunused-function` warnings in the mtc codebase that should not happen because in the same block we add `-Wno-unused-function`... Any idea @rhaschke ?

rhaschke:
> I believe I just pushed a commit that moves eigen3 to `-isystem` includes and thus fixes the warning from Eigen.

Yes, this fixed the problem for one package. In principle all packages would need this fix...
Instead, I declared all `catkin_INCLUDE_DIRS` as system includes in 2b2689e. I know this comes with other drawbacks...

> `-Wunused-function` warnings are triggered although we add `-Wno-unused-function`?!

The culprit was `moveit_build_options()` which [declares its own set of warnings](https://github.com/ros-planning/moveit/blob/96e425dddcdb4834bd0980d28e0a3a27f3f5eafb/moveit_core/cmake/moveit.cmake#L18-L26) on top of ours. Fixed via 5a30f8a.

PR #389, opened 2022-11-08
Remove task size in task_id

VideoSystemsTech:
For dynamic tasks (where the stages can change between builds) the task_id changes as well.
This causes the GUI to show a different task.
Fix by removing the task size in the task_id string.


rhaschke (commented):
This doesn't add the task size, but actually, the task pointer to get a _unique_ task id.
Thus, if you change your task pointer, it _should_ have a different id!

rhaschke:
Closing as this will break normal usage (we cannot distinguish multiple tasks running in the same process).
Please describe your problem in an issue. Probably the underlying issue is something different.

PR #402, opened 2022-12-01
[ros2] Expose childByIndex to user

mechwiz:
This PR adds a way to access a child of a container by index. I specifically needed this feature to access the last stage of a container (and not via finding the child by name) but this is more generalized for all usecases.

@JafarAbdi

### DIFF
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -53,6 +53,8 @@ public:
 
 	size_t numChildren() const;
 	Stage* findChild(const std::string& name) const;
+	Stage* front() const;
+	Stage* back() const;
 
 	/** Callback function type used by traverse functions
 	 *  Receives currently visited Stage and current depth in hierarchy
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -53,6 +53,7 @@ public:
 
 	size_t numChildren() const;
 	Stage* findChild(const std::string& name) const;
+	Stage* childByIndex(int index) const;
 
 	/** Callback function type used by traverse functions
 	 *  Receives currently visited Stage and current depth in hierarchy
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -53,8 +53,6 @@ public:
 
 	size_t numChildren() const;
 	Stage* findChild(const std::string& name) const;
-	Stage* front() const;
-	Stage* back() const;
 
 	/** Callback function type used by traverse functions
 	 *  Receives currently visited Stage and current depth in hierarchy
--- a/core/src/stages/fixed_state.cpp
+++ b/core/src/stages/fixed_state.cpp
@@ -64,8 +64,6 @@ bool FixedState::canCompute() const {
 
 void FixedState::compute() {
 	SubTrajectory trajectory;
-	scene_->getCurrentStateNonConst().enforceBounds();
-	scene_->getCurrentStateNonConst().update();
 	if (!properties().get<bool>("ignore_collisions") && scene_->isStateColliding()) {
 		trajectory.markAsFailure("in collision");
 	}
--- a/core/src/stages/fixed_state.cpp
+++ b/core/src/stages/fixed_state.cpp
@@ -64,6 +64,8 @@ bool FixedState::canCompute() const {
 
 void FixedState::compute() {
 	SubTrajectory trajectory;
+	scene_->getCurrentStateNonConst().enforceBounds();
+	scene_->getCurrentStateNonConst().update();
 	if (!properties().get<bool>("ignore_collisions") && scene_->isStateColliding()) {
 		trajectory.markAsFailure("in collision");
 	}
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -342,6 +342,10 @@ Stage* ContainerBase::findChild(const std::string& name) const {
 	return nullptr;
 }
 
+Stage* ContainerBase::childByIndex(int index) const {
+	return pimpl()->childByIndex(index, false)->get();
+}
+
 bool ContainerBase::traverseChildren(const ContainerBase::StageCallback& processor) const {
 	return pimpl()->traverseStages(processor, 0, 1);
 }
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -342,14 +342,6 @@ Stage* ContainerBase::findChild(const std::string& name) const {
 	return nullptr;
 }
 
-Stage* ContainerBase::front() const {
-	return pimpl()->children().front().get();
-}
-
-Stage* ContainerBase::back() const {
-	return pimpl()->children().back().get();
-}
-
 bool ContainerBase::traverseChildren(const ContainerBase::StageCallback& processor) const {
 	return pimpl()->traverseStages(processor, 0, 1);
 }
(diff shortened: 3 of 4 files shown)


core/src/stages/fixed_state.cpp
```diff
@@ -64,6 +64,8 @@ bool FixedState::canCompute() const {
 
 void FixedState::compute() {
 	SubTrajectory trajectory;
+	scene_->getCurrentStateNonConst().enforceBounds();
```
JafarAbdi, line 67: Should we print a warning here if the user specified state violate the bounds?

core/include/moveit/task_constructor/container.h
```diff
@@ -53,6 +53,8 @@ class ContainerBase : public Stage
 
 	size_t numChildren() const;
 	Stage* findChild(const std::string& name) const;
+	Stage* front() const;
+	Stage* back() const;
```
rhaschke, line 57: Instead of providing just front and back, I suggest exposing a variant of `ContainerBasePrivate::childByIndex()`.

rhaschke (commented):
I don't agree that `CurrentState` or `FixedState` should call `enforceBounds()` themselves.
MoveIt doesn't perform this kind of normalization throughout its code base and so shouldn't MTC. Sometimes you might explicity want to plan from 190° to 170°.

core/include/moveit/task_constructor/container.h
mechwiz: Thanks for the feedback! I will make that change

mechwiz:
Hmm ok, I will need to think of another way to handle this MTC bug - I essentially want to avoid this from occuring where J1 here is the continuous joint. It starts at 9.948 rads and then goes to 3.665 rads which causes the arm to plow through collisions. when in reality the arm should stay put and not move at all. I would show a vid but its proprietary. 
```
[mtc_server_node-20] Trajectory has 21 points over 2.544 seconds
[mtc_server_node-20]   waypoint   0 time 0.000 pos  9.948 -0.260  2.094  1.047 vel -0.000  0.000  0.000  0.000 acc  0.000  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint   1 time 0.133 pos  9.879 -0.260  2.094  1.047 vel -1.049  0.000  0.000  0.000 acc -10.790  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint   2 time 0.266 pos  9.669 -0.260  2.094  1.047 vel -2.098  0.000  0.000  0.000 acc -10.609  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint   3 time 0.399 pos  9.324 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc -7.826  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint   4 time 0.532 pos  8.941 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc -0.000  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint   5 time 0.665 pos  8.557 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc -0.000  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint   6 time 0.799 pos  8.173 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint   7 time 0.932 pos  7.789 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint   8 time 1.065 pos  7.405 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint   9 time 1.198 pos  7.021 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  10 time 1.331 pos  6.637 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc -0.000  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint  11 time 1.464 pos  6.253 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  12 time 1.597 pos  5.869 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc -0.000  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint  13 time 1.730 pos  5.485 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  14 time 1.863 pos  5.101 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc -0.000  0.000  0.000  0.000 
[mtc_server_node-20]   waypoint  15 time 1.996 pos  4.717 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  16 time 2.130 pos  4.333 -0.260  2.094  1.047 vel -2.885  0.000  0.000  0.000 acc  0.000 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  17 time 2.263 pos  3.977 -0.260  2.094  1.047 vel -2.217  0.000  0.000  0.000 acc  5.395 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  18 time 2.396 pos  3.752 -0.260  2.094  1.047 vel -1.168  0.000  0.000  0.000 acc  8.368 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  19 time 2.529 pos  3.666 -0.260  2.094  1.047 vel -0.119  0.000  0.000  0.000 acc  8.259 -0.000 -0.000 -0.000 
[mtc_server_node-20]   waypoint  20 time 2.544 pos  3.665 -0.260  2.094  1.047 vel  0.000 -0.000 -0.000 -0.000 acc  0.939 -0.000 -0.000 -0.000
```

rhaschke:
You could add your own WrapperStage to cleanup the PlanningScene's RobotState as you like.
Why, in first place, do you feed different RobotStates, if you don't actually want to move from 9.948 to 3.665 rads but to 9.948?

mechwiz:
I think what is happening is that as the continuous joint makes a few wraps as the robot moves, the angle in this case became 9.948 rads which is reflected in the current state stage. I then feed into MTC that I want to go to a named goal in a MoveTo stage which has the angle for the continuous joint at 3.665 rads - and then this issue occurs.

mechwiz:
So I did some digging and it looks like Moveit actually does do this kind of normalization for continuous joints via the FixedStartStateBounds adapter right here https://github.com/ros-planning/moveit2/blob/main/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp#L98 

From what I can tell though, this pipeline is only executed if using the pipeline planner in MTC (and you of course specify to use this adapter which I think is recommended). However, if using cartesian_path or joint_interpolation then this issue will occur

mechwiz:
OK, I updated this PR to just target exposing childByIndex to the user. I tested this within my project and it seems to work!

core/include/moveit/task_constructor/container.h
mechwiz: change made in most recent commit

PR #415, opened 2023-01-04
Fix call of computeGeneric

jspricke:
computeGeneric is templated and private so it will be undefined in the
shared object and when using the computeForward/Backward functions.

### DIFF
--- a/core/include/moveit/task_constructor/stage.h
+++ b/core/include/moveit/task_constructor/stage.h
@@ -291,8 +291,8 @@ public:
 
 	// Default implementations, using generic compute().
 	// Override if you want to use different code for FORWARD and BACKWARD directions.
-	virtual void computeForward(const InterfaceState& from) { computeGeneric<Interface::FORWARD>(from); }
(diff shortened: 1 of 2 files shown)


rhaschke:
Could you add a comment in the PR description, please, why the implementation should be in the .cpp file?
Do you get multiple definition errors?

rhaschke (approved):

jspricke:
added, sorry.

jspricke:
An other option would be to move the `computeGeneric` function to the header, if you prefer that.

rhaschke:
> added, sorry.

Where? I'm still curious why this is necessary.

rhaschke:
Merged into `master`: 770a8a0b170d07c549252f4370db6b65366c2d57. Will be considered with next forward port.

jspricke:
Thx, can I get this into the ros2 branch soon to fix the moveit2 tutorials?

rhaschke:
See #416

PR #420, opened 2023-01-28
[ROS2]: Sync with master branch

JafarAbdi:
Sync with master branch to cut a ros2 MTC release

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,7 +38,7 @@ jobs:
 
     env:
       CLANG_TIDY_ARGS: -quiet -export-fixes ${{ github.workspace }}/.work/clang-tidy-fixes.yaml
-      DOCKER_IMAGE: ghcr.io/ros-planning/moveit2:${{ matrix.env.IMAGE }}
+      DOCKER_IMAGE: moveit/moveit2:${{ matrix.env.IMAGE }}
       UNDERLAY: /root/ws_moveit/install
       # TODO: Port to ROS2
       # DOWNSTREAM_WORKSPACE: "github:ubi-agni/mtc_demos#master github:TAMS-Group/mtc_pour#master"
--- a/core/src/stages/generate_pose.cpp
+++ b/core/src/stages/generate_pose.cpp
@@ -72,7 +72,9 @@ void GeneratePose::compute() {
 	if (upstream_solutions_.empty())
 		return;
 
-	planning_scene::PlanningScenePtr scene = upstream_solutions_.pop()->end()->scene()->diff();
+	const SolutionBase& s = *upstream_solutions_.pop();
+	planning_scene::PlanningSceneConstPtr scene = s.end()->scene()->diff();
+
 	geometry_msgs::msg::PoseStamped target_pose = properties().get<geometry_msgs::msg::PoseStamped>("pose");
 	if (target_pose.header.frame_id.empty())
 		target_pose.header.frame_id = scene->getPlanningFrame();
@@ -82,6 +84,7 @@ void GeneratePose::compute() {
 	}
 
 	InterfaceState state(scene);
+	forwardProperties(*s.end(), state);  // forward registered properties from received solution
 	state.properties().set("target_pose", target_pose);
 
 	SubTrajectory trajectory;


rhaschke (approved):

rhaschke:
Handling docker images via `ghcr.io` is definitely broken! Is anybody actively working on this?
Did @v4hn agree to release MTC? He was always hesitant in the past...

v4hn:
No, I did not agree, but I also don't see any point in disagreeing anymore and this did not respond yet.

- Picknik uses it in multiple projects and aims to roll from releases
- after that much time the current state is a quasi-standard either way. If we want to rework the API to add important features or eventually fix some basic concerns, it is clearly better to have a release before.

That being said I don't know right now  whether there are important bugs that definitely need fixing before a release.

The other question is how we want to handle release versions between ROS and ROS2.

JafarAbdi:
> Handling docker images via `ghcr.io` is definitely broken! Is anybody actively working on this?

If I recall correctly, someone from PickNik said that they will be working on fixing it in the last maintainers meeting 

> The other question is how we want to handle release versions between ROS and ROS2.

My plan was to start from `2.0.0` for the ros2 branch, and hope we never reach that version in the master branch, or I could start from `0.2.0` since the master version is `0.0.0`. I'm happy with any solution that we agree on

rhaschke:
> The other question is how we want to handle release versions between ROS1 and ROS2.

I also suggest to use from 1.x for ROS1 and 2.x for ROS2.
Another idea: Could we use - in principle - the same version numbers for ROS1 and ROS2? This would reflect the actual state the best - code-wise ROS1 and ROS2 don't differ! To have different tags on both branches, we would need to prefix the version number in the tag like rosx.0.1.0.

JafarAbdi:
> I also suggest to use from 1.x for ROS1 and 2.x for ROS2. Another idea: Could we use - in principle - the same version numbers for ROS1 and ROS2? This would reflect the actual state the best - code-wise ROS1 and ROS2 don't differ! To have different tags on both branches, we would need to prefix the version number in the tag like rosx.0.1.0.

I like the second proposal, I just tested it using `ros2-x.x.x` and it works without any problem. I just had to write it in the `Release Tag` prompt and use `x.x.x` for the `Version` prompt. We just have to make sure to manually name the tag and push it every release

v4hn:
> I also suggest to use from 1.x for ROS1 and 2.x for ROS2.
> Another idea: Could we use - in principle - the same version numbers for ROS1 and ROS2? This would reflect the actual state the best - code-wise ROS1 and ROS2 don't differ! To have different tags on both branches, we would need to prefix the version number in the tag like rosx.0.1.0.

Isn't that the case with various ros/ros2 branches still? ;)
We probably want to keep the option to have releases in only one of the branches, so I would not perfectly synchronize them. `2.0.0` for ROS2 and `1.0.0` for ROS sounds sufficient to me.

rhaschke:
> We probably want to keep the option to have releases in only one of the branches, so I would not perfectly synchronize them. 

The proposed scheme doesn't require to release in sync. However, it ensures that release numbers actually indicate progress. A release `2.0` for ROS2 suggests being far more advanced than a release `1.0` for ROS1, although they probably will be based on the same source code (+/- ROS1<->ROS2 changes).

> `2.0.0` for ROS2 and `1.0.0` for ROS sounds sufficient to me.

It's misleading and we will run into trouble if we really want to bump the major version number - independent from ROS1/ROS2...

v4hn:
> I also suggest to use from 1.x for ROS1 and 2.x for ROS2.

You suggested it @rhaschke 😅. 
My migraine with keeping the same version number for both frameworks is that we are bound to desync the releases to some degree due to ROS/ROS2-only quirks. Then we will see version 1.3.4 in one of the branches only, but 1.3.5 in both again.

But I share your scepticism on abusing the major version number (as other ros modules did) as well.
If you prefer the prefix, I'm also more happy with it. Please use `ros2-1.0.0` though (note the dash instead of a period).
It would be easier with a "ros-agnostic" core with its own version, but that's just not realistic with the right moveit-dependency we have.

rhaschke:
I merged these changes in a different order and with cleaned commit messages in 293898690e323176c0f5c2fbe223affa53bde462
and I prepared releases for Noetic (https://github.com/ros/rosdistro/pull/36032) and ROS2 (tag created).
In order to release into ROS2 (Rolling and Humble?), I'm waiting for permissions (https://github.com/ros2-gbp/ros2-gbp-github-org/pull/191).

PR #423, opened 2023-02-06
pybind11 dependency

v4hn:
Hey Robert,

what is the state of [the pybind11 dependency](https://github.com/ros-planning/moveit_task_constructor/tree/master/core/python) you added?
Debian testing just moved to Python 3.11 and the pinned commit fails with the new version.

I was wondering whether we could move to `pybind11-dev`. Updating the pin to the current `smart_holder` branch on your remote also fixes my concrete problem. I don't know enough about which features you use/need to decide the best resolution. Do we still need your remote at all or was that just used when you played around with unique_ptr integration?


rhaschke:
Hi Michael,
I wasn't aware that I'm using my own fork of pybind11 here. It think, that's not needed anymore. So you can easily use the original pybind11 repo. However, we do need the `smart_holder` branch as this one enables `unique_ptr` handling.

rhaschke:
As you see from the failing CI job, latest pybind11 doesn't support Python 2.x anymore (as required for Melodic)...

v4hn:
> (as required for Melodic)

Looks like we will have a hard time going forward with support for newer python3 versions if we want to keep python2 compatibility. How open are you to drop melodic/python2 support from the `master` branch?
You just tagged the 0.1.0 release, so I think it's feasible and maintaining yet another branch in mtc just because of a different pin for the pybind11 subdirectory is quite drastic overhead.
For now, I'll just change the pin in my local directory as I'm not aware of anyone else being affected yet.

rhaschke:
Sorry for not replying here yet. The current pybind11 commit is the latest still working with python2 (see ced37a029cd1a5128992dd9d6075677f532eb6dd of recently merged #412). To forward to python 3.11, I suggest opening a new branch, [`drop-python2`](https://github.com/ubi-agni/moveit_task_constructor/tree/drop-python2),  to collect just those changes.

By the way, which OS do you use? Ubuntu 22.04 uses 3.10, doesn't it?

v4hn:
> To forward to python 3.11, I suggest opening a new branch, [`drop-python2`](https://github.com/ubi-agni/moveit_task_constructor/tree/drop-python2),  to collect just those changes.

To the best of my knowledge there is only one change necessary so far and I don't expect any others w.r.t. python any time soon: forwarding the pybind11 submodule to the current `smart_holder` branch (currently 8ece7d641ca6ce316e59fec6744b8517073bbe32). That's why I think it's empty overhead to keep an additional branch rebased on master just for this.

> By the way, which OS do you use? Ubuntu 22.04 uses 3.10, doesn't it?

I built this on Debian testing, i.e. bookworm. The state there will be included in Ubuntu 24.04 eventually and a few users pull in python3.11 via anaconda already. I don't think this very important to address upstream *yet*.
However, going forward, I would like to push for a decision to drop python2 (and thus ROS melodic) support from MoveIt/MTCs development branches to avoid getting stuck on old versions at the cost of breakage with newer versions.

PR #428, opened 2023-02-17
ComputeIK: Limit collision checking to JMG

rhaschke:
I was wondering, why ComputeIK reported collisions of static scene objects, while MoveIt happily ignores those.
The reason was that ComputeIK was doing collision checking for the full, unrestricted scene.
Now, it's limited to the JMG as well.
Maybe, MoveIt needs to be fixed to include downstream links of the JMG as well: If the JMG moves, those links will move as well... That could be the reason for end-effector collisions reported by users from time to time. However, this needs further inspection. I didn't have time for this (yet).


v4hn (approved):

v4hn:
Makes sense. I might have ran into similar issues as well in the past. Usually when you find it, it also means there is another bug somewhere in your project. :-)

PR #429, opened 2023-02-17
MultiPlanner

rhaschke:
This introduces `MultiPlanner`, a new meta planner implementing the `PlannerInterface`, that runs multiple alternative planners in sequence and returns the first found solution.
This is useful to sequence different planning strategies of increasing complexity, e.g. Cartesian or joint-space interpolation first, then OMPL, ...
This is (slightly) different from the Fallbacks container, as the MultiPlanner directly applies its planners to each individual planning job. In contrast, the Fallbacks container first runs the active child to exhaustion before switching to the next child, which possibly applies a different planning strategy.

As the children planners should be able to use different timeouts (than the stage employing `MultiPlanner`), I also introduced a `timeout` property for all planners.

Doing so, I noticed and fixed a bug in the `timeout` handling of `JointInterpolation`.

### DIFF
--- a/core/include/moveit/task_constructor/solvers/cartesian_path.h
+++ b/core/include/moveit/task_constructor/solvers/cartesian_path.h
@@ -56,8 +56,10 @@ public:
 	void setJumpThreshold(double jump_threshold) { setProperty("jump_threshold", jump_threshold); }
 	void setMinFraction(double min_fraction) { setProperty("min_fraction", min_fraction); }
 
-	void setMaxVelocityScaling(double factor) { setProperty("max_velocity_scaling_factor", factor); }
-	void setMaxAccelerationScaling(double factor) { setProperty("max_acceleration_scaling_factor", factor); }
+	[[deprecated("Replace with setMaxVelocityScalingFactor")]]  // clang-format off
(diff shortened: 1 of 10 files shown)


v4hn:
Sorry for leaving this specifically hanging. There's clearly no harm in supporting the new mechanism in MTC now that it's upstream, but I still think this whole project should actually be handled as a variation of the Fallback container within the MTC framework instead of having its own opaque implementation as a single planner.
However, to get the same behavior rather big changes are required in MTC (including parallel stage computations).

The other fixes/cleanups make sense too.

rhaschke:
As discussed in the initial comment, I don't see how this feature could be implemented as a Fallback container.

v4hn:
Not without support for running multiple children at the same time, that's what I said.
I definitely do not consider it out of scope though, but clearly out of current functionality.

rhaschke:
Even if you can run multiple children in parallel, the semantics is different: Fallback containers run their children to exhaustion before switching to the next child/method, while the `MultiPlanner` switches to the next method on failure of the previous one(s) for each and every job immediately.

v4hn:
Sorry, I meant to write a variation of *ParallelContainer*, not Fallback. I am well aware of the differences.

PR #432, opened 2023-02-28
Fix Solution::fillMessage()

### LINKED ISSUE #405: Error when displaying the last waypoint of a MoveRelative stage
This seems to happen only in RViz and only on the `lower object` sub-task , executed on a real robot without problem. Similar to the demo except that there's a `ModifyPlanningScene` between the MoveRelative and the GeneratePlacePose stages. The lift up does not exhibit this behavior.

```cpp
    /****************************************************
  ---- *               Allow Collision (hand object)   *
     ***************************************************/
    {
      auto stage = std::make_unique<stages::ModifyPlanningScene>("allow collision (hand,object)");
      stage->allowCollisions("thermocycler", attached_object_names_, false);
      place->insert(std::move(stage));
    }
```

https://user-images.githubusercontent.com/32679594/205476330-0bb1ea46-89cc-40ca-a4d1-e08b3649e0cf.mp4

rhaschke:
Need to generate full planning scene message if planning progressed backwards: The scene_diff field describes PlanningScene changes of the end scene relative to the start scene, while during planning it might be inversed.
Seemingly fixes #405. I would like to add some unit tests before merging.
I also noticed that `ModifyPlanningScene::(add|remove)Object` doesn't yet work in backwards mode. Shall we support this or just issue an error? To reverse, `removeObject()` would require the full `CollisionObject` description (to actually create the object), while currently (for the forward pass) the name suffices.

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -170,11 +170,15 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 		exec_traj.controller_names_ = sub_traj.execution_info.controller_names;
 
 		/* TODO add action feedback and markers */
-		exec_traj.effect_on_success_ = [this, sub_traj,
+		exec_traj.effect_on_success_ = [this,
+		                                &scene_diff = const_cast<::moveit_msgs::PlanningScene&>(sub_traj.scene_diff),
 		                                description](const plan_execution::ExecutableMotionPlan* /*plan*/) {
-			if (!moveit::core::isEmpty(sub_traj.scene_diff)) {
+			scene_diff.robot_state.joint_state = sensor_msgs::JointState();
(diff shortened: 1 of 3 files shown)


captain-yoshi (approved):
Confirming that this fixes #405.

v4hn (commented):
> Need to generate full planning scene message if planning progressed backwards

That's a cheap solution indeed. :+1: It is quite costly with meshes in the scene, but it's definitely better than the current broken state. An improved implementation could still implement a generic "diff" for independent scene objects. Please consider creating a new issue for this as future work when merging this PR.

CI doesn't like it yet. Seems to be related to scene diff handling during execution:

>  [ INFO] [1677622610.740722501]: Found a contact between 'object' (type 'Object') and 'panda_leftfinger' (type 'Robot link'), which constitutes a collision. Contact information is not stored.
  [ INFO] [1677622610.740764900]: Collision checking is considered complete (collision was found and 0 contacts are stored)
  [ INFO] [1677622610.740791500]: Trajectory component '7/19' is invalid after scene update

PR #450, opened 2023-03-30
Enable parallel planning with PipelinePlanner

sjahr:
This PR refactors the PipelinePlanner to enable the use of MoveIt's parallel planning API. Usage would look for example like this

```cpp
  // Configure a set of pipelines to be used
  std::unordered_map<std::string, std::string> pipeline_id_planner_id_map;
  pipeline_id_planner_id_map["ompl"] = "RRTConnectkConfigDefault";
  pipeline_id_planner_id_map["pilz_industrial_motion_planner"] = "LIN";
  pipeline_id_planner_id_map["chomp"] = "chomp";

  // Construct solver
  auto parallel_planner = std::make_shared<mtc::solvers::PipelinePlanner>(node_, pipeline_id_planner_id_map);
  
  // Optional: Configure custom solution selection and/or stopping criterion
  parallel_planner->setStoppingCriterionFunction(...);
  parallel_planner->setSolutionSelectionFunction(
      [](const std::vector<::planning_interface::MotionPlanResponse>& solutions) {...});
```

You can test the PR by using [my fork of the moveit2_tutorials](https://github.com/sjahr/moveit2_tutorials/tree/update_parallel_planning_tutorial). Build the custom MTC and moveit2_tutorials branch and run in two terminal windows:
 
1. `ros2 launch moveit2_tutorials mtc_demo.launch.py`
2. `ros2 launch moveit2_tutorials pick_place_demo.launch.py`

PR #451, opened 2023-04-03
Alternatives and fallback planners

sjahr:
Depends on https://github.com/ros-planning/moveit2/pull/2043

Implement planners proposed by @rhaschke [here](https://github.com/ros-planning/moveit_task_constructor/pull/450#issuecomment-1492246817)

FallbackPlanner is a renamed version of the already existing MultiPlanner with the extensions, that it is now possible to pass a stopping criterion to the planner.

AlternativesPlanner is an implementation of the parallel planning interface in Moveit but on an MTC solver level so it is possible to run multiple solvers in parallel and select a solution. I did not find an easy way to add a stopping criteria here, so each parallel solver uses it's complete timeout for the moment.

This PR does not invalidate #450 because that is a refactoring of solely the PipelinePlanner to support MoveIt2's parallel planning feature.


rhaschke (commented):
Do you plan to have these planners in ROS1 as well? If so, I suggest implementing them there first. 
(Our standard development branch is master / ROS1 in this repo and we forward-port changes to ROS2 via merge commits.)

I don't yet like the code redundancy of the two `plan()` functions for the AlternativesPlanner. I think this calls for a template.

Which problems did you face when defining a stopping criterion for AlternativesPlanner?
I guess it's difficult to stop other planning threads? How did you solve that with  ParallelPlanning?

rhaschke:
> This PR does not invalidate #450 because that is a refactoring of solely the PipelinePlanner to support MoveIt2's parallel planning feature.

I don't agree. Why would it be beneficial to have parallel planning implemented in PipelinePlanner too, when we already have `AlternativesPlanner`? I think we should build on modularity (combine simple planners via Fallback|Alternatives) instead of providing this functionality within PipelinePlanner again.

sjahr:
Sorry for my late response here.

> Do you plan to have these planners in ROS1 as well? If so, I suggest implementing them there first.
(Our standard development branch is master / ROS1 in this repo and we forward-port changes to ROS2 via merge commits.)

I don't have the capacity at the moment to implement this in ROS 1, since I am working most of the time with ROS 2.

> Which problems did you face when defining a stopping criterion for AlternativesPlanner?
I guess it's difficult to stop other planning threads? How did you solve that with ParallelPlanning?

For parallel planning within MoveIt 2 I used the member function `terminate` of the `PlanningPipeline` to terminate the pipelines (see [here](https://github.com/ros-planning/moveit2/blob/cbd18dfe57a919fec023964d357dd745e13e219f/moveit_ros/planning/planning_pipeline_interfaces/src/planning_pipeline_interfaces.cpp#L107)). For the PipelinePlanner it would be easy to expose it in the MTC solver itself, but implementing a terminate function for the Cartesian/ Joint Interpolator would have been a little more difficult, so my plan was to split it from this PR in a separate effort.

Additionally, I am running into a segfault when I am using the AlternativesPlanner planner. So I am assuming the solvers are not thread safe at the moment. I haven't taken the time to dig more into this

sjahr:
> I don't agree. Why would it be beneficial to have parallel planning implemented in PipelinePlanner too, when we already have `AlternativesPlanner`? I think we should build on modularity (combine simple planners via Fallback|Alternatives) instead of providing this functionality within PipelinePlanner again.

While both interfaces enable parallel planning, they are doing it on different hierachy levels. The Alternative Solver enables using multiple MTC solvers in parallel and the pipeline planner (as a single MTC solver) uses MoveIt's planning interface to use one or multiple planning pipelines. Having a MoveIt parallel planning interface is useful for using the feature independent of MTC e.g. with moveit_cpp or hybrid planning.
Having parallel planning implemented in the PipelinePlanner won't reduce modularity since it does not prevent anybody from using it as a simple planner. It rather offers more liberty on which interface you'd like to use (maybe they'll diverge in the future in terms of available features e.g. at this point terminating MTC solvers is not possible).

rhaschke:
I exactly want to avoid a redundant implementation at two different levels of hierarchy, which are then prone to diverge in implementation and features. Modularity stems from _simple_ building blocks, which can be nicely _combined_ with each other.

I'm not sure how parallel planning pipelines are configured in MoveItCpp: Do you dynamically pick several pipeline configs or do you need to define those statically in advance? In the former case, I don't see any limitation in using the AlternativesPlanner.
The tutorial lists an excerpt of a `moveit_cpp.yaml` config, which (due to the indentation of the "ompl_rrtc", "pilz_lin", and "chomp" configs) suggest the latter, more restrictive approach.

sjahr:
> Do you dynamically pick several pipeline configs or do you need to define those statically in advance?

The pipelines are created and initialized in advance & when you run plan() you decide which subset you'd like to use.

> I exactly want to avoid a redundant implementation at two different levels of hierarchy,
I don't see how this can be avoided for example because `MTC solver != MoveIt Planner`. Parallel planning is a useful feature on both levels.

sjahr:
Maybe this would be a good discussion to have in the Maintainer meeting :thinking:

sjahr:
@rhaschke I'd still like to see #450 merged. I think enabling the pipeline-level parallel planning feature is useful for MTC users.
I haven't found the time to continue this PR and the alternative planner currently doesn't work due to a segfault. But I see your point about modularity & avoiding a redundant implementation. Maybe merging #450 could be a temporary solution. At the previous maintainer meeting, @henningkayser came up with a potential solution to this:

> How about moving MTC’s solver interface into MoveIt only? We are lacking a consistent interface anyway (example: Cartesian Interpolator) and the solver is not really specific to MTC at all. The only concern is supporting MTC’s planner properties which should be solvable.

Pushing the solver interface down to MoveIt would resolve the issues with redundant implementations while keeping the feature available in both, MoveIt and MTC. Do you have time this week for a video call to discuss how we can move forward?

v4hn:
> Maybe merging https://github.com/ros-planning/moveit_task_constructor/pull/450 could be a temporary solution

There are no temporary solutions in MoveIt (those are called feature branches).
Things in the main branches tend to persist and you know that. :-)
At the same time endless debate never gets anything merged :innocent: 

In contrast to Robert I would encourage #450 over #451.
The point of #450 is to forward an existing MoveIt feature into MTC, which I still think is the wrong way of handling parallelism in MTC, but it's basically "plugin what is there".
Implementing MTC planners as vectors of planners as proposed here pretty much re-implements the whole concept of MTC's parallel containers inside a single stage. But the feature-set is even worse because inspecting only solutions produced by a specific planner and setting different solution properties depending on the planner has to be reimplemented in a different way than with the concept of having one planner per stage.

> Pushing the solver interface down to MoveIt would resolve the issues with redundant implementations

The MTC solver interface is very limited in that you cannot pass arbitrary goal constraints, not even multiple poses for different links. It suffices for what we modeled in MTC so far, but I believe it is definitely too limited to describe general MoveIt interface.

rhaschke:
> The MTC solver interface is very limited in that you cannot pass arbitrary goal constraints, not even multiple poses for different links. It suffices for what we modeled in MTC so far, but I believe it is definitely too limited to describe general MoveIt interface.

I agree. The MTC solver interface is very much tailored toward MTC's needs. I will think about #450 again, now having Michael's opinion on it.

sjahr:
> There are no temporary solutions in MoveIt (those are called feature branches).
Things in the main branches tend to persist and you know that. :-)
At the same time endless debate never gets anything merged innocent

:sweat_smile: 

> The MTC solver interface is very limited in that you cannot pass arbitrary goal constraints, not even multiple poses for different links. It suffices for what we modeled in MTC so far, but I believe it is definitely too limited to describe general MoveIt interface.

Ok, that makes sense. I am imagining this interface to provide the same functionality as the solver interface does: The capability to use a solver (Joint or Cartesian Interpolation) or a planner (more specifically: A planning pipeline) for a given MotionPlanRequest but more generalized as the MTC solver interface. Within MTC this new MoveIt interface replaces the solver interface. @henningkayser Is this understanding of your idea correct and can you elaborate a bit more on the idea and maybe how it should look like?

henningkayser:
>. @henningkayser Is this understanding of your idea correct and can you elaborate a bit more on the idea and maybe how it should look like?

Yes, my idea was more along the lines of reducing the number of planning interfaces that we have across MoveIt and MTC while also filling the gap in combining the different solvers that MoveIt has (as in, MoveIt doesn't provide a common interface for Cartesian interpolation, planners, joint interpolation). MTC's interface already provides this in a limited way, so I would very much be in favor of a single solver interface that is part of MoveIt, but is also used by MTC. (I don't think that's feasible without a significant amount of extra work, though).

About parallel planning: I think Cartesian interpolation is not thread-safe by default because of the IK solver (I think kdl and bio_ik are both not thread-safe, pick_ik should be). Until there is a way to enforce thread-safety, we shouldn't enable parallel planning for this type of solver. To me, MTC's stages and containers are sufficient for now.

I would advocate for merging #450 since the MoveIt-internal version is a feature that already exists and that is under active development. I agree that there is a certain conceptual overlap between MTC's container types and this feature, however MTC's approach does not suffice configuring and optimizing specific planner setups. And that's the whole point of this feature, allowing to experiment with and optimizing groups of planners and configurations that can be used as a default in MoveIt and in MTC, without having to implement new boilerplate code around that. It is not problem-specific, it is a tool for switching to say OMPL+STOMP+PILZ for all planning requests in an application instead of only solving with RRTConnect. That's where I see the advantage of enabling MoveIt's version for MTC.

PR #456, opened 2023-05-03
Handle failed trajectory in connect stage

Abishalini:
The following change helped mitigate trajectory discontinuities in potential solutions.

### DIFF
--- a/core/src/stages/connect.cpp
+++ b/core/src/stages/connect.cpp
@@ -159,7 +159,14 @@ void Connect::compute(const InterfaceState& from, const InterfaceState& to) {
 
 		robot_trajectory::RobotTrajectoryPtr trajectory;
 		success = pair.second->plan(start, end, jmg, timeout, trajectory, path_constraints);
-		sub_trajectories.push_back(trajectory);  // include failed trajectory
+		
+		// Do not push partial solutions
+		if (success) {
+			sub_trajectories.push_back(trajectory);
+		} else {
+			// Pushing a nullptr instead of a failed trajectory.
+			sub_trajectories.push_back(nullptr);
+		}
 
 		if (!success)
 			break;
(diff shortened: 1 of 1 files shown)


rhaschke (changes requested):
Can you please explain the addressed issue in more detail? As far as I understand, the failed trajectory is considered in complete solution sequences, which is definitely a bug.

However, not pushing the failed trajectory is just a workaround. In doing so, we loose the ability to inspect the failed trajectory for debugging purposes. We should better find and fix the bug...

So, please describe the bug as detailed as possible and ideally provide test code to (reliably) reproduce it.

rhaschke:
@Abishalini: Ping (https://github.com/ros-planning/moveit_task_constructor/pull/456#pullrequestreview-1411981342)

Abishalini:
While working on a pick and place pipeline, we occasionally stumbled across discontinuous trajectories and sometimes, it's chosen as the optimal solution causing robot e-stops. 
https://user-images.githubusercontent.com/17363793/236535749-cfb14aba-77a5-47b8-9fe2-b5cb96874948.mp4

I have not been able to reproduce this with certainty and used this temporary fix. 
I am currently out of office. I will add more information in the next few days.

henningkayser:
We are currently working on a minimal example. I'm not entirely sure why MTC is picking up the failed trajectory, but the story seems to go like this:

1. CartesianPlanner fails and returns an incomplete (but not null) [result](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/solvers/cartesian_path.cpp#L104)
2. Connect [receives trajectory and success result](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L161) and [stores the failed trajectory](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L162)
3. It also [stores the trajectory as subsolution because it's not null](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L198) and [connects it to the end scene](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L210) 
4. It marks the [SolutionSequence](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#LL177C2-L177C29) as failure (success is false)

... and the failed trajectory still somehow makes it into the accepted solution with finite cost.

rhaschke:
Thanks for the summary. Not sure, the bug is related to that. Having a minimal example will be definitely helpful!

PR #459, opened 2023-05-04
Rework execution

rhaschke:
This is a draft I had lying around for a while to rework MTC solution execution. Instead of calling the move_group's capability, the idea is to directly interface MoveIt's `TrajectoryExecutionManager` and thus allow callbacks for each sub-trajectory segment. To this end, this PR reimplements MoveIt's `PlanExecution` to:
- allow for both start and end callbacks for each segment in an `ExecutableMotionPlan` (the [original PlanExecution only allowed for end effects](https://github.com/ros-planning/moveit/blob/8635f82e4823b970479f360216bbe8292f79e8e4/moveit_ros/planning/plan_execution/include/moveit/plan_execution/plan_representation.h#L68))
  Of course, in principle it would suffice to have start _or_ end effects only. However, I found it simpler to define preparatory and cleanup tasks of a specific segment within that segment instead of shifting one or the other to the previous/next segment.
- to run a sequence of related trajectories without risking interference with another caller

The latter functionality was also implemented by https://github.com/ros-planning/moveit/pull/3243 meanwhile.

The old Task::execute() method was moved into a free function `execute()`, still calling the move_group capability.
However, a new free function `executableMotionPlan(SolutionBase)` is provided to create the `ExecutableMotionPlan` for use with the new `PlanExecution`.

This needs to be consolidated with https://github.com/ros-planning/moveit/pull/3243. 
@cambel and @v4hn: Can we organize a video conference to discuss open TODOs?

### DIFF
--- a/core/include/moveit/task_constructor/execution.h
+++ b/core/include/moveit/task_constructor/execution.h
@@ -35,6 +35,7 @@
 /* Authors: Robert Haschke */
 
 #pragma once
+#include <list>
 
 #include <actionlib/client/simple_action_client.h>
 #include <moveit/planning_scene_monitor/planning_scene_monitor.h>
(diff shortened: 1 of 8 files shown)


cambel:
@rhaschke sure

v4hn:
I wasn't aware you had that lying around. Cool!

I'm open for a call on TODOs, but from my perspective the main points are
- we need thorough reviews for https://github.com/ros-planning/moveit/pull/3243 so that it can be confidently merged to replace the current execution. I can't put priorities on things I don't need in my own work for a while already and I'm quite ashamed I did not get around to do it in all this time.
- MTC execution came up quite often during the GSoC meetings with @cambel as the implemented approach is right in the middle of PlanExecution, TrajectoryExecution, and MTC execution capabilities. I would hope/expect that this draft can be based on the new execution system and from your explanations the main thing missing are the start callbacks in @cambel's system (which I agree make a lot of sense to add).

PR #461, opened 2023-05-15
Patches and fixes for ros2 branch

sjahr:
This PR contains a couple of fixes and improvements by @JafarAbdi that we've been using on a fork for quite some time but never got contributed upstream. With his consent, I am making that up now with this PR.

If desired I can divide this PR into smaller ones but I wanted to hear your feedback first.

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -94,8 +94,10 @@ void ExecuteTaskSolutionCapability::initialize() {
 	                                             std::placeholders::_1, std::placeholders::_2)),
 	    ActionServerType::CancelCallback(
 	        std::bind(&ExecuteTaskSolutionCapability::preemptCallback, this, std::placeholders::_1)),
-	    ActionServerType::AcceptedCallback(
-	        std::bind(&ExecuteTaskSolutionCapability::goalCallback, this, std::placeholders::_1)));
+	    [this](const std::shared_ptr<rclcpp_action::ServerGoalHandle<ExecuteTaskSolutionAction>> goal_handle) {
+		    last_goal_future_ =
+		        std::async(std::launch::async, &ExecuteTaskSolutionCapability::goalCallback, this, goal_handle);
+	    });
 }
 
 void ExecuteTaskSolutionCapability::goalCallback(
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -94,7 +94,7 @@ void ExecuteTaskSolutionCapability::initialize() {
 	                                             std::placeholders::_1, std::placeholders::_2)),
 	    ActionServerType::CancelCallback(
 	        std::bind(&ExecuteTaskSolutionCapability::preemptCallback, this, std::placeholders::_1)),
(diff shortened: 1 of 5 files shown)


sea-bass (approved):

sea-bass (commented):
I think a lot of the member variables recently switched from having underscores to not having underscores, which is causing build failures. Take a look at that?

core/src/stage.cpp
```diff
@@ -634,6 +634,9 @@ void PropagatingEitherWay::computeGeneric(const InterfaceState& start) {
 	else
 		send<dir>(start, InterfaceState(end), std::move(trajectory));
 }
+// Explicitly instantiate templates for both directions
+template void PropagatingEitherWay::computeGeneric<Interface::FORWARD>(const InterfaceState& start);
+template void PropagatingEitherWay::computeGeneric<Interface::BACKWARD>(const InterfaceState& start);
```
v4hn, line 639: Why?
The template is private in the class and there should be no need to have a symbol for instantiations outside the object.

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -179,9 +179,10 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 			for (auto& scene_diff : scene_diffs) {
 				if (!moveit::core::isEmpty(scene_diff)) {
 					/* RCLCPP_DEBUG_STREAM(LOGGER, "apply effect of " << description); */
-					scene_diff.robot_state.joint_state = sensor_msgs::msg::JointState();
-					scene_diff.robot_state.multi_dof_joint_state = sensor_msgs::msg::MultiDOFJointState();
-					if (!context_->planning_scene_monitor_->newPlanningSceneMessage(scene_diff))
+					auto scene_diff_copy = scene_diff;
+					scene_diff_copy.robot_state.joint_state = sensor_msgs::msg::JointState();
+					scene_diff_copy.robot_state.multi_dof_joint_state = sensor_msgs::msg::MultiDOFJointState();
+					if (!context_->planning_scene_monitor_->newPlanningSceneMessage(scene_diff_copy))
```
v4hn, line 185: This is a fixup for https://github.com/ros-planning/moveit_task_constructor/pull/461/commits/72afdd22d2a941350294c33fae3cccaf585f5b44 , please merge the commits.
Also, you should set `robot_state.is_diff = true` explicitly if you empty the robot_state fields and do it *before* you check for `isEmpty` because the set robot state will often be the only reason why the scene diff is not empty.

Could you split this feature into a separate PR? It's fully independent of the tiny other cleanups and might break something in the future.

v4hn (commented):
Thank you! Makes me wonder why these patches didn't make it here over the last year...

capabilities/src/execute_task_solution_capability.cpp
sjahr: > This is a fixup for https://github.com/ros-planning/moveit_task_constructor/commit/72afdd22d2a941350294c33fae3cccaf585f5b44 , please merge the commits.

@v4hn Which commits are you referring to?

core/src/stage.cpp
sjahr: :thinking: It build without this too, so I removed it for now. @JafarAbdi Do you know if this is necessary?

JafarAbdi (approved):

PR #464, opened 2023-05-18
Add Constraints to ComputeIK

VideoSystemsTech:
This adds Constraints to the ComputeIK stage, similar to what is done in MoveIt's PositionIKRequest service.
This way we can filter out IK solutions before they are discarded anyway during planning.

### DIFF
--- a/core/src/stages/compute_ik.cpp
+++ b/core/src/stages/compute_ik.cpp
@@ -49,6 +49,18 @@
 #include <iterator>
 #include <ros/console.h>
 
+namespace {
+bool isIKSolutionValid(const planning_scene::PlanningScene* planning_scene,
+                       const kinematic_constraints::KinematicConstraintSet* constraint_set,
+                       moveit::core::RobotState* state, const moveit::core::JointModelGroup* jmg,
+                       const double* ik_solution) {
+	state->setJointGroupPositions(jmg, ik_solution);
+	state->update();
+	return (!planning_scene || !planning_scene->isStateColliding(*state, jmg->getName())) &&
+	       (!constraint_set || constraint_set->decide(*state).satisfied);
+}
+}  // namespace
+
 namespace moveit {
 namespace task_constructor {
 namespace stages {
@@ -62,6 +74,9 @@ ComputeIK::ComputeIK(const std::string& name, Stage::pointer&& child) : WrapperB
 	p.declare<bool>("ignore_collisions", false);
 	p.declare<double>("min_solution_distance", 0.1,
 	                  "minimum distance between seperate IK solutions for the same target");
+	p.declare<moveit_msgs::Constraints>(
+	    "constraints", moveit_msgs::Constraints(),
+	    "A set of constraints that the IK must obey; by default, this set of constraints is empty");
 
 	// ik_frame and target_pose are read from the interface
 	p.declare<geometry_msgs::PoseStamped>("ik_frame", "frame to be moved towards goal pose");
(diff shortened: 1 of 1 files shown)


core/src/stages/compute_ik.cpp
```diff
@@ -49,6 +49,18 @@
 #include <iterator>
 #include <ros/console.h>
 
+namespace {
+bool isIKSolutionValid(const planning_scene::PlanningScene* planning_scene,
+                       const kinematic_constraints::KinematicConstraintSet* constraint_set,
+                       moveit::core::RobotState* state, const moveit::core::JointModelGroup* jmg,
+                       const double* ik_solution) {
+	state->setJointGroupPositions(jmg, ik_solution);
```
rhaschke, line 57: This is redundant to line 384.

core/src/stages/compute_ik.cpp
```diff
@@ -62,6 +74,9 @@ ComputeIK::ComputeIK(const std::string& name, Stage::pointer&& child) : WrapperB
 	p.declare<bool>("ignore_collisions", false);
 	p.declare<double>("min_solution_distance", 0.1,
 	                  "minimum distance between seperate IK solutions for the same target");
+	p.declare<moveit_msgs::Constraints>(
+	    "constraints", moveit_msgs::Constraints(),
+	    "A set of constraints that the IK must obey; by default, this set of constraints is empty");
```
rhaschke, line 79:
```suggestion
	    "constraints", moveit_msgs::Constraints(), "additional constraints to obey");
```

core/src/stages/compute_ik.cpp
```diff
@@ -369,13 +385,19 @@ void ComputeIK::compute() {
 		ik_solutions.emplace_back();
 		auto& solution{ ik_solutions.back() };
 		state->copyJointGroupPositions(jmg, solution.joint_positions);
+
+		kinematic_constraints::KinematicConstraintSet kset(robot_model);
+		kset.add(constraints, scene->getTransforms());
+		auto kset_ptr = kset.empty() ? nullptr : &kset;
```
rhaschke, line 391: The definition of the KinematicConstraintSet should happen _once_ outside the lambda.

core/src/stages/compute_ik.cpp
```diff
@@ -49,6 +49,18 @@
...
 #include <ros/console.h>
 
+namespace {
+bool isIKSolutionValid(const planning_scene::PlanningScene* planning_scene,
+                       const kinematic_constraints::KinematicConstraintSet* constraint_set,
+                       moveit::core::RobotState* state, const moveit::core::JointModelGroup* jmg,
+                       const double* ik_solution) {
+	state->setJointGroupPositions(jmg, ik_solution);
+	state->update();
+	return (!planning_scene || !planning_scene->isStateColliding(*state, jmg->getName())) &&
```
rhaschke, line 59: Collision checking is redundant to line 399.

rhaschke (commented):
Thanks for this suggestion. I'm not fully convinced though. IMHO it doesn't make sense to define additional Cartesian constraints - all constraints are already fully defined by the Cartesian goal pose. Joint constraints though might be useful.
Rejected solutions should include the reason in the comments, distinguishing between collisions and constraints (as well as naming the failed constraints).

core/src/stages/compute_ik.cpp
VideoSystemsTech: Thanks

core/src/stages/compute_ik.cpp
VideoSystemsTech: Thanks

VideoSystemsTech:
> Thanks for this suggestion. I'm not fully convinced though. IMHO it doesn't make sense to define additional Cartesian constraints - all constraints are already fully defined by the Cartesian goal pose. Joint constraints though might be useful. Rejected solutions should include the reason in the comments, distinguishing between collisions and constraints (as well as naming the failed constraints).

I agree, Position and Orientation constraints shouldn't be needed, but I also think they might be useful as a further check of constraints. For example, planning for a pose which is outside global position/orientation constraints.

BTW, I cannot find a way to list which constraints were satisfied or not.

Issue #353, opened 2022-04-14
Using ExecuteTaskSolutionCapability to execute a solution is overriding the joint state values


JafarAbdi:
This is happening because in [execute_task_solution_capability.cpp#L179](https://github.com/ros-planning/moveit_task_constructor/blob/master/capabilities/src/execute_task_solution_capability.cpp#L179) it modifies the state in the planning scene monitor which overrides the values from `/joint_states`, `/tf`, `/tf_static` topics

Possible fixes could be

1- Reset the robot_state's joint_state and multi_dof_joint_state for scene_diff before applying the planning scene
```cpp
auto scene_diff = sub_traj.scene_diff; // Not ideal since we will copy the whole scene_diff before reseting joint_state & multi_dof_joint_state
scene_diff.robot_state.joint_state = sensor_msgs::msg::JointState();
scene_diff.robot_state.multi_dof_joint_state = sensor_msgs::msg::MultiDOFJointState();
```
2- In [task.cpp#L268-L269](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/task.cpp#L268-L269) after we fill up the message we reset those variables

Let me know which solution you prefer so I open a PR with the patch

JafarAbdi:
@v4hn @rhaschke do you any preferences?

rhaschke:
I guess the differences between actual joint state values (after execution) and those in `scene_diff` are marginal, right?
I would prefer the second option: the planning scene diff shouldn't comprise joint values in first place.
Sorry for the delay in replying.

sjahr:
I think this is addressed with #498. Please re-open if not.

mechwiz:
While #498 addresses the issue in option 2, that is only true if the execute function is called within MTC. However, if a goal is sent to the execution_task_solution server directly separate from within the execute function in MTC, then the joint values in that goal's trajectories will still override the joint state values. Hence I think option 1 is really the way to go to address this issue.

rhaschke:
The problem with #498 is simply that the cleanup of the scene_diff's joint states is only performed in Task::execute, but not in the message generation. I filed PR #504, which should solve that issue.

PR #467, opened 2023-05-23
Make mtc execute solution capability apply scene diffs

sjahr:
This PR contains a fix for #353 by @JafarAbdi that we've been using on a fork for quite some time but never got contributed upstream. With his consent, I opened this PR to bring it upstream

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -143,51 +143,73 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 	}
 
 	plan.plan_components.reserve(solution.sub_trajectory.size());
-	for (size_t i = 0; i < solution.sub_trajectory.size(); ++i) {
-		const moveit_task_constructor_msgs::msg::SubTrajectory& sub_traj = solution.sub_trajectory[i];
+	auto make_executable_trajectory =
+	    [&](const moveit_task_constructor_msgs::msg::SubTrajectory& sub_traj, const std::string& description,
+	        const std::function<bool(const plan_execution::ExecutableMotionPlan*)>& on_success_callback) {
+		    plan.plan_components.emplace_back();
+		    plan_execution::ExecutableTrajectory& exec_traj = plan.plan_components.back();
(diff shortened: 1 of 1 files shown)


capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -142,52 +142,75 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
...
+
+	auto make_apply_planning_scene_diff_cb = [this](const std::vector<moveit_msgs::msg::PlanningScene>& scene_diffs) {
+		return
+		    [this, scene_diffs = std::move(scene_diffs)](const plan_execution::ExecutableMotionPlan* /*plan*/) mutable {
+			    for (auto& scene_diff : scene_diffs) {
+				    if (!moveit::core::isEmpty(scene_diff)) {
+					    /* RCLCPP_DEBUG_STREAM(LOGGER, "apply effect of " << description); */
+					    scene_diff.robot_state.joint_state = sensor_msgs::msg::JointState();
+					    scene_diff.robot_state.multi_dof_joint_state = sensor_msgs::msg::MultiDOFJointState();
+					    if (!context_->planning_scene_monitor_->newPlanningSceneMessage(scene_diff))
```
sjahr, line 184: Comment of @v4hn:

> This is a fixup for https://github.com/ros-planning/moveit_task_constructor/commit/72afdd22d2a941350294c33fae3cccaf585f5b44 , please merge the commits.
Also, you should set robot_state.is_diff = true explicitly if you empty the robot_state fields and do it before you check for isEmpty because the set robot state will often be the only reason why the scene diff is not empty.

rhaschke (commented):
The changes are rather large and I don't see immediately what was actually changed.
Please add some more details to the main comment of this PR describing the actual problem and your solution for it.

rhaschke:
Given my feedback [here](https://github.com/ros-planning/moveit_task_constructor/issues/353#issuecomment-1140467392), I would have expected a much simpler solution, namely removing the corresponding RobotState members.

sjahr:
Closed in favor of #498

sjahr:
Closed in favor of #498

PR #470, opened 2023-05-26
Sync master into ros2 branch

rhaschke:


rhaschke:
I'm surprised to see `humble-source` failing due to [missing underscores in `MotionPlanResponse` members](https://github.com/ros-planning/moveit2/blob/b882b0b2c269a192cf72857b84b86a27f7bdfee8/moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h#L54).
Isn't Humble the current active ROS2 release and thus should comprise the corresponding [`main` branch changes](https://github.com/ros-planning/moveit2/blob/4ed043fcfbf127a479448d82b4543729dca63160/moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h#L59) from https://github.com/ros-planning/moveit2/pull/1939?

henningkayser:
@rhaschke thank you for working on this sync. At this point, `ros2` doesn't support the MoveIt `humble` branch anymore. However, there is a moveit_task_constructor `humble` branch as well that should work (haven't tested lately). mtc:ros2 moveit2:main + ROS 2 Humle should still work, though.

rhaschke:
@henningkayser, indeed, retriggering the builds it worked now. Don't know what the problem was before...

PR #474, opened 2023-07-03
Expose MultiPlanner to Python

VideoSystemsTech:
Exposes MultiPlanner (#429) to Python APIs.

### DIFF
--- a/core/python/bindings/src/solvers.cpp
+++ b/core/python/bindings/src/solvers.cpp
@@ -38,6 +38,7 @@
 #include <moveit/task_constructor/solvers/joint_interpolation.h>
 #include <moveit/task_constructor/solvers/multi_planner.h>
 #include <moveit_msgs/WorkspaceParameters.h>
+#include "utils.h"
 
 namespace py = pybind11;
 using namespace py::literals;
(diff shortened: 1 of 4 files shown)


core/python/bindings/src/solvers.cpp
```diff
@@ -47,6 +48,27 @@ PYBIND11_SMART_HOLDER_TYPE_CASTERS(PlannerInterface)
...
+}
+
+// implement operator[](index)
+template <typename T>
+typename T::value_type get_item(const T& container, long index) {
+	auto it = container.begin();
+	std::advance(it, normalize_index(container.size(), index));
+	return *it;
+}
+}  // anonymous namespace
```
rhaschke, line 71: Instead of replicating this code from `core.cpp`, I suggest to factor it out into a separate header file, e.g. `python_utils.h`.

core/python/bindings/src/solvers.cpp
```diff
@@ -116,6 +138,27 @@ void export_solvers(py::module& m) {
...
+		)")
+	    .def("__len__", &MultiPlanner::size)
+	    .def("__getitem__", &get_item<MultiPlanner>)
+	    .def(
+	        "add",
+	        [](MultiPlanner& self, const py::args& args) {
+		        for (auto it = args.begin(), end = args.end(); it != end; ++it)
+			        self.push_back(it->cast<PlannerInterfacePtr>());
+	        },
+	        "Insert a planner")
```
rhaschke, line 158:
```suggestion
	        "Insert one or more planners")
```

demo/scripts/multi_planner.py
```diff
@@ -0,0 +1,69 @@
...
+
+roscpp_initialize("mtc_tutorial_multi_planner")
+
+ompl_pipelinePlanner = core.PipelinePlanner("ompl")
+ompl_pipelinePlanner.planner = "RRTConnectkConfigDefault"
+pilz_pipelinePlanner = core.PipelinePlanner("pilz_industrial_motion_planner")
+pilz_pipelinePlanner.planner = "PTP"
+multiPlanner = core.MultiPlanner()
+multiPlanner.add(ompl_pipelinePlanner)
+multiPlanner.add(pilz_pipelinePlanner)
```
rhaschke, line 16: Probably you want to switch the order of these planners: the PTP planner that is prone to fail should come first.

rhaschke (commented):
Thanks for this contribution and sorry for the late reply. I have a few comments:

PR #476, opened 2023-07-07
Mark incomplete trajectories as failure

Abishalini:
Sometimes, the plan function in the `Connect` stage returns success with an incomplete trajectory. 
This predominantly happens when using OMPL's RRT planner. Here is a band-aid fix to mark such trajectories as failure while I dig deeper to find the source of this error.


Abishalini:
Made this PR to the wrong fork. Closing this

PR #490, opened 2023-10-06
Add planner name to trajectory info

sjahr:
Port of https://github.com/PickNikRobotics/moveit_task_constructor/pull/5

Make planner id part of the trajectory info to add information about the planner that created a trajectory

### DIFF
--- a/core/include/moveit/task_constructor/storage.h
+++ b/core/include/moveit/task_constructor/storage.h
@@ -275,8 +275,8 @@ class SolutionBase
 public:
 	virtual ~SolutionBase() = default;
 
-	inline const InterfaceState* start() const { return start_; }
-	inline const InterfaceState* end() const { return end_; }
+	[[nodiscard]] inline const InterfaceState* start() const { return start_; }
+	[[nodiscard]] inline const InterfaceState* end() const { return end_; }
 
 	/** Set the solution's start_state_
 	 *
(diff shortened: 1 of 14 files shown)


sea-bass (approved):

core/include/moveit/task_constructor/storage.h
```diff
@@ -275,8 +275,8 @@ class SolutionBase
 public:
 	virtual ~SolutionBase() = default;
 
-	inline const InterfaceState* start() const { return start_; }
-	inline const InterfaceState* end() const { return end_; }
+	[[nodiscard]] inline const InterfaceState* start() const { return start_; }
+	[[nodiscard]] inline const InterfaceState* end() const { return end_; }
```
rhaschke, line 279: @sjahr: Why should we want the `nodiscard` flag here? Yes, it's stupid to not use the result of a getter function. But there is no problem in discarding the return value.
Apart from that, this change is completely unrelated to this PR.

sjahr: My intention with these changes was to let the compiler prevent people from doing something stupid. We've started a while ago in moveit2 to improve the code quality by following [the best practices recommendations of Jason Turner](https://github.com/ros-planning/moveit2/issues/840) and using `[nodiscard]` liberally was one of his suggestions in the book. If someone wants to discard the return value of the getter function intentionally they can use std::ignore.

 I've added that as a small improvement here and did not see much value in opening a separate PR for that. If the reviewers would have requested it, I might have done that but I think for these small and non-critical changes it is easier to just clean them up along the way working on something bigger.

PR #492, opened 2023-10-10
disable hooks for unused rviz-based task construction

v4hn:
This implementation is far from functional right now and the button confuses users.

@rhaschke I don't know whether you want to keep the whole underlying framework (which is quite sophisticated) around, so I just disabled the UI hook.
I believe you also moved on to specifying tasks in python instead of trying to set them up visually.


rhaschke:
To answer your remaining questions: I didn't yet give up on building the construction interface, but I didn't find time...
The Python API is just another programming API, but not intended as a replacement for the GUI :wink:

PR #496, opened 2023-10-19
Run goalcallback asynchronously

sjahr:
Finally address [this comment](https://github.com/ros-planning/moveit_task_constructor/pull/461#issuecomment-1563080267) in #461.

Disclaimer: I am not the original author of these changes. 
This changes the behavior of the execute_task_solution_capability to not accept new execution requests if another request is currently processed. I assume the reason for this is to avoid interruptions because of incorrect execution commands. It is still possible to abort the execution with the CancelCallback.

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -94,8 +94,10 @@ void ExecuteTaskSolutionCapability::initialize() {
 	                                             std::placeholders::_1, std::placeholders::_2)),
 	    ActionServerType::CancelCallback(
 	        std::bind(&ExecuteTaskSolutionCapability::preemptCallback, this, std::placeholders::_1)),
-	    ActionServerType::AcceptedCallback(
-	        std::bind(&ExecuteTaskSolutionCapability::goalCallback, this, std::placeholders::_1)));
+	    [this](const std::shared_ptr<rclcpp_action::ServerGoalHandle<ExecuteTaskSolutionAction>> goal_handle) {
+		    last_goal_future_ =
+		        std::async(std::launch::async, &ExecuteTaskSolutionCapability::goalCallback, this, goal_handle);
+	    });
 }
 
 void ExecuteTaskSolutionCapability::goalCallback(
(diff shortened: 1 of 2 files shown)


capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -94,8 +94,10 @@ void ExecuteTaskSolutionCapability::initialize() {
 	                                             std::placeholders::_1, std::placeholders::_2)),
 	    ActionServerType::CancelCallback(
 	        std::bind(&ExecuteTaskSolutionCapability::preemptCallback, this, std::placeholders::_1)),
```
sea-bass, line 96: For consistency, can you convert the goal and cancel callbacks to use lambdas instead of `std::bind` as well?

sea-bass (approved):

sea-bass:
A little hard to read, but it's right on the edge of "these lambdas should call a member function" and "it's short enough to all be inline". I'm OK with this.

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -90,12 +90,21 @@ void ExecuteTaskSolutionCapability::initialize() {
...
 	    context_->moveit_cpp_->getNode(), "execute_task_solution",
-	    ActionServerType::GoalCallback(std::bind(&ExecuteTaskSolutionCapability::handleNewGoal, this,
-	                                             std::placeholders::_1, std::placeholders::_2)),
-	    ActionServerType::CancelCallback(
-	        std::bind(&ExecuteTaskSolutionCapability::preemptCallback, this, std::placeholders::_1)),
-	    ActionServerType::AcceptedCallback(
-	        std::bind(&ExecuteTaskSolutionCapability::goalCallback, this, std::placeholders::_1)));
+	    [this](const rclcpp_action::GoalUUID& /*uuid*/,
+	           const ExecuteTaskSolutionAction::Goal::ConstSharedPtr& /*goal*/) {
+		    if (last_goal_future_.valid() &&
```
sjahr, line 95:
```suggestion
                    // Reject new goal if another goal is currently processed
		    if (last_goal_future_.valid() &&
```

MarqRazz (approved):

PR #498, opened 2023-10-23
Reset joint values in scene diff for execution

sjahr:
Implements solution 2 of #353 and should solve it


core/src/task.cpp
```diff
@@ -281,6 +281,12 @@ moveit::core::MoveItErrorCode Task::execute(const SolutionBase& s) {
 	moveit_task_constructor_msgs::action::ExecuteTaskSolution::Goal goal;
 	s.toMsg(goal.solution, pimpl()->introspection_.get());
 
+	// Reset joint values in scene diff
```
MarqRazz, line 284: Not required but I prefer to tell other developers why we need to reset the values.

sjahr, line 284:
```suggestion
	// Reset joint values in scene diff to prevent any execution capability from
	// modifying the planning scene joint values
```

PR #499, opened 2023-10-26
Simplify formatting code with https://github.com/fmtlib

rhaschke:


rhaschke:
Bionic doesn't ship `libfmt` yet... :disappointed:

PR #502, opened 2023-11-02
Revert implicit trajectory info property inheritance

sjahr:
Addresses https://github.com/ros-planning/moveit_task_constructor/pull/355/files#r1379822616

@rhaschke If I want to inherit every property from the parent except the trajectory info, is this how it would be done:
```
  stage->properties().configureInitFrom(moveit::task_constructor::Stage::PARENT);
  stage->properties().set("trajectory_execution_info", custom_trajectory_info);
```

### DIFF
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -365,10 +365,6 @@ void ContainerBase::insert(Stage::pointer&& stage, int before) {
 	if (!stage)
 		throw std::runtime_error(name() + ": received invalid stage pointer");
 
-	if (stage->trajectoryExecutionInfo().controller_names.empty()) {
-		stage->setTrajectoryExecutionInfo(this->trajectoryExecutionInfo());
-	}
-
 	StagePrivate* impl = stage->pimpl();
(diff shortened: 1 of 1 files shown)


rhaschke (approved):

rhaschke:
Yes, explicitly setting properties has precedence over inheritance.

rhaschke:
Looks like the whole PR introduced a segfault in the unit tests.

sjahr:
@rhaschke Fortunately, it looks like #355 was not the culprit. @tylerjw was able to reproduce it locally with moveit2. The bug got introduced by https://github.com/ros-planning/moveit2/pull/2482 where we started using child loggers. Child loggers use `rcutils_array_list_remove(..)` which causes the issue [take a look at the traceback here](https://github.com/ros-planning/moveit_task_constructor/actions/runs/6731269933/job/18330877384?pr=502#step:4:1739). There's a PR open to fix this upstream in rcutils: https://github.com/ros2/rcutils/pull/434

PR #503, opened 2023-11-03
[DON'T MERGE] Test if #355 causes CI failure

sjahr:
This reverts commit 94d65143be3d9d9cdb6c564cf77e25c6ce62f6cb.


rhaschke:
Looks like #355 is the culprit!

sjahr:
Mhm, I cannot reproduce it locally and it seems like running the CI job on the ROS 2 branch manually again led to success: https://github.com/ros-planning/moveit_task_constructor/actions/runs/6745669013

PR #505, opened 2023-11-08
[ros2] Check joint bounds when interpolating

mechwiz:
Hey all,

I noticed an issue where the joint-interpolation planner would successfully plan trajectories that exceeded a joint's limits. This PR adds in the necessary logic to ensure such planning will fail as expected

### DIFF
--- a/core/src/solvers/joint_interpolation.cpp
+++ b/core/src/solvers/joint_interpolation.cpp
@@ -77,7 +77,7 @@ bool JointInterpolationPlanner::plan(const planning_scene::PlanningSceneConstPtr
 
 	// add first point
 	result->addSuffixWayPoint(from->getCurrentState(), 0.0);
-	if (from->isStateColliding(from_state, jmg->getName()))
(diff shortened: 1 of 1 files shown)


rhaschke (approved):
Makes sense. Thanks.

mechwiz:
This is an afterthought that I had but I'm wondering if instead of causing the plan to fail - we just cap the joint value at the limit if the goal goes past it, similar to what is done here (https://github.com/ros-planning/moveit2/blob/main/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp#L237) which I think gets called within the pipeline-planner if the sampling planner is used.

rhaschke:
The code you are referring to is responsible for configuring a JointConstraint data structure from a message, i.e. fixing a config error. Joints should not be clipped during planning, because then the path doesn't follow a straight-line in joint-space anymore. 
However, out-of-bound joints can only occur when either the start or end (joint) pose are out of bounds.
In both cases, I would expect the plan to fail.

PR #506, opened 2023-11-09
Update to the more recent JumpThreshold API

marioprats:


sjahr (approved):
Reference https://github.com/ros-planning/moveit2/pull/2506

PR #510, opened 2023-11-27
Add function to Introspection class to retrieve all solutions

Abishalini:
Currently, the way to retrieve solutions is to use the `get_solution` service, where you can send one solution ID as a request and get the corresponding solution.

In this PR, I have added a function to get the entire solution ID and solution map.
This would be useful for another introspection tool we are building at PickNik.

### DIFF
--- a/core/include/moveit/task_constructor/introspection.h
+++ b/core/include/moveit/task_constructor/introspection.h
@@ -80,6 +80,10 @@ public:
 	/// fill task state message for publishing the current task state
 	moveit_task_constructor_msgs::msg::TaskStatistics&
 	fillTaskStatistics(moveit_task_constructor_msgs::msg::TaskStatistics& msg);
+
+	/// Function to get a map between solution id and solutions.
+	std::map<uint32_t, moveit_task_constructor_msgs::msg::Solution> getAllSolutions();
+
 	/// publish the current state of task
 	void publishTaskState();
 
--- a/core/src/introspection.cpp
+++ b/core/src/introspection.cpp
@@ -161,7 +161,7 @@ public:
 
 	/// mapping from stages to their id
 	std::map<const StagePrivate*, moveit_task_constructor_msgs::msg::StageStatistics::_id_type> stage_to_id_map_;
-	boost::bimap<uint32_t, const SolutionBase*> id_solution_bimap_;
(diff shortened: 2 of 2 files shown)


rhaschke (commented):
I deliberatively decided against providing a (ROS) accessor to all solutions, because the amount of data to transfer is rather big for complex tasks.
In the interactive introspection with rviz, users need to select individual solutions anyway, such that the communication overhead can be limited to a single solution.

Could you motivate, why such a function is needed?
You also didn't expose the function as a ROS service yet. How do you plan to use it?

Abishalini:
I wanted to create a database of all the MTC tasks so that we can go back to analyze previous solutions. So I need to access all the solutions in one go. I am not sure if I should instead call the `get_solution` service multiple time to get all the solutions. Also, the `get_solution` service is unique for each task. I think the task ID is appended to the service's name. In this case, I cannot use the service to get solutions for older tasks since the services `get_solution_task_id` are destroyed when a newer task is run.

Since you mentioned the solutions data can be large, I will analyze the performance in a bit.

I don't plan to use ROS service to get the task information and solutions. This is how I'm using this currently
```
// Get task description
moveit_task_constructor_msgs::msg::TaskDescription task_description;
task.introspection().fillTaskDescription(task_description);

// Get task statistics
moveit_task_constructor_msgs::msg::TaskStatistics task_statistics;
task.introspection().fillTaskStatistics(task_statistics);

// Get solution id and solutions map
const auto solutions_map = task.introspection().getAllSolutions();
```

Abishalini:
Please correct me if I'm wrong. It looks like calling `solutions()` would return the successful solutions of the overall task. 
How would I access the successful and failed solutions of each stage?

rhaschke:
> How would I access the successful and failed solutions of each stage?

Just use the method `failures()`:
https://github.com/ros-planning/moveit_task_constructor/blob/a90557279ca306fc30d39ec463af401022314b17/core/include/moveit/task_constructor/task.h#L139-L140

Note, that each sub stage provides these methods as well (to access solutions/failures of a particular stage):
https://github.com/ros-planning/moveit_task_constructor/blob/a90557279ca306fc30d39ec463af401022314b17/core/include/moveit/task_constructor/stage.h#L227-L228

PR #514, opened 2023-12-06
Print warning if no controllers are configured for trajectory execution

sjahr:
If no controllers are configured for trajectory execution of a sub-trajectory, moveit's trajectory execution manager takes care of selecting a controller. Since this might lead to unexpected behavior, I'd like to add this warning to remind users to explicitly define the controller name property in their tasks

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -188,13 +188,14 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 		exec_traj.controller_name = sub_traj.execution_info.controller_names;
 
 		/* TODO add action feedback and markers */
-		exec_traj.effect_on_success = [this, sub_traj, exec_traj](const plan_execution::ExecutableMotionPlan* /*plan*/) {
-			if (!moveit::core::isEmpty(sub_traj.scene_diff)) {
-				RCLCPP_DEBUG_STREAM(LOGGER, "apply effect of " << exec_traj.description);
-				return context_->planning_scene_monitor_->newPlanningSceneMessage(sub_traj.scene_diff);
-			}
-			return true;
-		};
+		exec_traj.effect_on_success =
+		    [this, sub_traj, description = exec_traj.description](const plan_execution::ExecutableMotionPlan* /*plan*/) {
+			    if (!moveit::core::isEmpty(sub_traj.scene_diff)) {
+				    RCLCPP_DEBUG_STREAM(LOGGER, "apply effect of " << description);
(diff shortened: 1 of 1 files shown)


capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -180,21 +178,28 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 		}
 		exec_traj.trajectory = std::make_shared<robot_trajectory::RobotTrajectory>(model, group);
 		exec_traj.trajectory->setRobotTrajectoryMsg(state, sub_traj.trajectory);
+		// Check that sub trajectories that contain a valid trajectory has controllers configured
```
sea-bass, line 181:
```suggestion
		// Check that sub trajectories that contain a valid trajectory have controllers configured.

File: capabilities/src/execute_task_solution_capability.cpp [hunk 2]
```diff
@@ -180,21 +178,28 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
...
+		if (!sub_traj.trajectory.joint_trajectory.points.empty() && sub_traj.execution_info.controller_names.empty()) {
+			RCLCPP_WARN(LOGGER,
+			            "The trajectory of stage '%i' from task '%s' does not have any controllers specified for "
+			            "execution. This might lead to weird behavior.",
```
sea-bass, line 185: Definitely be more specific here.

```suggestion
			            "trajectory execution. This might lead to unexpected controller selection.",

File: capabilities/src/execute_task_solution_capability.cpp [hunk 3]
```diff
@@ -180,21 +178,28 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
...
+			            "The trajectory of stage '%i' from task '%s' does not have any controllers specified for "
+			            "trajectory execution. This might lead to unexpected controller selection.",
+			            sub_traj.info.stage_id, solution.task_id.c_str());
+		}
 		exec_traj.controller_name = sub_traj.execution_info.controller_names;
 
 		/* TODO add action feedback and markers */
-		exec_traj.effect_on_success = [this, sub_traj,
-		                               description](const plan_execution::ExecutableMotionPlan* /*plan*/) {
+		exec_traj.effect_on_success = [this, sub_traj, exec_traj](const plan_execution::ExecutableMotionPlan* /*plan*/) {
```
sjahr, line 191:
```suggestion
		exec_traj.effect_on_success = [this, sub_traj, description=exec_traj.description](const plan_execution::ExecutableMotionPlan* /*plan*/) {
```

PR #525, opened 2024-01-18
Propagate errors from planners.

Abishalini:
Sometimes, it is hard to follow why a planner failed for stages `Connect`, `MoveTo`, and `MoveRelative`.

Here is an effort to propagate the error messages from the planners. I have incorporated the use of `tl::expected` so that the plan function can now return a bool success or an error string. The stages which use planners have been modified to add the error string to the solutions. 

Example - 
image
Before this PR, the comment would be empty and I couldn't see any errors on console too which would help debug.


core/include/moveit/task_constructor/solvers/pipeline_planner.h
```diff
@@ -126,7 +126,7 @@ class PipelinePlanner : public PlannerInterface
 	 * \param [in] path_constraints Path contraints for the planning problem
 	 * \return true If the solver found a successful solution for the given planning problem
 	 */
-	bool plan(const planning_scene::PlanningSceneConstPtr& from, const planning_scene::PlanningSceneConstPtr& to,
+	tl::expected<bool, std::string> plan(const planning_scene::PlanningSceneConstPtr& from, const planning_scene::PlanningSceneConstPtr& to,
```
sjahr, line 129: The `\return` comment needs to be updated (applies to all changes of `plan()`

core/src/solvers/joint_interpolation.cpp
```diff
@@ -140,12 +140,12 @@ bool JointInterpolationPlanner::plan(const planning_scene::PlanningSceneConstPtr
 		// TODO(v4hn): planners need a way to add feedback to failing plans
 		// in case of an invalid solution feedback should include unwanted collisions or violated constraints
 		RCLCPP_WARN(LOGGER, "IK failed for pose target");
-		return false;
```
sjahr, line 143: :point_up: I guess this todo can be removed? Once you're done with this PR? Offering the option to return an error string counts as feedback I guess

core/src/stages/connect.cpp
```diff
@@ -148,6 +148,7 @@ void Connect::compute(const InterfaceState& from, const InterfaceState& to) {
 	intermediate_scenes.push_back(start);
 
 	bool success = false;
+	std::string error_message = "";
```
sjahr, line 151: Remove?

core/src/stages/connect.cpp
```diff
@@ -176,7 +184,7 @@ void Connect::compute(const InterfaceState& from, const InterfaceState& to) {
 	if (!solution)  // success == false or merging failed: store sequentially
 		solution = makeSequential(trajectory_planner_vector, intermediate_scenes, from, to);
 	if (!success)  // error during sequential planning
-		solution->markAsFailure();
+		solution->markAsFailure("Hardcode an error message");
```
sjahr, line 187: Can you add a comment to explain this hard coded failure msg?

core/src/stages/move_relative.cpp
```diff
@@ -195,10 +195,18 @@ bool MoveRelative::compute(const InterfaceState& state, planning_scene::Planning
...
 	if (getJointStateFromOffset(direction, dir, jmg, scene->getCurrentStateNonConst())) {
 		// plan to joint-space target
-		success = planner_->plan(state.scene(), scene, jmg, timeout, robot_trajectory, path_constraints);
+		const auto planner_solution_maybe =
+		    planner_->plan(state.scene(), scene, jmg, timeout, robot_trajectory, path_constraints);
+		if (planner_solution_maybe.has_value()) {
+			success = planner_solution_maybe.value();
+		}
+		if (!success) {
+			error_message = planner_solution_maybe.error();
```
sjahr, line 208: Does this work? Doesn't the `.error()` operator only return a value if `.has_value == true`? I guess this section would be simpler if plan() doesn't return a boolean

core/src/stages/move_relative.cpp
```diff
@@ -286,8 +294,14 @@ bool MoveRelative::compute(const InterfaceState& state, planning_scene::Planning
...
 		const Eigen::Isometry3d& offset = scene->getCurrentState().getGlobalLinkTransform(link).inverse() * ik_pose_world;
 
-		success =
+		const auto planner_solution_maybe =
 		    planner_->plan(state.scene(), *link, offset, target_eigen, jmg, timeout, robot_trajectory, path_constraints);
+		if (planner_solution_maybe.has_value()) {
+			success = planner_solution_maybe.value();
+		}
+		if (!success) {
+			error_message = planner_solution_maybe.error();
```
sjahr, line 303: Where does the error message come from? Isn't this overwriting a previous error message?

sjahr (commented):
Thank you! This change will help a lot with debug-ability!

While reading the code I was wondering if it make sense to keep the boolean return type if we always return true or unexpected? Maybe I am missing something but what about changing this to either:
return tl::expected<void, std::string> or return MoveItErrorCode?

core/src/stages/connect.cpp
Abishalini: I was just testing something and forgot to remove this.

core/src/stages/move_relative.cpp
Abishalini: I have modified the plan function to return a MoveItErrorCode so now I check for the error type

core/package.xml
```diff
@@ -22,6 +22,7 @@
 	<depend>rclcpp</depend>
 	<depend>rviz_marker_tools</depend>
 	<depend>tf2_eigen</depend>
+	<depend>tl_expected</depend>
```
sjahr, line 25: Remove?

core/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ find_package(moveit_task_constructor_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rviz_marker_tools REQUIRED)
 find_package(tf2_eigen REQUIRED)
+find_package(tl_expected REQUIRED)
```
sjahr, line 17: Remove?

core/CMakeLists.txt
```diff
@@ -44,5 +45,6 @@ ament_export_dependencies(moveit_task_constructor_msgs)
 ament_export_dependencies(rclcpp)
 ament_export_dependencies(rviz_marker_tools)
 ament_export_dependencies(tf2_eigen)
+ament_export_dependencies(tl_expected)
```
sjahr, line 48: Remove?

sjahr (approved):
You'll need to apply some small cleanups but otherwise this is good to go in my opinion :+1:

DaniGarciaLopez:
Thanks a lot @Abishalini for this feature! Is there any chance it can be backported to humble?

Abishalini:
@DaniGarciaLopez If you are building from source, the main branch of MoveIt2 and ros2 branch of MTC are compatible with humble. If not, feel free to cherry-pick these changes and make a PR to the humble branch.

DaniGarciaLopez:
> @DaniGarciaLopez If you are building from source, the main branch of MoveIt2 and ros2 branch of MTC are compatible with humble. If not, feel free to cherry-pick these changes and make a PR to the humble branch.

I actually compiled MoveIt and MTC from source, but encountered numerous errors related to MoveItErrorCodes because I didn't realize I was using moveit_msgs from humble binaries. After compiling moveit_msgs from the ros2 branch, MoveIt from the main, and MTC from ros2, it is now working fine. Thanks!

Issue #485, opened 2023-09-12
Discontinuity in trajectory


leander2189:
I am building a complex movment comprised of several stages. In one of them, a connecting stage using RRT sampling planner, a discontinuty appears, and the movement is taken as correct

https://github.com/ros-planning/moveit_task_constructor/assets/95743148/e4637737-23ec-4d62-a96d-76976d26cc19

I suspect that the issue might not be with MTC itself, but rather with the sampling planner configuration.

Can you point me in which direction I should look to debug this issue?

That stage is build using this:
```cpp
// Sampling planner
sampling_planner_ = std::make_shared<moveit::task_constructor::solvers::PipelinePlanner>(node_);
sampling_planner_->setProperty("goal_joint_tolerance", 1e-5);

auto stage = std::make_unique<mtc::stages::Connect>(
            name, mtc::stages::Connect::GroupPlannerVector{{"complete_chain", sampling_planner_}});

stage->setTimeout(timeout);
stage->properties().configureInitFrom(Stage::PARENT);
```

I am also getting some logs that I suspect may be related, but don't know which config parameters I should tweak
```bash
[operation_solver-14] [ERROR 1694508450.076982062] [moveit.ros_planning.planning_pipeline]: Computed path is not valid. Invalid states at index locations: [ 51 ] out of 59. Explanations follow in command line. Contacts are published on /display_contacts (generatePlan() at ./planning_pipeline/src/planning_pipeline.cpp:329)
[operation_solver-14] [INFO 1694508450.077185331] [moveit_collision_detection_fcl.collision_common]: Found a contact between 'work_object' (type 'Object') and 'ee_hilok' (type 'Robot attached'), which constitutes a collision. Contact information is not stored. (collisionCallback() at ./collision_detection_fcl/src/collision_common.cpp:319)
[operation_solver-14] [INFO 1694508450.077203612] [moveit_collision_detection_fcl.collision_common]: Collision checking is considered complete (collision was found and 0 contacts are stored) (collisionCallback() at ./collision_detection_fcl/src/collision_common.cpp:350)
[operation_solver-14] [ERROR 1694508450.077629306] [moveit.ros_planning.planning_pipeline]: Completed listing of explanations for invalid states. (generatePlan() at ./planning_pipeline/src/planning_pipeline.cpp:357)
[...]

PR #532, opened 2024-02-21
End-state deviates from goal state

### LINKED ISSUE #531: Connect stage: incompatible end and start state marked as valid
This might be related to #405.

On some of my task solutions, I get incompatible end and start robot state that manages to be valid. I can clearly see the discontinuous joints from one stage to the other. I can replicate the bug quite easily on my setup. Latest MTC commit [a905572](https://github.com/ros-planning/moveit_task_constructor/commit/a90557279ca306fc30d39ec463af401022314b17) with MoveIt [8f39ef7](https://github.com/ros-planning/moveit/commit/8f39ef78bcae5940f5aad097c3c2fd4a4667b224) (same error on a 3 month old commit).

When executing the solution through the `move_group/ExecuteTaskSolutionCapability`, you can clearly see that something is wrong between stages with the discontinuous joint positions. This is shown below.

Just another though, shouldn't the execution stop because of the large position error between subtrajectory msgs? Maybe it sends the full trajectory in one go.

https://github.com/ros-planning/moveit_task_constructor/assets/32679594/3a089556-c2d7-4098-9d11-533283601d29

Discontinuous joint positions from one stage to the other.
image

captain-yoshi:
This fixes #531 but may not fix the actual problem, e.g. it may be the planner that silently fails or something else in the framework.

I just validate the last trajectory waypoint with the goal in the `ConnectStage`. If there is a deviation, the solution is marked as failed and a comment is set. I'm pretty sure you can optimize some of the code ;)

---

On a side note, I had to make sure that the joints where compared correctly (-pi = pi). Currently, when comparing joint position in the [Connect::compatible](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/src/stages/connect.cpp#L124-L125) method, this is not checked. You might reject some valid solutions, e.g the IK solver gives some joint positions in a negative defined angle versus the RobotState who may have a positive defined angle.

Solution 1 et 3 are valid but not 2 and 4.

image

### DIFF
--- a/core/src/stages/connect.cpp
+++ b/core/src/stages/connect.cpp
@@ -43,8 +43,6 @@
 #include <moveit/planning_scene/planning_scene.h>
 #include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
 
-#include <angles/angles.h>
-
 using namespace trajectory_processing;
 
 namespace moveit {
@@ -57,8 +55,7 @@ Connect::Connect(const std::string& name, const GroupPlannerVector& planners) : 
 
 	auto& p = properties();
 	p.declare<MergeMode>("merge_mode", WAYPOINTS, "merge mode");
-	p.declare<double>("max_joint_deviation", 1e-4,
-	                  "maximum allowed joint position difference between end-state and goal-sate ");
+	p.declare<double>("max_distance", 1e-4, "maximally accepted distance between end and goal sate");
 	p.declare<moveit_msgs::Constraints>("path_constraints", moveit_msgs::Constraints(),
 	                                    "constraints to maintain during trajectory");
 	properties().declare<TimeParameterizationPtr>("merge_time_parameterization",
@@ -136,57 +133,11 @@ bool Connect::compatible(const InterfaceState& from_state, const InterfaceState&
 	return true;
 }
 
-bool Connect::validateEndTrajectoryDeviation(const moveit::core::JointModelGroup* jmg,
-                                             const robot_trajectory::RobotTrajectoryPtr trajectory,
-                                             const moveit::core::RobotState& goal_state, double max_joint_deviation,
-                                             std::string& comment) {
-	const std::size_t waypoint_count = trajectory->getWayPointCount();
-
-	if (!waypoint_count) {
-		comment = "Cannot validate end trajectory deviation with an empty trajectory";
-		return false;
-	}
-
-	for (const moveit::core::JointModel* jm : jmg->getJointModels()) {
-		const unsigned int num = jm->getVariableCount();
-
-		if (!num)
-			continue;
-
-		Eigen::Map<const Eigen::VectorXd> positions_goal(goal_state.getJointPositions(jm), num);
-		Eigen::Map<const Eigen::VectorXd> positions_last_waypoint(
-		    trajectory->getWayPoint(waypoint_count - 1).getJointPositions(jm), num);
-
-		double min_distance;
-		for (std::size_t i = 0; i < num; ++i) {
-			if (jm->getType() == robot_state::JointModel::REVOLUTE)
-				min_distance = angles::shortest_angular_distance(positions_last_waypoint[i], positions_goal[i]);
-			else
-				min_distance = positions_last_waypoint[i] - positions_goal[i];
-
-			ROS_DEBUG_STREAM_NAMED("Connect",
-			                       "angular deviation: " << min_distance << " between trajectory last waypoint and goal.");
-
-			if (std::abs(min_distance) > max_joint_deviation) {
-				std::stringstream ss;
-				ss << "Deviation in joint " << jm->getName() << ": [" << positions_last_waypoint.transpose() << "] != ["
-				   << positions_goal.transpose() << "]";
-				comment = ss.str();
-				ROS_ERROR_STREAM_NAMED("Connect", comment);
-
-				return false;
-			}
-		}
-	}
-
-	return true;
-}
-
 void Connect::compute(const InterfaceState& from, const InterfaceState& to) {
 	const auto& props = properties();
 	double timeout = this->timeout();
 	MergeMode mode = props.get<MergeMode>("merge_mode");
-	double max_joint_deviation = props.get<double>("max_joint_deviation");
+	double max_distance = props.get<double>("max_distance");
 	const auto& path_constraints = props.get<moveit_msgs::Constraints>("path_constraints");
 
 	const moveit::core::RobotState& final_goal_state = to.scene()->getCurrentState();
@@ -197,7 +148,7 @@ void Connect::compute(const InterfaceState& from, const InterfaceState& to) {
 	intermediate_scenes.push_back(start);
 
 	bool success = false;
-	std::string deviation_comment;
+	std::string comment;
 	std::vector<double> positions;
 	for (const GroupPlannerVector::value_type& pair : planner_) {
 		// set intermediate goal state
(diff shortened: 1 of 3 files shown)


core/src/stages/connect.cpp
```diff
@@ -132,6 +134,51 @@ bool Connect::compatible(const InterfaceState& from_state, const InterfaceState&
...
+		Eigen::Map<const Eigen::VectorXd> positions_goal(goal_state.getJointPositions(jm), num);
+		Eigen::Map<const Eigen::VectorXd> positions_last_waypoint(
+		    trajectory->getWayPoint(waypoint_count - 1).getJointPositions(jm), num);
+
+		double min_distance;
+		for (std::size_t i = 0; i < num; ++i) {
+			if (jm->getType() == robot_state::JointModel::REVOLUTE)
+				min_distance = angles::shortest_angular_distance(positions_last_waypoint[i], positions_goal[i]);
+			else
+				min_distance = positions_last_waypoint[i], positions_goal[i];
```
rhaschke, line 162: I guess you want to compute the joint difference:
```suggestion
				min_distance = std::abs(positions_last_waypoint[i] -  positions_goal[i]);
```

core/src/stages/connect.cpp
```diff
@@ -132,6 +134,51 @@ bool Connect::compatible(const InterfaceState& from_state, const InterfaceState&
...
+
+			ROS_DEBUG_STREAM_NAMED("Connect",
+			                       "angular deviation: " << min_distance << " between trajectory last waypoint and goal.");
+
+			if (std::abs(min_distance) > 1e-4) {
```
rhaschke, line 167: The magic number 1e-4 should be configurable, e.g. via a stage property.

rhaschke (commented):
Reading your solution to #531, I think that this is not related to incompatible start and end states: What you check is that the end-state actually reaches the goal state.
I think, your issue is a duplicate of #485, which was caused by a bug in MoveIt2.
Are you using MoveIt1 or MoveIt2? If you were using MoveIt1, it would be great if you could setup a failing unittest for your issue. Then, I can search for the underlying problem instead of applying cosmetic corrections :wink:

captain-yoshi:
> What you check is that the end-state actually reaches the goal state.

Yes ! Changed the title.

> This fixes https://github.com/ros-planning/moveit_task_constructor/issues/531 but may not fix the actual problem, e.g. it may be the planner that silently fails or something else in the framework.

Just found the culprit. Even though LIN Pilz receives a JointConstraint, it will compute the [FK](https://github.com/ros-planning/moveit/blob/master/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp#L100) to compute the goal pose. Currently the LIN planner only plans from a [cartesian goal](https://github.com/ros-planning/moveit/blob/master/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp#L147). Will create an issue in MoveIt.

Not tested but I'm pretty sure I could recreate the problem by changing the demo pick and place and replacing this [line](https://github.com/ros-planning/moveit_task_constructor/blob/master/demo/src/pick_place_task.cpp#L219) with the LIN Pilz planner. Or would you prefer a unit-test in MoveIt ?

For testing this PR, is there a way to fake a planner response that would offset the goal state outside the threshold limits ?

captain-yoshi:
To replicate this error, just use this mtc fork [captain-yoshi/moveit_task_constructor/tree/trigger-error-532](https://github.com/captain-yoshi/moveit_task_constructor/tree/trigger-error-532) and start the pick & place demo. This includes the validation in this PR.

You should see that it's failing and that for some solution the end-state does not match the goal-state.

image

captain-yoshi:
After reading more on the [Pilz LIN](https://ros-planning.github.io/moveit_tutorials/doc/pilz_industrial_motion_planner/pilz_industrial_motion_planner.html#the-lin-motion-command) planner, it generates linear Cartesian trajectory between goal and start poses.

The solution would be to feed the joint constraint goal instead of redoing an IK.

rhaschke:
Yes, the Pilz planner performs linear interpolation in Cartesian space. If you want to perform linear interpolation in joint space, there is the `JointInterpolationPlanner` in MTC available for this.

> Not tested but I'm pretty sure I could recreate the problem by changing the demo pick and place and replacing this [line](https://github.com/ros-planning/moveit_task_constructor/blob/master/demo/src/pick_place_task.cpp#L219) with the LIN Pilz planner. Or would you prefer a unit-test in MoveIt ?

So you claim, that the Pilz planner generates incomplete trajectories, which are considered to be valid by MTC?
Please create a unit test in MTC triggering exactly this issue. Thanks.
 
> For testing this PR, is there a way to fake a planner response that would offset the goal state outside the threshold limits?

Why would you like to do so? If a planner behaves like that, this planner is buggy and should get fixed.

core/src/stages/connect.cpp
```diff
@@ -132,10 +136,57 @@ bool Connect::compatible(const InterfaceState& from_state, const InterfaceState&
...
+			continue;
+
+		Eigen::Map<const Eigen::VectorXd> positions_goal(goal_state.getJointPositions(jm), num);
+		Eigen::Map<const Eigen::VectorXd> positions_last_waypoint(
+		    trajectory->getWayPoint(waypoint_count - 1).getJointPositions(jm), num);
+
+		double min_distance;
+		for (std::size_t i = 0; i < num; ++i) {
+			if (jm->getType() == robot_state::JointModel::REVOLUTE)
+				min_distance = angles::shortest_angular_distance(positions_last_waypoint[i], positions_goal[i]);
```
rhaschke, line 163: You must not use `angles::shortest_angular_distance` here, as we need to reach the _exact_ goal state - subsequent stages rely on this fact!

rhaschke (changes requested):
I see the need for this PR. Some planners, like Pilz or MTC's Cartesian interpolation, may not reach the exact goal state, although the Cartesian pose matches.
See the comment below to address before merging.

captain-yoshi:
> I see the need for this PR. Some planners, like Pilz or MTC's Cartesian interpolation, may not reach the exact goal state, although the Cartesian poses match.

Exactly ! This is necessary to be able to use such planners. Furthermore, it can catch errors in other planners, e.g. #485. Finally, it guards against the `move_group/ExecuteTaskSolutionCapability` which seems to shoot the hole trajectory without doing any validation against the end state and the start state of sub-trajectories.

core/src/stages/connect.cpp
captain-yoshi: Hum... I'm pretty sure it removed valid solutions... Will investigate more and report back soon.

rhaschke:
> it guards against the `move_group/ExecuteTaskSolutionCapability` which seems to shoot the hole trajectory without doing any validation against the end state and the start state of sub-trajectories.

The discontinuity doesn't happen between the end-state and start-state of to consecutive trajectories but between the next-to-last and last state of the Connect trajectories. Thus, MoveIt's task execution cannot guard against this discontinuity.

rhaschke:
I pushed a simplifying commit, using `RobotState::distance()` to compute the joint-space distance between the two states. No need to reimplement the wheel ;-)

captain-yoshi:
All good on my end :) Great optimization by the way :1st_place_medal:

core/src/stages/connect.cpp
rhaschke: We cannot accept equality up to multiples to 360°. The goal state needs to be reached exactly. Otherwise subsequent stages will run into trouble because they relied on a different start state.

PR #538, opened 2024-03-06
[WIP] Move Relative with Pilz does not reject trajectory with collision

captain-yoshi:
It is my understanding that the `MoveRelative` stage min/max distance cannot be used with Pilz Industrial Motion Planner in it's default behaviour:

The min/max distance option relies on an early stop on collisions to work. Unlike MoveIt Cartesian Interpolator, the Pilz does not check for collision by default as pointed out by ros-planning/moveit/issues/2942. This results in the trajectory still containing waypoints that are in collision.

This could lead to potential collisions. Should the default behaviour of Pilz align more with the "early stop on collisions" ? If not, we should at least document this behavior in MTC. I don't think there is fullproof solution for this usecase...

I added 2 tests which demonstrates the issue and valid behaviour:
- `cartesianCollisionMinMaxDistance`: Set's a collision at the end of the trajectory. The plan will succeed because the minimum distance is cleared. Early stop on collisions. This test should be added in this package.
- `cartesianPilzCollisionMinMaxDistance`: Plan should have NOT succeeded because waypoints are in collision. No early stop on collisions (Default). Just for demonstration.

---

I will probably submit a PR to be able to modify the default behaviour for Pilz in MoveIt. Would these 3 modes make sense ?
1. Don't stop on collisions (default current behaviour)
2. Early stop on self-collisions
3. Early stop on collisions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,10 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         env:
-          - IMAGE: master-source
-            TARGET_CMAKE_ARGS: >-
-              -DCMAKE_BUILD_TYPE=Release
-              -DCMAKE_CXX_FLAGS="-Werror -Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls"
           - IMAGE: noetic-source
             NAME: ccov
             TARGET_CMAKE_ARGS: -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="--coverage"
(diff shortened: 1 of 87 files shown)


captain-yoshi:
Even if Pilz would stop early on collisions, it would not work because the time parameterization is done on the hole trajectory in a `PipelinePlanner`, unlike the `CartesianPath` solver...

rhaschke (approved):
This is a nice new unit test illustrating the issues with the Pilz planner.

In order to resolve the issue, we need to truncate the trajectory returned by the (Pilz) pipeline planner to the valid part.
https://github.com/ros-planning/moveit/pull/3577 suggests to do that in MoveIt's planning pipeline. 
@v4hn: I would like your opinion, on which way to go.

rhaschke:
> Even if Pilz would stop early on collisions, it would not work because the time parameterization is done on the hole trajectory in the PipelinePlanner, unlike the CartesianPath solver...

Obviously, truncation of the trajectory needs to happen before time parameterization.
Having thought more about the issue, I am convinced that the issue should be tackled in the Pilz planner: it shouldn't generate invalid trajectory waypoints in first place, but stop on the last valid waypoint. Of course, this requires (additional) validity checks. On the other hand, we skip some IK computations in case the trajectories becomes invalid...

v4hn:
Thanks for documenting the issue @captain-yoshi !

> I am convinced that the issue should be tackled in the Pilz planner: it shouldn't generate invalid trajectory waypoints in first place, but stop on the last valid waypoint.

I fully agree.

> Of course, this requires (additional) validity checks.

Well, the pilz planner does perform any validity checks as far as I know, so they should be added either way.
We could keep them optional (with the default ON).

PR #541, opened 2024-03-12
Correctly handle failures of the pipeline planner

rhaschke:
The boolean return value of `generatePlan()` does not indicate the validity of the plan. 
Instead, we need to check `res.error_code_`.


sjahr (approved):
Makes sense. In which case does `generatePlan` need to return a boolean value at all?

rhaschke:
> In which case does `generatePlan` need to return a boolean value at all?

`generatePlan()` always returns a boolean and it actually indicates that planning was solved _and_ returned a valid path.
`res.error_code_` just provides a more detailed error code. Closing here.

Issue #540, opened 2024-03-10
ros2 branch was cleaned up and force-pushed


rhaschke:
I went through all commits pushed onto the `ros2` branch in recent months from PickNik's side and cleaned them up to the standards of this repo.
Many of them I moved to the `master` branch and eventually merged them back into the ros2 branch. 
Using merging instead of cherry-picking, one can more easily see which commits are on which branch.

**Attention: I force-pushed the ros2 branch to the new, cleaned version.
Please update your local workspaces, resetting your local `ros2` branch:** `git remote update; git reset --hard origin/ros2`.

Only a few commits, whose value I still consider questionable, are left on the ros2 branch. These include:
- #450 (I think a dedicated solver is more suitable in the context of MTC)
- #490 (this info is not displayed anywhere, integrating it into the comment makes more sense - which was realized in #523)

Some others are ros2-specific, for example:
- #496 (only relevant for the ros2 action server)
- #506

For now, I kept the old ros2 branch with name [`ros2-old`](https://github.com/ros-planning/moveit_task_constructor/tree/ros2-old). I will delete this branch in a few weeks.
The new branch should™ include all functionality of the old one as well as all improvements from the master branch.
See [here](https://github.com/ros-planning/moveit_task_constructor/compare/ros2-old..ros2) for all differences. To highlight my changes even more, I created a [cleanup branch](https://github.com/ubi-agni/moveit_task_constructor/commits/ros2-cleanup/), where individual changes can be seen more easily.
In the new ros2 branch, those cleanups are squash-merged into the corresponding main commits, though.

[Comparing PickNik's ros2 branch with the old ros2 branch](https://github.com/ros-planning/moveit_task_constructor/compare/ros2-old..PickNikRobotics:moveit_task_constructor:ros2), apart from many missing commits, I noticed larger deviations in `ExecuteTaskSolutionCapability`. So far, I don't understand, why you, @henningkayser, refactored this code so much. In any case, I think you should consider switching to the new `ros2` branch.

sea-bass:
Force pushing to a major branch and rewriting history is not ideal for repos that have a significant number of users such as this one. Especially when (not hypothetically) people may depend on specific commit hashes for their work.

IMO it would have been preferable to do a "normal" pull request with your intended improvements/cleanups and tag the authors of the disputed PRs for review so that they can learn from your feedback.

[...]

PR #542, opened 2024-03-12
Temporarily fix issues with execution of pick+place demo

rhaschke:
- Connect stage: Relax validity check of reached end state
  Looks like ROS2 uses a more relaxed goal constraint threshold than ROS1.
For this reason, all end states reached by Connect solutions of pick+place demo
are rejected. This commit relaxes the max_distance threshold of Connect as well.
**MoveIt2 maintainers should check MoveIt2's goal constraint threshold and use a similar value in Connect stage.**
Fixes https://github.com/ros-planning/moveit_task_constructor/issues/540#issuecomment-1988323674

- Revert "Fix generation of Solution msg: consider backward operation"
This reverts commit 4aac679257fb91bed9174096761879d1f14182fc, because MoveIt2 is missing a port of https://github.com/ros-planning/moveit/pull/3538.
Fixes https://github.com/ros-planning/moveit2_tutorials/issues/881

### DIFF
--- a/core/src/stages/connect.cpp
+++ b/core/src/stages/connect.cpp
@@ -57,7 +57,7 @@ Connect::Connect(const std::string& name, const GroupPlannerVector& planners) : 
 
 	auto& p = properties();
 	p.declare<MergeMode>("merge_mode", WAYPOINTS, "merge mode");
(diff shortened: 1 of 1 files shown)


sjahr:
I've opened a PR to port the necessary change to moveit2. Once that is in, you don't need to revert that commit anymore I think

PR #545, opened 2024-03-18
Discuss some force-pushed changes

sjahr:
@rhaschke I have some questions around the changes you created on the ros2 branch. I'd like to use this PR to understand them better and potentially make some updates.

### DIFF
--- a/core/include/moveit/task_constructor/solvers/cartesian_path.h
+++ b/core/include/moveit/task_constructor/solvers/cartesian_path.h
@@ -63,11 +63,11 @@ public:
 
 	void init(const moveit::core::RobotModelConstPtr& robot_model) override;
 
-	Result plan(const planning_scene::PlanningSceneConstPtr& from, const planning_scene::PlanningSceneConstPtr& to,
+	MoveItErrorCode plan(const planning_scene::PlanningSceneConstPtr& from, const planning_scene::PlanningSceneConstPtr& to,
 	            const moveit::core::JointModelGroup* jmg, double timeout, robot_trajectory::RobotTrajectoryPtr& result,
 	            const moveit_msgs::msg::Constraints& path_constraints = moveit_msgs::msg::Constraints()) override;
 
-	Result plan(const planning_scene::PlanningSceneConstPtr& from, const moveit::core::LinkModel& link,
+	MoveItErrorCode plan(const planning_scene::PlanningSceneConstPtr& from, const moveit::core::LinkModel& link,
 	            const Eigen::Isometry3d& offset, const Eigen::Isometry3d& target, const moveit::core::JointModelGroup* jmg,
 	            double timeout, robot_trajectory::RobotTrajectoryPtr& result,
 	            const moveit_msgs::msg::Constraints& path_constraints = moveit_msgs::msg::Constraints()) override;
--- a/core/include/moveit/task_constructor/solvers/pipeline_planner.h
+++ b/core/include/moveit/task_constructor/solvers/pipeline_planner.h
@@ -59,10 +59,10 @@ MOVEIT_CLASS_FORWARD(PipelinePlanner);
 class PipelinePlanner : public PlannerInterface
 {
 public:
-	/** Simple Constructor to use only a single pipeline
+	/** \brief Simple Constructor to use only a single pipeline
 	 * \param [in] node ROS 2 node
 	 * \param [in] pipeline_name Name of the planning pipeline to be used.
-	 * This is also the assumed namespace where the parameters of this pipeline can be found
+	 * This is also the assumed namespace where the parameters of this pipeline can be found.
 	 * \param [in] planner_id Planner id to be used for planning. Empty string means default.
 	 */
 	PipelinePlanner(const rclcpp::Node::SharedPtr& node, const std::string& pipeline_name = "ompl",
@@ -71,13 +71,16 @@ public:
 
 	/** \brief Constructor
 	 * \param [in] node ROS 2 node
-	 * \param [in] pipeline_id_planner_id_map map containing pairs of pipeline and plugin names to be used for planning
-	 * \param [in] stopping_criterion_callback callback to decide when to stop parallel planning
-	 * \param [in] solution_selection_function callback to choose the best solution from multiple ran pipelines
+	 * \param [in] pipeline_id_planner_id_map Map containing pairs of pipeline and plugin names to be used for planning
+	 * \param [in] planning_pipelines Optional: A map with the names and initialized corresponding planning pipelines
+	 * \param [in] stopping_criterion_callback Optional: Callback to decide when to stop parallel planning
+	 * \param [in] solution_selection_function Optional: Callback to choose the best solution from multiple ran pipelines
 	 */
 	PipelinePlanner(
 	    const rclcpp::Node::SharedPtr& node,
 	    const std::unordered_map<std::string, std::string>& pipeline_id_planner_id_map,
+	    const std::unordered_map<std::string, planning_pipeline::PlanningPipelinePtr>& planning_pipelines =
+	        std::unordered_map<std::string, planning_pipeline::PlanningPipelinePtr>(),
 	    const moveit::planning_pipeline_interfaces::StoppingCriterionFunction& stopping_criterion_callback = nullptr,
 	    const moveit::planning_pipeline_interfaces::SolutionSelectionFunction& solution_selection_function =
 	        &moveit::planning_pipeline_interfaces::getShortestSolution);
@@ -101,7 +104,9 @@ public:
 	void setSolutionSelectionFunction(
 	    const moveit::planning_pipeline_interfaces::SolutionSelectionFunction& solution_selection_function);
 
-	/** \brief If not yet done, initialize pipelines from pipeline_id_planner_id_map
+	/** \brief This function is called when an MTC task that uses this solver is initialized. If no pipelines are
+	 * configured when this function is invoked, the planning pipelines named in the 'pipeline_id_planner_id_map' are
+	 * initialized with the given robot model.
 	 * \param [in] robot_model robot model used to initialize the planning pipelines of this solver
 	 */
 	void init(const moveit::core::RobotModelConstPtr& robot_model) override;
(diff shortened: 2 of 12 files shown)


core/include/moveit/task_constructor/solvers/cartesian_path.h
```diff
@@ -63,11 +63,11 @@ class CartesianPath : public PlannerInterface
 
 	void init(const moveit::core::RobotModelConstPtr& robot_model) override;
 
-	Result plan(const planning_scene::PlanningSceneConstPtr& from, const planning_scene::PlanningSceneConstPtr& to,
+	MoveItErrorCode plan(const planning_scene::PlanningSceneConstPtr& from, const planning_scene::PlanningSceneConstPtr& to,
```
sjahr, line 66: Why did you decide to introduce Result instead of using error codes? Aren't we loosing information by reducing the error code to a boolean?

core/include/moveit/task_constructor/solvers/pipeline_planner.h
```diff
@@ -71,13 +71,16 @@ class PipelinePlanner : public PlannerInterface
...
-	 * \param [in] solution_selection_function callback to choose the best solution from multiple ran pipelines
+	 * \param [in] pipeline_id_planner_id_map Map containing pairs of pipeline and plugin names to be used for planning
+	 * \param [in] planning_pipelines Optional: A map with the names and initialized corresponding planning pipelines
+	 * \param [in] stopping_criterion_callback Optional: Callback to decide when to stop parallel planning
+	 * \param [in] solution_selection_function Optional: Callback to choose the best solution from multiple ran pipelines
 	 */
 	PipelinePlanner(
 	    const rclcpp::Node::SharedPtr& node,
 	    const std::unordered_map<std::string, std::string>& pipeline_id_planner_id_map,
+	    const std::unordered_map<std::string, planning_pipeline::PlanningPipelinePtr>& planning_pipelines =
```
sjahr, line 82: What was you motivation to remove this argument?

core/include/moveit/task_constructor/solvers/pipeline_planner.h
```diff
@@ -101,7 +104,9 @@ class PipelinePlanner : public PlannerInterface
 	void setSolutionSelectionFunction(
 	    const moveit::planning_pipeline_interfaces::SolutionSelectionFunction& solution_selection_function);
 
-	/** \brief If not yet done, initialize pipelines from pipeline_id_planner_id_map
+	/** \brief This function is called when an MTC task that uses this solver is initialized. If no pipelines are
```
sjahr, line 107: Not a strong opinion here, but I think more verbose comments are helpful. Where some of them wrong?

core/src/solvers/joint_interpolation.cpp
```diff
@@ -78,10 +78,10 @@ PlannerInterface::Result JointInterpolationPlanner::plan(const planning_scene::P
 	// add first point
 	result->addSuffixWayPoint(from->getCurrentState(), 0.0);
 	if (from->isStateColliding(from_state, jmg->getName()))
-		return { false, "Start state is in collision!" };
+		return MoveItErrorCode(MoveItErrorCodes::START_STATE_IN_COLLISION, "Start state is in collision!");
```
sjahr, line 81: These returned in `ros-old` just FAILURE, I changed them so that the error code actually provides value

core/src/solvers/pipeline_planner.cpp
```diff
@@ -116,9 +124,11 @@ void PipelinePlanner::init(const core::RobotModelConstPtr& robot_model) {
 		}
 		planning_pipelines_ = moveit::planning_pipeline_interfaces::createPlanningPipelineMap(std::move(pipeline_names),
 		                                                                                      robot_model, node_);
+
 		// Check if it is still empty
 		if (planning_pipelines_.empty()) {
-			throw std::runtime_error("Failed to initialize PipelinePlanner: Could not create any valid pipeline");
+			throw std::runtime_error(
+			    "Cannot initialize PipelinePlanner: Could not create any valid entries for planning pipeline maps!");
```
sjahr, line 131:
```suggestion
			throw std::runtime_error("Failed to initialize PipelinePlanner: Could not create any valid pipeline");
```

core/src/stages/connect.cpp
```diff
@@ -261,7 +261,6 @@ SubTrajectoryPtr Connect::merge(const std::vector<PlannerIdTrajectoryPair>& sub_
 	                                              properties().get<moveit_msgs::msg::Constraints>("path_constraints")))
 		return SubTrajectoryPtr();
 
-	return std::make_shared<SubTrajectory>(trajectory);
```
sjahr, line 264: Small clean-up, we can merge this in any case

core/include/moveit/task_constructor/solvers/cartesian_path.h
rhaschke: Originally, #525 introduced `tl::expected<bool, std::string>` to return both, a boolean result and an error message. Later, this was changed to `MoveItErrorCode`, which - in ROS2 - supports the error message and an error source. As this extended MoveItErrorCode is not available in ROS1 and the purpose was just to return an additional string message, I introduced an explicit data type that is uniformly usable in ROS1 and ROS2.
If at all, the source information is lost (it was never fed into the error code).

core/include/moveit/task_constructor/solvers/pipeline_planner.h
rhaschke: There were two/three conflicting approaches to instantiate the internal pipelines map:
1. By passing a map of instantiated pipelines via the "optional" `planning_pipelines` argument
2. By passing the `pipeline_id_planner_id_map`, which was used to create the former if not provided
3. A `pipeline_id_planner_id_map` property of the stage

In any case, the arguments `pipeline_id_planner_id_map` and `planning_pipelines` were designed to be _mutually exclusive_. This should have been implemented with two separate constructor methods to avoid inconsistencies.

Eventually, the property value was the driving factor for actual planning. Hence, I decided to just go for this property and - for convenience - initialize that property from the passed map argument.

I dropped the first option completely, because instantiating pipelines during stage construction is not trivial. It requires to pass the same `RobotModelPtr` that is used by the task. Having this option, puts you at risk to have a pipelines map that is inconsistent with the map property.

core/include/moveit/task_constructor/solvers/pipeline_planner.h
rhaschke: Which information is missing with the shortened version?
\brief comments should be brief one-liners :wink:

core/src/solvers/joint_interpolation.cpp
rhaschke: Note, that the error code is never actually considered in code - we just check for success. 
It might be useful for inspection during debugging though.

rhaschke:
Thanks for your questions. I hope I answered all of them. Unfortunately, the diff link I provided originally, doesn't allow for comments. I just noticed that now.
I'm collecting fixups to the ros2 branch in a new PR: #547. Thanks for your input!
Did somebody already figure out, why the default goal reaching threshold is relaxed in ROS2 compared to ROS1?

core/include/moveit/task_constructor/solvers/cartesian_path.h
rhaschke: I don't consider reverting to MoveItErrorCode and I don't see something missing so far.

sjahr: > If at all, the source information is lost (it was never fed into the error code).

Good point that should be done, otherwise that field is pointless.

I see, have you considered enhancing the error code with a string message in moveit1? The main motivation here was to improve the debug-ability of moveit in general. The source field was introduced to infer the planning pipeline that created the error message when using parallel planning. Without that, I think it is just possible to tell that one of the planners failed with the given error message. Not which one. Alternatively, the string could be enhanced with that information.

core/src/solvers/pipeline_planner.cpp
```diff
@@ -208,11 +220,10 @@ PlannerInterface::Result PipelinePlanner::plan(const planning_scene::PlanningSce
 			// Choose the first solution trajectory as response
 			result = solution.trajectory;
 			last_successful_planner_ = solution.planner_id;
-			return { true, "" };
+			return MoveItErrorCode(solution.error_code.val, solution.error_code.message, solution.error_code.source);
 		}
```
sjahr, line 224:
```suggestion
		}
			return MoveItErrorCode(solution.error_code.val, solution.error_code.message, solution.error_code.source);		
```
As pointed out above, the complete error code should be returned in any case

core/include/moveit/task_constructor/solvers/pipeline_planner.h
sjahr: Makes sense, thanks for that explanation!

core/src/solvers/joint_interpolation.cpp
sjahr: Yes, we did that in some custom applications. Given that MoveIt generates this error code, it would be great not to loose this information in the MTC solution.

sjahr:
@rhaschke Thanks for your detailed explanations. That is very helpful and going through you changes was insightful too!

> Did somebody already figure out, why the default goal reaching threshold is relaxed in ROS2 compared to ROS1?

I can take a look into it. With Michael's in moveit2 the MTC pick and place demo worked for me without needing to adjust the threshold but maybe I am missing something here.

core/include/moveit/task_constructor/solvers/cartesian_path.h
rhaschke: Yes, I shortly considered augmenting MoveItErrorCode in MoveIt1 as well. However, I decided against, because the error message is redundant as we already have a fixed mapping from the error code to a human-readable message (`MoveItErrorCode::toString()`). Further, the source member was not filled regularly - only in planning adapters. As far as I can see, the planning pipeline name is not yet stored in that field.

Note, that there are _two_ MoveItErrorCode header files (I think one should be removed):
- `moveit_core/utils/include/moveit/utils/moveit_error_code.h` (regularly used)
- `moveit_ros/hybrid_planning/hybrid_planning_manager/hybrid_planning_manager_component/include/moveit/hybrid_planning_manager/moveit_error_code_interface.h` (used only once)

core/src/solvers/joint_interpolation.cpp
rhaschke: How did you access that code in custom applications? It is never exposed from any stage!?

rhaschke:
> The MTC pick and place demo worked for me without needing to adjust the threshold.

The code using this threshold was only recently added in the master branch. So, if pick+place worked before on the old ros2 branch, this doesn't come at a surprise for me.

core/src/solvers/pipeline_planner.cpp
```diff
@@ -54,12 +54,14 @@ namespace task_constructor {
 namespace solvers {
 
 using PipelineMap = std::unordered_map<std::string, std::string>;
+;
```
sjahr, line 57:
```suggestion
```

core/src/solvers/joint_interpolation.cpp
sjahr: You're right, I think was wrong about that, sorry

Issue #543, opened 2024-03-17
Custom Generator stage with different start and end states


DaniGarciaLopez:
I'm attempting to create a custom Generator stage to follow a set of Cartesian waypoints. Specifically, I want it to be a Generator stage so that it provides the start and end states to adjacent stages. This way, I could place this stage between two Connect stages, for example.

The issue I'm encountering is that when I compute the trajectory and call the spawn method, it copies the start and end states ([here](https://github.com/ros-planning/moveit_task_constructor/blob/ros2/core/src/stage.cpp#L230)), thus preventing me from having a trajectory that starts and ends at different waypoints. It's only valid for symmetrical trajectories. Other methods, such as [sendForward](https://github.com/ros-planning/moveit_task_constructor/blob/e1b891ba5a346eae14454b3c23f74052d9ada23e/core/include/moveit/task_constructor/stage.h#L316), [sendBackward](https://github.com/ros-planning/moveit_task_constructor/blob/e1b891ba5a346eae14454b3c23f74052d9ada23e/core/include/moveit/task_constructor/stage.h#L319), and [connect](https://github.com/ros-planning/moveit_task_constructor/blob/e1b891ba5a346eae14454b3c23f74052d9ada23e/core/include/moveit/task_constructor/stage.h#L418), receive start and end InterfaceStates, but these are not available in the Generator class.

I was considering inheriting from the PropagatingEitherWay class. Initially, I believed it could only propagate in one direction (from bottom to top or top to bottom). However, checking [this comment](https://github.com/ros-planning/moveit_task_constructor/blob/e1b891ba5a346eae14454b3c23f74052d9ada23e/core/include/moveit/task_constructor/stage.h#L288C47-L288C66), it might also be able to propagate in both directions, acting as a Generator? Althought there is no both direction propagation in its [direction settings](https://github.com/ros-planning/moveit_task_constructor/blob/e1b891ba5a346eae14454b3c23f74052d9ada23e/core/include/moveit/task_constructor/stage.h#L296C2-L301C4). Some clarification on this point would be appreciated.

[...]

PR #546, opened 2024-03-18
Set independant start and end states in a Generator stage

DaniGarciaLopez:
Releated to https://github.com/ros-planning/moveit_task_constructor/issues/543. This PR enables the possibility to set independent start and end states when calling the `spawn` method of a Generator stage, while maintaining the previous functionality that uses the same state.

I didn't implement [this overload](https://github.com/ros-planning/moveit_task_constructor/blob/e1b891ba5a346eae14454b3c23f74052d9ada23e/core/include/moveit/task_constructor/stage.h#L369C2-L373C3) as it implies that the trajectory will be empty and it makes no sense to set different start and end states when the robot is not actually moving, right? Or am I missing any specific use case where this could be handy?

I noticed that [`connect`](https://github.com/ros-planning/moveit_task_constructor/blob/e1b891ba5a346eae14454b3c23f74052d9ada23e/core/include/moveit/task_constructor/stage.h#L418) uses const lvalue reference (`const InterfaceState&`) instead of rvalue (`InterfaceState&&`). Are there any design preferences?

Is there an easy way to backport this to humble branch or I should cherry-pick another PR?

### DIFF
--- a/core/include/moveit/task_constructor/stage.h
+++ b/core/include/moveit/task_constructor/stage.h
@@ -365,6 +365,7 @@ public:
 
 	virtual bool canCompute() const = 0;
 	virtual void compute() = 0;
+	void spawn(InterfaceState&& from, InterfaceState&& to, SubTrajectory&& trajectory);
 	void spawn(InterfaceState&& state, SubTrajectory&& trajectory);
 	void spawn(InterfaceState&& state, double cost) {
 		SubTrajectory trajectory;
--- a/core/src/stage.cpp
+++ b/core/src/stage.cpp
@@ -219,26 +219,30 @@ void StagePrivate::sendBackward(InterfaceState&& from, const InterfaceState& to,
 	newSolution(solution);
 }
 
-void StagePrivate::spawn(InterfaceState&& state, const SolutionBasePtr& solution) {
+void StagePrivate::spawn(InterfaceState&& from, InterfaceState&& to, const SolutionBasePtr& solution) {
 	assert(prevEnds() && nextStarts());
 
-	computeCost(state, state, *solution);
+	computeCost(from, to, *solution);
 
 	if (!storeSolution(solution, nullptr, nullptr))
 		return;  // solution dropped
 
-	auto from = states_.insert(states_.end(), InterfaceState(state));  // copy
-	auto to = states_.insert(states_.end(), std::move(state));
+	auto from_it = states_.insert(states_.end(), std::move(from));
+	auto to_it = states_.insert(states_.end(), std::move(to));
 
-	solution->setStartState(*from);
-	solution->setEndState(*to);
+	solution->setStartState(*from_it);
+	solution->setEndState(*to_it);
 
 	if (!solution->isFailure()) {
-		prevEnds()->add(*from);
-		nextStarts()->add(*to);
+		prevEnds()->add(*from_it);
+		nextStarts()->add(*to_it);
 	}
 
 	newSolution(solution);
+}
+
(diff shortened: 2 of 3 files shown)


core/include/moveit/task_constructor/stage.h
```diff
@@ -365,6 +365,7 @@ class Generator : public ComputeBase
 
 	virtual bool canCompute() const = 0;
 	virtual void compute() = 0;
+	void spawn(InterfaceState&& from, InterfaceState&& to, SubTrajectory&& trajectory);
```
rhaschke, line 368: Please rename state arguments to `start` and `end`, which gives them a more concrete meaning.

core/src/stage.cpp
```diff
@@ -219,28 +219,32 @@ void StagePrivate::sendBackward(InterfaceState&& from, const InterfaceState& to,
...
-		nextStarts()->add(*to);
+		prevEnds()->add(*from_it);
+		nextStarts()->add(*to_it);
 	}
 
 	newSolution(solution);
 }
 
+void StagePrivate::spawn(InterfaceState&& state, const SolutionBasePtr& solution) {
+    spawn(std::move(state), std::move(state), solution); // copy start and end states
```
rhaschke, line 245: You cannot move `state` twice. The first `std::move` needs to create a copy:
```suggestion
    spawn(InterfaceState(state), std::move(state), solution);
```

core/src/stage.cpp
```diff
@@ -219,28 +219,32 @@ void StagePrivate::sendBackward(InterfaceState&& from, const InterfaceState& to,
...
 
-	computeCost(state, state, *solution);
-
-	if (!storeSolution(solution, nullptr, nullptr))
+	if (!storeSolution(solution, &from, &to))
 		return;  // solution dropped
 
-	auto from = states_.insert(states_.end(), InterfaceState(state));  // copy
-	auto to = states_.insert(states_.end(), std::move(state));
+	auto from_it = states_.insert(states_.end(), InterfaceState(from));
```
rhaschke, line 230: Here you can move:
```suggestion
	auto from_it = states_.insert(states_.end(), std::move(from));
```

core/src/stage.cpp
```diff
@@ -699,7 +707,7 @@ MonitoringGenerator::MonitoringGenerator(const std::string& name, Stage* monitor
   : Generator(new MonitoringGeneratorPrivate(this, name)) {
 	setMonitoredStage(monitored);
 }
-
+
```
rhaschke, line 710: Please remove the extra white space
```suggestion

```

rhaschke (commented):
Thanks for this contribution. I have a few comments to address.

DaniGarciaLopez:
Thanks for the review @rhaschke, changes addressed!

rhaschke:
I noticed that the remaining functions also use the argument names `from` and `to`. For this reason, I reverted the requested change. Sorry for the confusion.

As this change should become part of master branch too, could you please (squash)rebase onto and retarget to master?
Thanks.

DaniGarciaLopez:
I also personally believe that using `start` and `end` is more intuitive. However, to maintain consistency, IMHO sticking with `from` and `to` seems to be the clearer option.

Commits squashed!

DaniGarciaLopez:
Not sure how to retarget to master with that many commits ahead. Can you retarget the PR or should I open a new one?

DaniGarciaLopez:
Ok, thanks. Done!

rhaschke (approved):

PR #554, opened 2024-04-04
#719 change mismatched variable names

anikk94:
When building with moveit2 tutorials, I got the same build errors described in #719. I then made the corrections @FlorianHou mentioned and moved on. I want to push these changes to the ros2 branch which is what is used in the tutorials (by vcs import). 
#719


rhaschke:
These changes are required for MoveIt Humble only. We do have a separate branch for this already: `humble`.

PR #569, opened 2024-05-15
Update .pre-commit-config.yaml

mosfet80:
updated pre commit libraries


rhaschke:
This is already part of #560, which is waiting for a review. Closing here.

PR #570, opened 2024-05-15
Update prerelease.yaml

mosfet80:
switch from ros1 distro "noetic" to ros 2 distro "humble"


rhaschke:
Thanks for pointing this out. I made the ROS_DISTRO an interactive argument in https://github.com/ubi-agni/moveit_task_constructor/commit/797c798da9cc6858a4c814ad0ab46d27edaf823c, which is a more generic approach. Closing here.

PR #590, opened 2024-07-05
Add property to enable/disable pruning at runtime

rhaschke:
Rebase and cleanup of #358. Closes #358.


core/CMakeLists.txt
```diff
@@ -40,9 +40,7 @@ catkin_package(
 )
 
 add_compile_options(-fvisibility-inlines-hidden)
-if(ENABLE_PRUNING)
-	add_compile_definitions(ENABLE_PRUNING)
-endif()
+add_compile_definitions(ENABLE_PRUNING=$<BOOL:${ENABLE_PRUNING}>)
```
sjahr, line 43: What is that doing?

core/src/task.cpp
```diff
@@ -92,6 +92,7 @@ const ContainerBase* TaskPrivate::stages() const {
 
 Task::Task(const std::string& ns, bool introspection, ContainerBase::pointer&& container)
   : WrapperBase(new TaskPrivate(this, ns), std::move(container)) {
+	setPruning(ENABLE_PRUNING);
```
sjahr, line 95: Wouldn't it be possible to just have a C++ API with defaulting to true without exposing this via CMake?

sjahr (commented):
I like the idea of having a simple property based API around this.I am not fully understanding the macros around this, is it necessary to have a macro-based API around this?

core/CMakeLists.txt
rhaschke: This is exposing the cmake option as a compiler definition to serve as the default of `setPruning()` below.

PR #593, opened 2024-07-15
MoveRelative: gracefully handle identical min/max distance

### LINKED ISSUE #586: Min distance not reached when setting the same min/max values in the MoveRelative stage
Consider the case where I want to move up a link by using the `setMinMaxDistance` API with the same distance for min and max. (Normally for such a case, we would not set the min/max distance and just set the direction accordingly)

```cpp
    auto stage = std::make_unique<moveit::task_constructor::stages::MoveRelative>("lift up", solver);
    
    stage->setMinMaxDistance(0.06, 0.06);

    // Set upward direction
    geometry_msgs::Vector3Stamped vec;
    vec.header.frame_id = from_object;
    vec.vector.x = -1.0;
    stage->setDirection(vec);
```
In some cases, I get the error below:

image

I think we should add a `min_distance_tolerance` property. What do you think ?
```cpp
// min_distance reached?
if (min_distance > 0.0) {
    success = distance >= min_distance - props.get<double>("min_distance_tolerance");
if (!success) {
    char msg[100];
    snprintf(msg, sizeof(msg), "min_distance not reached (%.17g < %.17g)", distance, min_distance);
    solution.setComment(msg);
}
```

rhaschke:
When min_distance == max_distance > 0.0, the minimal distance might be missed due to numerical errors.
To avoid this, deactivate minimal distance check and run the full distance as given by max_distance.
Fixes #586. @captain-yoshi, please review.

### DIFF
--- a/core/src/stages/move_relative.cpp
+++ b/core/src/stages/move_relative.cpp
@@ -185,6 +185,11 @@ bool MoveRelative::compute(const InterfaceState& state, planning_scene::Planning
 
 	double max_distance = props.get<double>("max_distance");
 	double min_distance = props.get<double>("min_distance");
+
+	// if min_distance == max_distance, resort to moving full distance (negative min_distance)
+	if (max_distance > 0.0 && std::fabs(max_distance - min_distance) < std::numeric_limits<double>::epsilon())
+		min_distance = -1.0;
+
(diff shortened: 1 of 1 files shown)


captain-yoshi (approved):
Works on my setup. Thanks :)

PR #595, opened 2024-07-15
Fix `MoveRelative` potentially empty trajectory issue

pac48:
We have run into an issue a few times with `MoveRelative` relative that causes a crash. If the ``PlannerInterface` implementation returns a trajectory with no waypoints, but a non-null `robot_trajectory`, then the back of an empty vector is accessed via `robot_trajectory->getLastWayPointPtr()` leading to a crash. 

A potential fix is to consider empty trajectories as a failed plan. We run into this issue frequently when the robot starts in or very close to a collision.

### DIFF
--- a/core/src/stages/move_relative.cpp
+++ b/core/src/stages/move_relative.cpp
@@ -291,7 +291,8 @@ bool MoveRelative::compute(const InterfaceState& state, planning_scene::Planning
 		if (!success)
 			comment = result.message;
 
-		if (robot_trajectory) {  // the following requires a robot_trajectory returned from planning
+		if (robot_trajectory && robot_trajectory->getWayPointCount() > 0) {  // the following requires a robot_trajectory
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
Thanks. Don't you need the same check here as well:
https://github.com/moveit/moveit_task_constructor/blob/e3cc52cd0f018d91c119675cc11ee6d921e4109c/core/src/stages/move_relative.cpp#L330

pac48:
> Thanks. Don't you need the same check here as well:
> 
> https://github.com/moveit/moveit_task_constructor/blob/e3cc52cd0f018d91c119675cc11ee6d921e4109c/core/src/stages/move_relative.cpp#L330

Yes, we need to add the check there as well. Thanks for pointing that out.

rhaschke (approved):

Issue #312, opened 2021-11-26
Scheduling


rhaschke:

DaniGarciaLopez:
Any updates on this feature? As I understand, this would greatly reduce the planning time as it allows parallel planning in different threads, right? 

We are very interested in this feature as our planning time is already too long for our application. The main solution we have found is to precompute the task offline, as mentioned [here](https://github.com/ros-planning/moveit_task_constructor/issues/198), but sometimes it is necessary to do it in real time.

Are there any plans to implement it in the near future?

rhaschke:
Unfortunately, we didn't have time to start working on this and I don't foresee any change of that situation.
However, you are highly welcome to contribute to the project.

PR #597, opened 2024-07-16
Fix early preempt when planning

captain-yoshi:
We have a usecase in which we need to plan the task in another thread and sometimes preempt it. When preempting early, this does not work because the preempt flag is reseted inside the `plan()` method.

This fix makes the preempt flag changeable from the user. So for the first planning nothing changes. But users must reset the preempt flag before each subsequent planning (if the preempt method is used).

----
Even with this fix, preempting a task can take some time. The preempt flag is only checked at the start of the plan compute loop. #312 seems far away, but I think we could reduce the time it takes to preempt a task without too much changes ?


rhaschke (commented):
> When preempting early, this does not work because the preempt flag is reseted inside the plan() method.

I don't yet understand the race condition. Do you mean, `preempt_requested_` may be set before it is initialized in `plan()` and then the request is lost?

core/include/moveit/task_constructor/task.h
```diff
@@ -126,9 +126,10 @@ class Task : protected WrapperBase
 	void init();
 
 	/// reset, init scene (if not yet done), and init all stages, then start planning
+	/// when preempting a plan, you must reset the preempt flag before subsequent planning
 	moveit::core::MoveItErrorCode plan(size_t max_solutions = 0);
-	/// interrupt current planning (or execution)
-	void preempt();
+	/// interrupt current planning (or execution) or reset the preempt flag (false)
```
rhaschke, line 131: Execution cannot be preempted with this call.
```suggestion
	/// interrupt current planning or reset the preempt flag (false)
```

rhaschke (commented):
I don't like that the user needs to remember to reset the preempt_requested_ flag.
What about resetting the flag when leaving Task::plan()?

core/include/moveit/task_constructor/task.h
captain-yoshi: That would be more clean. We would need a try catch block and catch everything in the plan() to reset the value.

captain-yoshi:
> > When preempting early, this does not work because the preempt flag is reseted inside the plan() method.
> 
> I don't yet understand the race condition. Do you mean, `preempt_requested_` may be set before it is initialized in `plan()` and then the request is lost?

Yes exactly.

captain-yoshi:
> What about resetting the flag when leaving Task::plan()?

@rhaschke This would work in most of the cases (see last commit). But consider the following case:

| Thread + Sequence | Method          | Info                                                              |
|-------------------|-----------------|-------------------------------------------------------------------|
| Main              | Task::plan()    | The plan finisihes and the flag is resetted.                      |
| T1                 | Task::preempt() | Flag is set to true;                                              |
| Main or T1         | Task::plan()    | This planning will not work as the preempted flag is set to true. |

So I don't think it can be done it automatically...

core/test/test_container.cpp
```diff
@@ -672,3 +672,27 @@ TEST(Task, timeout) {
...
+	resetMockupIds();
+
+	Task t;
+	t.setRobotModel(getModel());
+
+	auto timeout = std::chrono::milliseconds(10);
+	t.add(std::make_unique<GeneratorMockup>(PredefinedCosts::constant(0.0)));
+	t.add(std::make_unique<TimedForwardMockup>(timeout));
+
+	// preempt before any stage computation is done
```
rhaschke, line 689:
```suggestion
	// preempt before preempt_request_ is reset in plan()
```

core/test/test_container.cpp
```diff
@@ -672,3 +672,27 @@ TEST(Task, timeout) {
...
+	{
+		std::thread thread{ [&ec, &t] { ec = t.plan(1); } };
```
rhaschke, line 691: What about introducing a small sleep before t.plan() in order to emphasize the issue even more?

core/src/task.cpp
```diff
@@ -234,31 +234,39 @@ void Task::compute() {
...
+			compute();
+			for (const auto& cb : impl->task_cbs_)
+				cb(*this);
+			if (impl->introspection_)
+				impl->introspection_->publishTaskState();
+		};
+		return success_or(moveit::core::MoveItErrorCode::PLANNING_FAILED);
+	} catch (const std::exception& e) {
+		pimpl()->preempt_requested_ = false;
+		throw;  // rethrow the original exception
```
rhaschke, line 268: The try-catch-block is the wrong tool to cleanup before leaving the function scope: it is only considered if an exception is thrown, not when leaving the function via return. Please revert this.
What you need is a scope guard:
https://stackoverflow.com/questions/50182244/simple-way-to-execute-code-at-the-end-of-function-scope

rhaschke (commented):
I'm curious how you manager to call preempt() before plan() actually started its work.
I do see the theoretical possibility, but practically that shouldn't occur.

core/src/task.cpp
captain-yoshi: Would using this external [library](https://ricab.github.io/scope_guard/) be acceptable ? At a first glance it seems well tested and easy to use.

captain-yoshi:
> I'm curious how you manager to call preempt() before plan() actually started its work.

We started using BehaviorTree.CPP  to create MTC tasks. I was adding a test to test the preempting of a BT node when planning. Here is the sample.
```xml
<root BTCPP_format="4" >
  <BehaviorTree ID="MainTree">
    <Sequence>
      <Script code="test := true"/>
      <ReactiveSequence name="root">
        <ForceFailure _skipIf="test">
          <Script code=" result:='error' " />
        </ForceFailure>

        <Script code=" result:='error' " _onSuccess="test = false" />
        <PlanMTCTask task="{mtc_task}" max_solutions="1000" />

      </ReactiveSequence>
    </Sequence>
  </BehaviorTree>
</root>
```

So planning is first started in another thread. Then the main thread will check if the BT node `PlanMTCTask` is done. After that the ReactiveSequence will halt all children because of ForceFailure triggering on the second pass.

captain-yoshi:
@rhaschke Ready for another review with all changes requested.

This adds an external dependency to the [scope_guard](https://ricab.github.io/scope_guard/) library.

rhaschke (approved):

rhaschke:
If you agree to my simplification, this is ready to merge.

captain-yoshi:
LGTM and thanks for the review.

PR #603, opened 2024-08-12
Avoid segfault and improve error warnings

sjahr:
Check for nullptr in computeIK and print warnings when returning a nullptr

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -46,7 +46,7 @@
 #include <fmt/format.h>
 
 namespace {
-
+const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_task_constructor_visualization.execute_task_solution");
 // TODO: move to moveit::core::RobotModel
 const moveit::core::JointModelGroup* findJointModelGroup(const moveit::core::RobotModel& model,
                                                          const std::vector<std::string>& joints) {
@@ -73,12 +73,10 @@ const moveit::core::JointModelGroup* findJointModelGroup(const moveit::core::Rob
 				return jmg;
 		}
 	}
-
+	RCLCPP_WARN(LOGGER, "Could not find JointModelGroup that actuates {}", fmt::join(joints, ", "));
 	return nullptr;
 }
 }  // namespace
-
-static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_task_constructor_visualization.execute_task_solution");
 
 namespace move_group {
 
--- a/core/src/container.cpp
+++ b/core/src/container.cpp
@@ -345,6 +345,7 @@ Stage* ContainerBase::findChild(const std::string& name) const {
 			else if (auto* parent = dynamic_cast<const ContainerBase*>(child.get()))
 				return parent->findChild(name.substr(pos + 1));
 		}
+	RCLCPP_WARN_STREAM(rclcpp::get_logger("ContainerBase"), fmt::format("Child '{}' not found", name));
 	return nullptr;
 }
 
--- a/core/src/stages/compute_ik.cpp
+++ b/core/src/stages/compute_ik.cpp
@@ -314,6 +314,11 @@ void ComputeIK::compute() {
 
 		link = scene->getCurrentState().getRigidlyConnectedParentLinkModel(ik_pose_msg.header.frame_id);
 
+		if (!link) {
+			RCLCPP_WARN_STREAM(
+			    LOGGER, fmt::format("ik frame '{}' is not rigidly connected to any link", ik_pose_msg.header.frame_id));
+			return;
+		}
 		// transform target pose such that ik frame will reach there if link does
 		target_pose = target_pose * ik_pose.inverse() * scene->getCurrentState().getFrameTransform(link->getName());
 	}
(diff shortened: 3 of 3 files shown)


core/src/container.cpp
```diff
@@ -345,6 +345,7 @@ Stage* ContainerBase::findChild(const std::string& name) const {
 			else if (auto* parent = dynamic_cast<const ContainerBase*>(child.get()))
 				return parent->findChild(name.substr(pos + 1));
 		}
+	RCLCPP_WARN_STREAM(rclcpp::get_logger("ContainerBase"), fmt::format("Child '{}' not found", name));
```
rhaschke, line 348: I'm not convinced that the warning should be thrown here. The caller should handle a potential nullptr and may not want to have a warning issued.

core/src/stages/compute_ik.cpp
```diff
@@ -314,6 +314,11 @@ void ComputeIK::compute() {
 
 		link = scene->getCurrentState().getRigidlyConnectedParentLinkModel(ik_pose_msg.header.frame_id);
 
+		if (!link) {
```
rhaschke, line 317: `knowsFrameTransform()`, a few lines above, should have validated that the frame exists.
Under what circumstances do you run into a segfault here?
I cannot imagine a valid frame that returns NULL here.

core/src/stages/compute_ik.cpp
```diff
@@ -314,6 +314,11 @@ void ComputeIK::compute() {
...
+			RCLCPP_WARN_STREAM(
+			    LOGGER, fmt::format("ik frame '{}' is not rigidly connected to any link", ik_pose_msg.header.frame_id));
```
rhaschke, line 319: The error message doesn't make sense: an existing frame needs to be connected to some link.

core/src/stages/compute_ik.cpp
```diff
@@ -336,23 +341,25 @@ void ComputeIK::compute() {
...
-		solution.setComment(s.comment() + " eef in collision: " + listCollisionPairs(collisions.contacts, ", "));
-		auto colliding_scene{ scene->diff() };
-		colliding_scene->setCurrentState(sandbox_state);
-		spawn(InterfaceState(colliding_scene), std::move(solution));
-		return;
-	} else
-		generateVisualMarkers(sandbox_state, appender, links_to_visualize);
+	auto const& connected_parent_link_model = moveit::core::RobotModel::getRigidlyConnectedParentLinkModel(link);
+	if (connected_parent_link_model && connected_parent_link_model->getParentJointModel()) {
+		const auto& links_to_visualize = connected_parent_link_model->getParentJointModel()->getDescendantLinkModels();
```
rhaschke, line 346: Again, under which circumstances do you run into a segfault here?
`getRigidlyConnectedParentLinkModel()` should always return a valid LinkModel.

rhaschke (commented):
I don't yet see the need for these extra checks. Can you please elaborate on your error condition.

core/src/stages/compute_ik.cpp
sjahr: There seem to be cases under which [getRigidlyConnectedParentLinkModel](https://github.com/moveit/moveit2/blob/aece55c3ebe4a88bb9c09b85fa12479b94690c0b/moveit_core/robot_state/src/robot_state.cpp#L910) returns a nullptr. Are they all covered by `knowsFrameTransform`?

core/src/stages/compute_ik.cpp
sjahr, line 319: Would this be more precise?
```suggestion
			    LOGGER, fmt::format("ik frame '{}' is not connected to any parent link", ik_pose_msg.header.frame_id));
```

core/src/stages/compute_ik.cpp
sjahr: You're right, if link is guaranteed not to be a nullptr, `getRigidlyConnectedParentLinkModel` doesn't return a nullptr :+1:

core/src/container.cpp
sjahr: We could consider making it a debug statement? The main point of this print statement is to make debuggability better

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -73,13 +73,11 @@ const moveit::core::JointModelGroup* findJointModelGroup(const moveit::core::Rob
 				return jmg;
 		}
 	}
-
+	RCLCPP_WARN(LOGGER, fmt::format("Could not find JointModelGroup that actuates {}", fmt::join(joints, ", ")));
```
sjahr, line 76:
```suggestion
```

core/src/stages/compute_ik.cpp
rhaschke: > There seem to be cases under which [getRigidlyConnectedParentLinkModel](https://github.com/moveit/moveit2/blob/aece55c3ebe4a88bb9c09b85fa12479b94690c0b/moveit_core/robot_state/src/robot_state.cpp#L910) returns a nullptr. Are they all covered by `knowsFrameTransform`?

Yes, the should be covered by `knowsFrameTransform`. If not, that's probably a bug. Please state your concrete problem.

core/src/stages/compute_ik.cpp
rhaschke: As I said, I expect this to return NULL if and only if frame doesn't exist. Do you have other evidence?

core/src/container.cpp
rhaschke: Yes, if that helps you.

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -46,7 +46,7 @@
 #include <fmt/format.h>
 
 namespace {
-
+const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_task_constructor_visualization.execute_task_solution");
```
rhaschke, line 49: Please revert those changes too.

rhaschke (changes requested):

capabilities/src/execute_task_solution_capability.cpp
sjahr, line 49:
```suggestion
```

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -78,8 +78,6 @@ const moveit::core::JointModelGroup* findJointModelGroup(const moveit::core::Rob
 }
 }  // namespace
```
sjahr, line 80:
```suggestion
static const rclcpp::Logger LOGGER = rclcpp::get_logger("moveit_task_constructor_visualization.execute_task_solution");

```

core/src/container.cpp
sjahr, line 348:
```suggestion
	RCLCPP_DEBUG_STREAM(rclcpp::get_logger("ContainerBase"), fmt::format("Child '{}' not found", name));
```

capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -77,7 +76,6 @@ const moveit::core::JointModelGroup* findJointModelGroup(const moveit::core::Rob
 	return nullptr;
 }
 }  // namespace
```
sjahr, line 78:
```suggestion
}  // namespace

```

core/src/stages/compute_ik.cpp
rhaschke: @sjahr: I'm still curious to learn about your concrete issue. 
In which case does `getRigidlyConnectedParentLinkModel` returns NULL, which is not handled before by `knowsFrameTransform`?

sjahr: Sorry, I cannot share more info on this at the moment :disappointed: I'll try debugging it further and hopefully figure out what the real source behind this is.

rhaschke: I'm not expecting a detailed error analysis, but ideally a code snippet reproducing your segfault - I assume you observed a segfault in some code.

sjahr: Yes, unfortunately I don't have access to that code. If you'd prefer that I can close this PR for now and come back later with more solid changes?

sjahr:
Closing this for now and I'll open a PR / issue when I have a better understanding of the problem or a solution. Thanks for your feedback!

PR #608, opened 2024-08-26
Update run.launch.py

KarthiAru:
Added the path for the panda_config.yaml file. This is required to run the pick_place_demo example.

### DIFF
--- a/demo/launch/run.launch.py
+++ b/demo/launch/run.launch.py
@@ -17,6 +17,11 @@ def generate_launch_description():
         executable=LaunchConfiguration("exe"),
         output="screen",
         parameters=[
+            os.path.join(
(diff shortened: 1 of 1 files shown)


demo/launch/run.launch.py
```diff
@@ -17,6 +17,11 @@ def generate_launch_description():
         executable=LaunchConfiguration("exe"),
         output="screen",
         parameters=[
+            os.path.join(
+                get_package_share_directory("moveit_task_constructor_demo"),
+                "config",
+                "panda_config.yaml",
+                ),
```
rhaschke, line 24: Are you sure this works?
As far as I know, parameters should be a list of dictionaries (representing the parameter values).

rhaschke:
Cherry-picked #607.

PR #610, opened 2024-09-01
clean up dependencies for rviz_marker_tools

v4hn:
`liburdfdom-dev` is not actually used. Only `liburdfdom-headers-dev`.

I cleaned up a few other things along the way while working on building packages for it in [my ros-o github builder](https://v4hn.github.io/ros-o-builder/).

### DIFF
--- a/rviz_marker_tools/package.xml
+++ b/rviz_marker_tools/package.xml
@@ -9,9 +9,26 @@
 
 	<buildtool_depend>catkin</buildtool_depend>
 
-	<depend>visualization_msgs</depend>
-	<depend>geometry_msgs</depend>
-	<depend>roscpp</depend>
-	<depend>tf2_eigen</depend>
-	<depend>liburdfdom-dev</depend>
+	<build_depend>eigen</build_depend>
+	<build_export_depend>eigen</build_export_depend>
+
+	<build_depend>geometry_msgs</build_depend>
+	<build_export_depend>geometry_msgs</build_export_depend>
+	<exec_depend>geometry_msgs</exec_depend>
+
+	<build_depend>liburdfdom-headers-dev</build_depend>
+
+	<build_depend>roscpp</build_depend>
+	<exec_depend>roscpp</exec_depend>
+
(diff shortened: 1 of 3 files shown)


rviz_marker_tools/package.xml
```diff
@@ -9,9 +9,26 @@
...
-	<depend>geometry_msgs</depend>
-	<depend>roscpp</depend>
-	<depend>tf2_eigen</depend>
-	<depend>liburdfdom-dev</depend>
+	<build_depend>eigen</build_depend>
+	<build_export_depend>eigen</build_export_depend>
+
+	<build_depend>geometry_msgs</build_depend>
+	<build_export_depend>geometry_msgs</build_export_depend>
+	<exec_depend>geometry_msgs</exec_depend>
```
rhaschke, line 17: Why do you prefer this verbose form over `<depend>`?

rviz_marker_tools/package.xml
```diff
@@ -9,9 +9,26 @@
...
+
+	<build_depend>liburdfdom-headers-dev</build_depend>
+
+	<build_depend>roscpp</build_depend>
+	<exec_depend>roscpp</exec_depend>
```
rhaschke, line 22: Why is `build_export_depend` missing here?

rhaschke (commented):
Apart from the very verbose dependency declaration in package.xml, I approve.

rviz_marker_tools/package.xml
v4hn: because the header exposes no roscpp include - only the source uses it [once](https://github.com/moveit/moveit_task_constructor/blob/184d39897ae28158e57800535f50ace93d0a1016/rviz_marker_tools/src/marker_creation.cpp#L324) - and arguably that line should actually throw instead of printing to the log.

rviz_marker_tools/package.xml
v4hn: I just preferred to stay with one style and some of the others don't use all three lines.
I collapsed the relevant lines now.

PR #611, opened 2024-09-02
Update API: JumpThreshold -> CartesianPrecision

rhaschke:
This should be merged only after the next MoveIt release.

### DIFF
--- a/core/include/moveit/task_constructor/solvers/cartesian_path.h
+++ b/core/include/moveit/task_constructor/solvers/cartesian_path.h
@@ -59,14 +59,16 @@ public:
 
 	void setStepSize(double step_size) { setProperty("step_size", step_size); }
 	void setPrecision(const moveit::core::CartesianPrecision& precision) { setProperty("precision", precision); }
-	[[deprecated("Replace with setPrecision")]]  // clang-format off
-	void setJumpThreshold(double /*jump_threshold*/) {}
+	template <typename T = float>
+	void setJumpThreshold(double) {
+		static_assert(std::is_integral<T>::value, "setJumpThreshold() is deprecated. Replace with setPrecision.");
+	}
 	void setMinFraction(double min_fraction) { setProperty("min_fraction", min_fraction); }
 
 	[[deprecated("Replace with setMaxVelocityScalingFactor")]]  // clang-format off
-	void setMaxVelocityScaling(double factor) { setMaxVelocityScalingFactor(factor); }
+	void setMaxVelocityScaling(double factor) { setMaxVelocityScalingFactor(factor); }  // clang-format on
 	[[deprecated("Replace with setMaxAccelerationScalingFactor")]]  // clang-format off
-	void setMaxAccelerationScaling(double factor) { setMaxAccelerationScalingFactor(factor); }
+	void setMaxAccelerationScaling(double factor) { setMaxAccelerationScalingFactor(factor); }  // clang-format on
 
 	void init(const moveit::core::RobotModelConstPtr& robot_model) override;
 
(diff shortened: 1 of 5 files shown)


core/src/solvers/cartesian_path.cpp
```diff
@@ -54,7 +54,7 @@ CartesianPath::CartesianPath() {
 	auto& p = properties();
 	p.declare<geometry_msgs::PoseStamped>("ik_frame", "frame to move linearly (use for joint-space target)");
 	p.declare<double>("step_size", 0.01, "step size between consecutive waypoints");
-	p.declare<double>("jump_threshold", 1.5, "acceptable fraction of mean joint motion per step");
+	p.declare<moveit::core::CartesianPrecision>("precision", { 1e-3, 1e-2, 1e-4 }, "precision of linear path");
```
v4hn, line 57: Why explicitly provide values here and not rely [on the defaults](https://github.com/moveit/moveit/blob/026c512b12e1723630deef9bf810066d7e95641c/moveit_core/robot_state/include/moveit/robot_state/cartesian_interpolator.h#L50-L55)? And why is the third value different from MoveIt's default of `1e-5`?

core/python/bindings/src/solvers.cpp
```diff
@@ -99,6 +100,19 @@ void export_solvers(py::module& m) {
 	    .property<double>("max_step", "float: Limit any (single) joint change between two waypoints to this amount")
 	    .def(py::init<>());
 
+	py::class_<moveit::core::CartesianPrecision>(m, "CartesianPrecision", "precision for Cartesian interpolation")
+	    .def(py::init([](double translational, double rotational, double max_resolution) {
+		         return new moveit::core::CartesianPrecision{ translational, rotational, max_resolution };
+	         }),
+	         py::arg("translational") = 0.001, py::arg("rotational") = 0.01, py::arg("max_resolution") = 1e-5)
```
v4hn, line 107: I guess you could instantiate a default `CartesianPrecision` here and use the values from there?

core/include/moveit/task_constructor/solvers/cartesian_path.h
```diff
@@ -57,7 +58,9 @@ class CartesianPath : public PlannerInterface
 	void setIKFrame(const std::string& link) { setIKFrame(Eigen::Isometry3d::Identity(), link); }
 
 	void setStepSize(double step_size) { setProperty("step_size", step_size); }
-	void setJumpThreshold(double jump_threshold) { setProperty("jump_threshold", jump_threshold); }
+	void setPrecision(const moveit::core::CartesianPrecision& precision) { setProperty("precision", precision); }
+	[[deprecated("Replace with setPrecision")]]  // clang-format off
```
v4hn, line 62: Let's not provide a no-op function for something that was rather essential before. :neutral_face: 
People tend to ignore deprecations and still assume things will work.
How about a [`static_assert` on use](https://stackoverflow.com/a/18814056/21260084)?

v4hn (approved):
Looks good except for my few comments.

Do you really want to keep this lying here until a new moveit release?
I know we went for supporting released and master branch of moveit in the past, but I'm not sure I see the need for it.
Of course we should hold a new *release* tag until moveit is released as well.

core/include/moveit/task_constructor/solvers/cartesian_path.h
rhaschke: Somewhat ugly, but ok to enforce deprecation.

v4hn:
Thanks for addressing the feedback. :+1: 

> Do you really want to keep this lying here until a new moveit release? I know we went for supporting released and master branch of moveit in the past, but I'm not sure I see the need for it. Of course we should hold a new release tag until moveit is released as well.

What's your plan for this? I would merge here if not for your statement above.

rhaschke:
> What's your plan for this? I would merge here if not for your statement above.

I prefer to have a MoveIt release. Otherwise we enforce people to build from source.
I am happy to cut a new MoveIt release the next days, if there are no objections.

PR #615, opened 2024-09-23
Wrapper FMT header for compatibility

v4hn:
for special eigen formatter, which is not available in fmt9+ (found on Debian testing).

https://github.com/fmtlib/fmt/issues/3465
https://stackoverflow.com/a/73755864/21260084

@rhaschke Not sure whether you are ok with the generic wrapper header as you usually argue for minimal includes. The concrete issue currently only affects a single location where an Eigen datatype is formatted. But I believe as we use libfmt all over the place (as you introduced at some point), I would also expect to be able to format Eigen datastructures everywhere without additional includes & helpers.

### DIFF
--- a/core/include/moveit/task_constructor/fmt_p.h
+++ b/core/include/moveit/task_constructor/fmt_p.h
@@ -1,0 +1,49 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2024, University of Hamburg
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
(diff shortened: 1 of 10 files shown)


rhaschke (commented):
In principle I approve. But the name of the wrapper include (`fmt_p.h`) is too cryptic.
What is the meaning of `_p`?
I suggest simply `fmt.h` or `fmt_wrapper.h`.

v4hn:
> But the name of the wrapper include (fmt_p.h) is too cryptic.
> What is the meaning of _p? I suggest simply fmt.h or fmt_wrapper.h.

[You introduced that suffix](https://github.com/moveit/moveit_task_constructor/blob/99ccc115e041c5e80b8e653815a62936c3d392ec/core/CMakeLists.txt#L49-L50) and it implies that the header will not be installed.
I did not want to provide such a generic wrapper with MTC in general - also to avoid potential [ODR violations](https://github.com/fmtlib/fmt/issues/3465#issuecomment-1565749875) -, but still provide it for all the subfolders in the `core` package.

rhaschke:
> [You introduced that suffix](https://github.com/moveit/moveit_task_constructor/blob/99ccc115e041c5e80b8e653815a62936c3d392ec/core/CMakeLists.txt#L49-L50) and it implies that the header will not be installed.

I see. There the `_p` suffix stands for _private_. I'm okish to use that suffix here as well - now that I know ;-)
But, I would be ok as well to install the header. This doesn't immediately trigger ODR violations.

v4hn:
> now that I know ;-)

:black_joker: 

>  But, I would be ok as well to install the header. This doesn't immediately trigger ODR violations.

I'm aware. I prefer not to install it unless/until we need it in some very related downstream package too.

rhaschke:
It is easily possible that we will use fmt in another package of this repo (aside from core) in the future.

v4hn:
> until we need it

It's not an external API for now and it takes one sed to change it if needed. :dog2:

PR #626, opened 2024-11-06
[WIP] Quick access of stages by id/index

captain-yoshi:
This is somewhat related to #194. For pre-execution purposes.

We can already achieve this without any modification to the current code. This PR is just for demonstration.

1- Create a flat buffer of stages in BFS/DFS order
2- The indexes represents (minus 1) the given `stage_id` from [SolutionInfo.msg](.msgs/msg/SolutionInfo.msg)
3- Use the stage_id to get the stage from the vector directly in O(1) complexity
4- You can retrieve your properties

I think the `stage_id` is primarily used for RViz, thus instropection must be enabled. 

```cpp
task->fillFlatBufferStages();
task->plan();

...

// introspection needed to populate message
moveit_task_constructor_msgs::Solution solution_msg;
task.solutions().front()->toMsg(solution_msg, &task.introspection());

for (std::size_t i = 0; i < solution_msg.sub_trajectory.size(); ++i)
{
    auto& sub_trajectory = solution_msg.sub_trajectory[i];

    uint32_t id = sub_trajectory.info.id;
    uint32_t stage_id = sub_trajectory.info.stage_id;

    // retrieve stage
    const moveit::task_constructor::Stage* stage = task.stages(stage_id - 1);
    if(!stage)
        throw std::runtime_error("explosion")
    
    // retrieve properties
    int my_property = stage->properties().get<int>("my-awesome-property")
}


```

----

Would there be any interest in adding this sort of behavior inside MTC ?

If yes then:
- Could be always on, or enabled by the user.
- Must work without introspection, needs a new ID ?

If no:
- Must work without introspection, needs a new ID ?

### DIFF
--- a/core/include/moveit/task_constructor/task.h
+++ b/core/include/moveit/task_constructor/task.h
@@ -98,6 +98,9 @@ public:
 	void insert(Stage::pointer&& stage, int before = -1) override;
 	void clear() final;
 
+	/// creates a flat buffer of stages in breath-first search order
+	void fillFlatBufferStages();
+
 	/// enable introspection publishing for use with rviz
 	void enableIntrospection(bool enable = true);
 	Introspection& introspection();
@@ -150,6 +153,7 @@ public:
 	/// access stage tree
 	ContainerBase* stages();
 	const ContainerBase* stages() const;
+	const ContainerBase* stages(uint32_t stage_id) const;
 
 	/// properties access
 	PropertyMap& properties();
@@ -165,6 +169,8 @@ protected:
 
 private:
 	using WrapperBase::init;
+
+	std::vector<const ContainerBase*> bfs_stages_;  // breadth-first search
 };
 
 inline std::ostream& operator<<(std::ostream& os, const Task& task) {
--- a/core/include/moveit/task_constructor/task_p.h
+++ b/core/include/moveit/task_constructor/task_p.h
@@ -58,6 +58,7 @@ public:
 
 	const std::string& ns() const { return ns_; }
 	const ContainerBase* stages() const;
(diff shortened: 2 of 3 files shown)


rhaschke:
As far as I can see, this PR attempts to provide access to a `Stage* ` from the `stage_id` field in a `SolutionInfo.msg`.
While the suggested approach (of filling a lookup vector) might currently work, it seems to be very brittle.
If such functionality is desired, a corresponding lookup function should be added to `Introspection` instead.
As solutions are processed/executed external to the generating MTC node, adding properties to the solution seems more reasonable to me.

captain-yoshi:
Agreed that the current approach is best. But I think it cannot serialise complex objects stored in a property, e.x. std::function ?

Even though I could change my properties to be serialisable, is there a way to loop through the solution internally without even needing the lookup vector ? Or it would be too complicated.

```cpp
// 1- current way using Msg
moveit_task_constructor_msgs::Solution solution_msg;
task.solutions().front()->toMsg(solution_msg, &task.introspection());

// 2- using directly the SolutionBase
auto solution = task.solutions().front();

// example
for(sub_solution : solution) {
    stage = sub_solution.stage;
    sub_trajectory = sub_solution.sub_trajectory;
}

```

rhaschke:
> Is there a way to loop through the solution internally without even needing the lookup vector ?

In principle, you could inspect the SolutionBase directly. However, you need to use dynamic casts to determine the actual type. Message generation relies on virtual functions.

captain-yoshi:
Will probably just change my approach with post-planning. It will scale better. Thanks for the answers.

PR #633, opened 2024-12-01
Update libs into  .pre-commit-config.yaml

mosfet80:
Changelog:

https://github.com/psf/black/releases/tag/24.10.0

https://github.com/pre-commit/pre-commit-hooks/releases/tag/v5.0.0


rhaschke:
I had to revert that change as the new version of black requires python 3.9 or later:
`ERROR: Package 'black' requires a different Python: 3.8.10 not in '>=3.9'`

PR #645, opened 2025-01-18
uniform cmake version to the higest used : 3.11

mosfet80:
this also fix cmake <3.10 deprecation


rhaschke:
Where does the choice 3.11 come from? In any case, such basic changes should target the master branch. If will apply a similar change myself. Closing here.

PR #651, opened 2025-02-02
ComputeIK: report failures the modern way

rhaschke:
Don't use command-line warnings, but spawn failure solutions. This fixes https://github.com/moveit/moveit_task_constructor/issues/650#issuecomment-2629030464:
```
Failing stage(s):
compute ik (0/4): ik frame unknown in robot: 'foo'
```

Additionally, to correctly report ComputeIK's failure comments, I had to adapt `explainFailure()`.

### DIFF
--- a/core/include/moveit/task_constructor/container.h
+++ b/core/include/moveit/task_constructor/container.h
@@ -80,7 +80,7 @@ public:
 
 	virtual bool canCompute() const = 0;
 	virtual void compute() = 0;
-	void explainFailure(std::ostream& os) const override;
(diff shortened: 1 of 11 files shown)


DaniGarciaLopez:
Thanks for this improvement @rhaschke! There is one missing check in `generate_pose.cpp` [here](https://github.com/moveit/moveit_task_constructor/blob/83185469924f457f1415a021c86b56d5de4324af/core/src/stages/generate_pose.cpp#L79C2-L82C3), otherwise LGTM.

PR #653, opened 2025-02-09
Provide feedback during ExecuteTaskSolution action

rhaschke:
Alternative implementation to #652

### DIFF
--- a/capabilities/src/execute_task_solution_capability.cpp
+++ b/capabilities/src/execute_task_solution_capability.cpp
@@ -166,21 +166,26 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
 		exec_traj.trajectory_->setRobotTrajectoryMsg(state, sub_traj.trajectory);
 		exec_traj.controller_names_ = sub_traj.execution_info.controller_names;
 
-		/* TODO add action feedback and markers */
-		exec_traj.effect_on_success_ = [this,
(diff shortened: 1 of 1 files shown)


capabilities/src/execute_task_solution_capability.cpp
```diff
@@ -166,21 +166,26 @@ bool ExecuteTaskSolutionCapability::constructMotionPlan(const moveit_task_constr
...
-				return context_->planning_scene_monitor_->newPlanningSceneMessage(scene_diff);
-			}
-			return true;
-		};
+		exec_traj.effect_on_success_ =
+		    [this, &scene_diff = const_cast<::moveit_msgs::PlanningScene&>(sub_traj.scene_diff), description, i,
+		     no = solution.sub_trajectory.size()](const plan_execution::ExecutableMotionPlan* /*plan*/) {
+			    // publish feedback
+			    moveit_task_constructor_msgs::ExecuteTaskSolutionFeedback feedback;
+			    feedback.sub_id = i;
```
DaniGarciaLopez, line 174: Just a quick remark. If we keep it as is now (publishing the feedback only after the subtrajectory finishes), I think it makes more sense to use `i + 1` here to match what is parsed in `exec_traj.description_`. Otherwise, if we decide to publish feedback at the start of the subtrajectory as well, publishing 0 initially would be reasonable.

PR #665, opened 2025-02-24
Remove clamping of joint constraint in MoveRelative stage

### LINKED ISSUE #664: Do not enforce position bounds for a joint MoveRelative
Is there a reason why a target relative joint position is clamped to a valid position ? I would expect to have an error when a MoveRelative cannot be achieved. Removing this [line](https://github.com/moveit/moveit_task_constructor/blob/master/core/src/stages/move_relative.cpp#L97)  stops the planning with a `Waypoint is out of bounds!` error instead of passing.

captain-yoshi:
Fixes #664. Do not enforce position bounds, let the stage fail if joints are outside the limits instead of clamping to valid positions.

There are maybe some who would actually want to move relative up to the position limit. But there's a lot chances that this will fail in subsequent stages.


rhaschke (approved):
Thanks for providing the PR.

PR #672, opened 2025-03-31
Add node name remap rule for rviz2

Juliaj:
Address https://github.com/moveit/moveit2_tutorials/issues/991 where two nodes are launched with same name rviz2. 

This is companion PR for https://github.com/moveit/moveit2_tutorials/pull/1030.

### DIFF
--- a/demo/launch/demo.launch.py
+++ b/demo/launch/demo.launch.py
@@ -37,9 +37,8 @@ def generate_launch_description():
     rviz_node = Node(
         package="rviz2",
         executable="rviz2",
-        name="rviz2",
         output="log",
-        arguments=["-d", rviz_config_file],
+        arguments=["-d", rviz_config_file, "--ros-args", "--remap", "rviz:__name:=rviz2"],
         parameters=[
             moveit_config.robot_description,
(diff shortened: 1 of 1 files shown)


rhaschke:
I can only underline https://github.com/moveit/moveit2_tutorials/pull/1030#pullrequestreview-2728353208.
Aren't _all_ the node names still forced to be `rviz2` this way?

Juliaj:
Thanks @rhaschke for your feedback. I fell back to the suggestion from @sea-bass and removed the name attribute. PR is updated if you could give it another look.  

Regarding to your question, 

> Aren't _all_ the node names still forced to be `rviz2` this way?

With name remapping change for rviz2. Three of the nodes renamed to rviz2 which is aligned with package name rviz2. The rviz_visual_tools_gui hasn't been renamed, here is some inconsistency.  See https://github.com/moveit/moveit2_tutorials/pull/1030#issuecomment-2766730142 for differences among three options. 
```
/rviz2
/rviz2_private_139289413291728
/rviz2_private_139289816608576
/rviz_visual_tools_gui
```

rhaschke (approved):

PR #684, opened 2025-05-23
Add ability to preempt execution and add test

MarqRazz:
This PR is an update of #363 and adds a test to ensure it is working as expected.

Additional changes I made to make it possible:
- Added the ability to not start Rviz when running demo.launch.py (I use this launch file in my test and needed the ability to disable it)
- Added a persistent node and action client to `Task`
Extended testing of my motion system has found that recreating the node on each execution request can cause a seg fault after many hours of running. I am continuing to test this and will update with any results, but if you feel this is out of scope I can move this feature to a different PR. Here is a link to the [upstream issue](https://github.com/eclipse-cyclonedds/cyclonedds/issues/2022) with cyclonedds and changes they have made.
```
[my_motion_node-9] my_motion_node: ./src/core/ddsc/src/dds_handles.c:430: dds_handle_drop_childref_and_pin: Assertion `(cf & HDL_REFCOUNT_MASK) > 0' failed.
[my_motion_node-9] Stack trace (most recent call last) in thread 686402:
[my_motion_node-9] #27   Object "/usr/lib/aarch64-linux-gnu/ld-linux-aarch64.so.1", at 0xffffffffffffffff, in 
[my_motion_node-9] #26   Object "/usr/lib/aarch64-linux-gnu/libc.so.6", at 0xffffb3adba4b, in 
[my_motion_node-9] #25   Object "/usr/lib/aarch64-linux-gnu/libc.so.6", at 0xffffb3a7595b, in 
[my_motion_node-9] #24   Object "/usr/lib/aarch64-linux-gnu/libstdc++.so.6.0.33", at 0xffffb3d81adf, in 
...
[my_motion_node-9] #19   Object "/root/ws/install/my_motion_node", at 0xaaaad172b93b, in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use_cold()
[my_motion_node-9] #18   Object "/root/ws/install/my_motion_node", at 0xaaaad176fff7, in std::_Sp_counted_deleter<rclcpp_action::Client<moveit_task_constructor_msgs::action::ExecuteTaskSolution>*, rclcpp_action::create_client<moveit_task_constructor_msgs::action::ExecuteTaskSolution>(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeGraphInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeLoggingInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeWaitablesInterface>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<rclcpp::CallbackGroup>, rcl_action_client_options_s const&)::{lambda(rclcpp_action::Client<moveit_task_constructor_msgs::action::ExecuteTaskSolution>*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose()
[my_motion_node-9] #17   Object "/root/ws/install/my_motion_node", at 0xaaaad176f58f, in rclcpp_action::create_client<moveit_task_constructor_msgs::action::ExecuteTaskSolution>(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeGraphInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeLoggingInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeWaitablesInterface>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<rclcpp::CallbackGroup>, rcl_action_client_options_s const&)::{lambda(rclcpp_action::Client<moveit_task_constructor_msgs::action::ExecuteTaskSolution>*)#1}::operator()(rclcpp_action::Client<moveit_task_constructor_msgs::action::ExecuteTaskSolution>*) const
```

### DIFF
--- a/core/src/task.cpp
+++ b/core/src/task.cpp
@@ -283,13 +283,15 @@ void Task::resetPreemptRequest() {
 }
 
 moveit::core::MoveItErrorCode Task::execute(const SolutionBase& s) {
-	// Add random ID to prevent warnings about multiple publishers within the same node
-	auto node = rclcpp::Node::make_shared("moveit_task_constructor_executor_" +
-	                                      std::to_string(reinterpret_cast<std::size_t>(this)));
+	// If this is the first call to execute create a persistent node that can be used to call the action server
+	if (!execute_solution_node_) {
+		execute_solution_node_ = rclcpp::Node::make_shared("moveit_task_constructor_executor");
+	}
 	auto ac = rclcpp_action::create_client<moveit_task_constructor_msgs::action::ExecuteTaskSolution>(
-	    node, "execute_task_solution");
+	    execute_solution_node_, "execute_task_solution");
 	if (!ac->wait_for_action_server(0.5s)) {
-		RCLCPP_ERROR(node->get_logger(), "Failed to connect to the 'execute_task_solution' action server");
+		RCLCPP_ERROR(execute_solution_node_->get_logger(),
+		             "Failed to connect to the 'execute_task_solution' action server");
 		return moveit::core::MoveItErrorCode::FAILURE;
 	}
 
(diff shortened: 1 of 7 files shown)


rhaschke:
> > It would be much more reasonable to provide access to the ActionServer:
> 
> The problem with this approach is that the user would need access to the `ActionClient` and `node`

Agreed. Given that situation, handling the cancellation internal to task seems reasonable.
I have one other idea to avoid the busy loop: One could simply cancel the execution goal when calling `Task::preempt()`.
This would require saving the latest goal_handle.

> I can copy the launch file into the execution test or move it over to the demo package, do you have a preference on how I fix this?

Could you simply reuse [`core/test/test.launch.py`](https://github.com/moveit/moveit_task_constructor/blob/bab7b0453216beace13bb8fb8a8517c524af2a8e/core/test/test.launch.py)?

core/src/task.cpp
```diff
@@ -283,13 +281,15 @@ void Task::resetPreemptRequest() {
 }
 
 moveit::core::MoveItErrorCode Task::execute(const SolutionBase& s) {
-	// Add random ID to prevent warnings about multiple publishers within the same node
-	auto node = rclcpp::Node::make_shared("moveit_task_constructor_executor_" +
-	                                      std::to_string(reinterpret_cast<std::size_t>(this)));
```
rhaschke, line 288: We still need a task-specific execution node as we might have multiple task instances.

rhaschke:
> > I have one other idea to avoid the busy loop: One could simply cancel the execution goal when calling Task::preempt().
> > This would require saving the latest goal_handle.
> 
> I'm still looking into this but wouldn't it still require a busy loop blocking `execute` from returning until motion or cancel is complete?

No. I think you would simply use the existing code:
`rclcpp::spin_until_future_complete(node, result_future)`

MarqRazz:
I was actually thinking of calling spin multiple times on an executor. For example the [single_threaded_executor](https://github.com/ros2/rclcpp/blob/jazzy/rclcpp/src/rclcpp/executors/single_threaded_executor.cpp#L31) throws an exception if you do this.

I tried saving the goal handle and moved the cancel call to `preempt` and my test segfaults if I wait for the cancel result.
```

rhaschke:
Why does move_group crash on shutdown? This shouldn't happen.

rhaschke (approved):

PR #686, opened 2025-05-28
Add functions for adding markers to solutions

mjimenezm00:
Add and reset marker management functions in Introspection

- `addMarkers`: Associates a vector of visualization markers with a solution, either by solution ID or by SolutionBase reference.
- `resetMarkers`: Clears all markers associated with a specific solution ID.
- `resetAllMarkers`: Clears all stored markers for all solutions.

These functions enable storing, updating, and clearing visualization markers for solutions.


rhaschke:
I don't yet see the use case of this. Markers should be added to the solutions, not to the introspection instance.
Can you please elaborate on your intended use case?

PR #689, opened 2025-05-30
Reset preempt_request after canceling execution

MarqRazz:
Integrating preempt execution into my downstream application I found that when reusing my Task object between requests it would fail planning after preempting motion because the `preempt_requested_` flag was never reset. Sorry for missing this in #684 I thought the flag was reset on each `plan()` request but it looks like #597 has some use case where that causes issues.

### DIFF
--- a/capabilities/test/test_task_execution.cpp
+++ b/capabilities/test/test_task_execution.cpp
@@ -60,6 +60,12 @@ TEST_F(PandaMoveTo, preemptExecution) {
 	}
 
 	EXPECT_EQ(result.val, moveit_msgs::msg::MoveItErrorCodes::PREEMPTED);
+
+	// After preempting motion reset the task and make sure we can successfully plan and execute motion again
+	t.reset();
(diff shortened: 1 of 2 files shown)


capabilities/test/test_task_execution.cpp
```diff
@@ -60,6 +60,13 @@ TEST_F(PandaMoveTo, preemptExecution) {
 	}
 
 	EXPECT_EQ(result.val, moveit_msgs::msg::MoveItErrorCodes::PREEMPTED);
+
+	// After preempting motion reset the task and make sure we can successfully plan and execute motion again
+	rclcpp::sleep_for(std::chrono::seconds(1));
```
rhaschke, line 65: Why is this sleep necessary?

rhaschke (approved):

MarqRazz:
Can we please back port this to Jazzy? I'd be happy to help with this in the future if you could help me come up to speed on the process.

rhaschke:
Done.

PR #697, opened 2025-06-25
qt6 Support

mosfet80:
Fix compile with qt6


rhaschke:
This mixes Qt6 changes and ament_target_dependencies() replacement.

Issue #698, opened 2025-07-01
GeneratePose creates many solutions - possible option for new filter?


joseph-moore-CO:
I'm planning a pick and place task where I use ComputeIK to generate IK solutions for both the pick and the place poses. The task contains several planning scene modifications to attach objects and adjust collisions, and a central Connect stage to join them together. The planning scene is very constrained so a large number of the IK combinations will be unable to successfully plan the connection.

In order to maintain compatibility of the two sides of the problem at the Connect stage, the place pose GeneratePose generator stage is monitoring one of the ModifyPlanningScene stages from the "pick section" which has the collision object attached.

The monitored stage has multiple solutions, so I end up creating 8 solutions for my GeneratePose stage which have slightly different notions of where the robot is, but have exactly the same pose. The wrapping ComputeIK stage then creates >150 total solutions, which bogs down the pipeline with failed Connect plans from practically the same states.

My thought was to create a wrapper stage which only passes on the first/first N solution(s). I could then wrap the GeneratePose stage with that to filter out to just one solution before passing into ComputeIK. Does this sound like a reasonable approach, and would it make a useful contribution to this project?

Here's a sketch of my entire planning task.

↕ CurrentState
↓ ModifyPlanningScene "/allow_collisions"
• Connect
↑ MoveTo
↕ ComputeIK
       ↕ GeneratePose (monitors "/allow_collisions")
↓ ModifyPlanningScene "/object_attached "(adds and attaches collision object)
↓ MoveTo
↓ MoveTo
↓ ModifyPlanningScene (forbids collisions)
• Connect
↑ ModifyPlanningScene (allows collisions)
↑ MoveTo
↑ MoveTo
↑ MoveTo
↕ ComputeIK
     ↕ GeneratePose (monitors "/object_attached", generates 8 solutions)
↓ ModifyPlanningScene (detaches object)
↓ MoveTo

PR #699, opened 2025-07-07
Add wrapper stage to limit the number of solutions returned by the child (#698)

joseph-moore-CO:
Following from issue #698 https://github.com/moveit/moveit_task_constructor/issues/698, this PR adds a new wrapper stage called LimitSolutions which adds a limit to the number of solutions which are transmitted by the child stages.

As suggested by @rhaschke (https://github.com/moveit/moveit_task_constructor/issues/698#issuecomment-3024509137), this is performed lazily with the lowest cost upstream solution being lifted on each call to compute.


joseph-moore-CO:
Mistake on raising this PR. I had intended it to go into the ros2 branch!

PR #701, opened 2025-07-15
Fix build issues by replacing fmt::join with rcpputils::join for better cross-platform compatibility

idesign0:
When building on macOS (Apple Silicon), the use of `fmt::join` causes build errors due to incompatibilities with the available version of the `fmt` library. The error typically indicates that no matching function for `join` exists.

This PR replaces `fmt::join` with `rcpputils::join`, which provides equivalent functionality and is fully compatible with ROS 2 environments on macOS.

- Replacing `fmt::join` with `rcpputils::join` resolves build errors on macOS while ensuring compatibility across all other supported platforms.

### DIFF
--- a/capabilities/CMakeLists.txt
+++ b/capabilities/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.16)
 project(moveit_task_constructor_capabilities)
 
 find_package(fmt REQUIRED)
+find_package(rcpputils REQUIRED)
 find_package(catkin REQUIRED COMPONENTS
   actionlib
   moveit_core
@@ -33,7 +34,7 @@ add_library(${PROJECT_NAME}
 )
 add_dependencies(${PROJECT_NAME} ${catkin_EXPORTED_TARGETS})
 target_include_directories(${PROJECT_NAME} SYSTEM PUBLIC ${catkin_INCLUDE_DIRS})
-target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} fmt::fmt)
+target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} fmt::fmt rcpputils::rcpputils)
 
 install(TARGETS ${PROJECT_NAME}
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
--- a/core/test/test_move_relative.cpp
+++ b/core/test/test_move_relative.cpp
@@ -174,7 +174,7 @@ TEST_F(PandaMoveRelativeJoint, jointOutsideBound) {
 	// move joint inside limit
 	auto initial_jpos = scene->getCurrentState().getJointPositions("panda_joint7");
 	move->setDirection([initial_jpos] {
-			return std::map<std::string, double>{ { "panda_joint7", 2.0 - *initial_jpos } };
+		return std::map<std::string, double>{ { "panda_joint7", 2.0 - *initial_jpos } };
 	}());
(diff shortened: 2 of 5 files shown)


idesign0:
Yes, that’s correct — locally(Humble) it works without explicitly including or linking rcpputils may be due to indirect includes or environment setup, so I kept it that way.

> I would expect that you need a corresponding `#include` directive and link against the library?

Yes, you are right! — locally(Humble) it works without explicitly including or linking rcpputils may be due to indirect includes or environment setup, so I kept it that way.

rhaschke:
I guess you want to target the `ros2` branch. By the way, which version of fmtlib do you use on MacOS?

idesign0:
> I guess you want to target the `ros2` branch. By the way, which version of fmtlib do you use on MacOS?

yes you have point. i have fmt 11.2.0 in macOS.

rhaschke:
> i have fmt 11.2.0 in macOS.

fmt 11 has `join()` in `fmt/ranges.h`.

capabilities/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.16)
 project(moveit_task_constructor_capabilities)
 
 find_package(fmt REQUIRED)
+find_package(rcpputils REQUIRED)
```
rhaschke, line 5: you need to declare the requirement in package.xml as well.

core/test/test_move_relative.cpp
```diff
@@ -174,15 +174,15 @@ TEST_F(PandaMoveRelativeJoint, jointOutsideBound) {
 	// move joint inside limit
 	auto initial_jpos = scene->getCurrentState().getJointPositions("panda_joint7");
 	move->setDirection([initial_jpos] {
-		return std::map<std::string, double>{ { "panda_joint7", 2.0 - *initial_jpos } };
+			return std::map<std::string, double>{ { "panda_joint7", 2.0 - *initial_jpos } };
```
rhaschke, line 177: Why changing this indentation?

core/test/test_move_relative.cpp
```diff
@@ -174,15 +174,15 @@ TEST_F(PandaMoveRelativeJoint, jointOutsideBound) {
...
 	}());
 	EXPECT_TRUE(this->t.plan()) << "Plan should succeed, joint inside limit";
 
 	this->t.reset();
 
 	// move joint outside limit: 2.8973
 	move->setDirection([initial_jpos] {
-		return std::map<std::string, double>{ { "panda_joint7", 3.0 - *initial_jpos } };
+			return std::map<std::string, double>{ { "panda_joint7", 3.0 - *initial_jpos } };
```
rhaschke, line 185: Same here.

rhaschke (commented):
I'm not yet convinced to switch to rcpputils. Why not simply including the correct fmt header?

core/test/test_move_relative.cpp
idesign0: Pre-commit uses a specific clang-format version and config that may differ from my local setup. By manually formatting, I accidentally added extra space. I’ll fix it to match the pre-commit style and avoid failures.

idesign0:
> > i have fmt 11.2.0 in macOS.
> 
> fmt 11 has `join()` in `fmt/ranges.h`.

I wasn’t aware of that. It’s better to include fmt/ranges.h then.

idesign0:
> > It’s better to include fmt/ranges.h then.
> 
> 
> 
> Do you plan providing a PR for this approach?

I agree. Yes I will create one in couple of days. 👍

PR #704, opened 2025-07-18
drop internal pybind11

mosfet80:
drop internal pybind11
use use ros pybynd11_vendor


rhaschke:
This will be handled as soon as https://github.com/ros2/pybind11_vendor/pull/32 is merged and released.

PR #708, opened 2025-08-13
Enable collisions visualization

rhaschke:
Further improvements of #691, targeting the master branch.


mjimenezm00:
Hi! First of all, thanks for implementing this. Will it be added to the `ros2` branch as well? I'm currently working with it, and having this feature there would be really useful

PR #723, opened 2025-11-16
Add ability to preempt the task when a stage fails (preempt_on_failure)

DaniGarciaLopez:
In our tasks, we make extensive use of the `PredicateFilter` stage to perform certain checks in the scene before other stages are executed. This stage is expected to always succeed for the task to proceed. However, the task currently continues planning and looks for alternative solutions in earlier stages, which unnecessarily prolongs planning time—even though we already know that if this stage fails, the task is not feasible and there is no reason to continue.

With this in mind, we propose adding a `preempt_on_failure` property. If set to true, the stage must always succeed; if the stage fails even once, the task is immediately preempted and planning aborts.

We can currently do something similar using `addSolutionCallback` and throwing an exception when any solution fails, but this does not update introspection correctly.

Thoughts, @rhaschke?

### DIFF
--- a/core/include/moveit/task_constructor/stage.h
+++ b/core/include/moveit/task_constructor/stage.h
@@ -211,6 +211,11 @@ public:
 		return properties().get<TrajectoryExecutionInfo>("trajectory_execution_info");
 	}
 
+	/// If true, a failure of this stage will cause immediate preemption of the task
+	void setPreemptOnFailure(bool preempt_on_failure) { setProperty("preempt_on_failure", preempt_on_failure); }
+	/// Get whether this stage is configured to preempt the task on failure
+	bool preemptOnFailure() const { return properties().get<bool>("preempt_on_failure"); }
+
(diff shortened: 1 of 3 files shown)


DaniGarciaLopez:
> Why don't you simply call `Task::preempt()` in the callback?

Because we don't have easy access to the task pointer. Is there a way to get it from the solution?

I tried traversing each parent, but the top-level container is not a `Task`, so this unfortunately doesn’t work:

```cpp
applicability_filter->addSolutionCallback(
  [](const mtc::SolutionBase& solution) {
    if (solution.isFailure()) {
      const mtc::ContainerBase* container = solution.creator()->parent();
      while (container && container->parent())
          container = container->parent();

      const mtc::Task* task = dynamic_cast<const mtc::Task*>(container);
      if (task)
        const_cast<mtc::Task*>(task)->preempt();
    }
  }
);
```

PR #745, opened 2026-04-22
LimitSolutions: prevent child compute calls after max_solutions is reached

DaniGarciaLopez:
Currently, when `max_solutions` is reached, the child may still continue computing additional solutions. This was not an issue with a simple `Generator` stage, but with a large `SerialContainer` with multiple nested stages, the computational cost becomes significant.

Moreover, these additional solutions are not propagated, since the limit for forwarding solutions has already been reached. Thus, this extra computation is wasted.

With the change proposed in this PR, once the limit of forwarded solutions (`max_solutions`) is reached, the child container will stop generating further solutions.

For future reference, as discussed [here](https://github.com/moveit/moveit_task_constructor/issues/728#issuecomment-3612165496), a better approach might be to relax the `max_solutions` restriction and allow incremental forwarding of solutions, before the task fails. While limiting solutions helps reduce combinatorial explosion in long tasks, it also risks missing valid alternatives. @rhaschke Do you see a simple way—without major refactoring—to achieve this using the current wrapper implementation?

### DIFF
--- a/core/src/stages/limit_solutions.cpp
+++ b/core/src/stages/limit_solutions.cpp
@@ -73,10 +73,10 @@ void LimitSolutions::compute() {
 	if (forwarded_solutions >= max_solutions)
 		return;
 
-	if (forwarded_solutions < max_solutions && WrapperBase::canCompute())
+	if (WrapperBase::canCompute())
 		WrapperBase::compute();
 
-	if (!upstream_solutions_.empty() && forwarded_solutions < max_solutions) {
+	if (!upstream_solutions_.empty()) {
 		++forwarded_solutions;
 		liftSolution(*upstream_solutions_.pop());
 	}
--- a/core/src/stages/limit_solutions.cpp
+++ b/core/src/stages/limit_solutions.cpp
@@ -54,24 +54,32 @@ void LimitSolutions::reset() {
 }
 
 void LimitSolutions::onNewSolution(const SolutionBase& s) {
-	uint32_t max_solutions = properties().get<uint32_t>("max_solutions");
+	const uint32_t max_solutions = properties().get<uint32_t>("max_solutions");
+
 	if (forwarded_solutions + upstream_solutions_.size() < max_solutions)
 		upstream_solutions_.push(&s);
 }
 
 bool LimitSolutions::canCompute() const {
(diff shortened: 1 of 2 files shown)


core/src/stages/limit_solutions.cpp
```diff
@@ -54,24 +54,32 @@ void LimitSolutions::reset() {
...
+	const uint32_t max_solutions = properties().get<uint32_t>("max_solutions");
 
-	if (upstream_solutions_.empty())
+	// once the limit is reached do not call compute on the child anymore
+	if (forwarded_solutions >= max_solutions)
 		return;
 
-	++forwarded_solutions;
-	liftSolution(*upstream_solutions_.pop());
+	if (forwarded_solutions < max_solutions && WrapperBase::canCompute())
```
rhaschke: The check is redundant:
```suggestion
	if (WrapperBase::canCompute())
```

core/src/stages/limit_solutions.cpp
```diff
@@ -54,24 +54,32 @@ void LimitSolutions::reset() {
...
+		WrapperBase::compute();
+
+	if (!upstream_solutions_.empty() && forwarded_solutions < max_solutions) {
```
rhaschke: Here too:
```suggestion
	if (!upstream_solutions_.empty()) {
```

rhaschke (commented):
In general looks good. I think the two last checks for `forwarded_solutions < max_solutions` are redundant and can be dropped - `forwarded_solutions` doesn't change in between, does it?

rhaschke:
> For future reference, as discussed [here](https://github.com/moveit/moveit_task_constructor/issues/728#issuecomment-3612165496), a better approach might be to allow incremental forwarding of solutions. @rhaschke Do you see a simple way—without major refactoring—to achieve this using the current wrapper implementation?

No, the wrapper approach here, doesn't allow for this solution either.

core/src/stages/limit_solutions.cpp
DaniGarciaLopez: You're right, thanks for that simplification!

rhaschke:
Thanks.

PR #749, opened 2026-06-06
Fix `DisplaySolution::fillMessage()` drops `info` and `execution_info` from SubTrajectory messages

lunarifish:
When clicking the "Execute solution" button on the Motion Planning Tasks panel, the action goal sent to the `execute_task_solution` action server was missing the `SolutionInfo` and `TrajectoryExecutionInfo` fields in every `SubTrajectory`.

The internal `DisplaySolution::Data` struct had no info_ or execution_info_ fields. Consequently, `DisplaySolutionfillMessage()` had nothing to write back for `SubTrajectory::info` and `SubTrajectory::execution_info`, so they were always sent as empty values.

Any action server that relies on `execution_info.controller_names` or other `SolutionInfo` fields like `markers`, `planner_id`, `cost`, `id` would receive empty data. I happened to hit this because my action server uses `execution_info`, and I triggered it through the panel button.

### DIFF
--- a/visualization/visualization_tools/include/moveit/visualization_tools/display_solution.h
+++ b/visualization/visualization_tools/include/moveit/visualization_tools/display_solution.h
@@ -83,8 +83,6 @@ class DisplaySolution
 		std::string comment_;
 		/// id of creating stage
 		uint32_t creator_id_;
-		/// solution info (stage_id, markers, etc.)
-		moveit_task_constructor_msgs::msg::SolutionInfo info_;
 		/// trajectory execution info (controller_names, etc.)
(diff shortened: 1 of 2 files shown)


rhaschke (commented):
I deliberately chose to drop the `SolutionInfo` to reduce the memory footprint.
However, I agree to store the `ExecutionInfo` as this is obviously required for execution.
All other fields (like markers) are not required for solution execution.

lunarifish:
Okay! I have removed `SolutionInfo` from the `Data` struct.

rhaschke (approved):

PR #752, opened 2026-07-27
fix qt6compile

mosfet80:
Fix compile with qt6
fix trailing whitespace


rhaschke:
Superseeded by #754 and #757

